PR #12, opened 2023-09-11
fix(docs): update exporting.mdx to include nr instrumentation

lazyplatypus:
Let me know if there is anything i need to edit :)


nirga (commented):
Thanks! Looks good! I'd also add it to the main README for visibility.

nirga (approved):

PR #66, opened 2023-09-15
fix(traceloop-sdk): flushing in notebooks

### LINKED ISSUE #65: 🐛 Bug Report: missing traces when running on Colab

Traceloop SDK

### 📜 Description

Combined with the haystack implementation, traces sometime do not appear in the dashboard.

### 👟 Reproduction steps

This is a simple example that when run in a colab, do not always product traces:
```python
import os
from haystack.nodes import PromptNode, PromptTemplate, AnswerParser
from haystack.pipelines import Pipeline

Traceloop.init(app_name="haystack_app")

prompt = PromptTemplate(
    prompt="Tell me a joke about {query}\n",
    output_parser=AnswerParser(),
)

prompt_node = PromptNode(
    model_name_or_path="gpt-4",
    api_key=os.getenv("OPENAI_API_KEY"),
    default_prompt_template=prompt,
)

pipeline = Pipeline()
pipeline.add_node(component=prompt_node, name="PromptNode", inputs=["Query"])

query = "OpenTelemetry"
result = pipeline.run(query)
print(result["answers"][0].answer)
```

### 👍 Expected behavior

Traces should be produced 

See https://github.com/deepset-ai/haystack-integrations/pull/51#issuecomment-1720992240

### 🤖 Python Version

_No response_

This doesn't happen if we explicitly annotate methods with `@workflow`. The main difference is that this annotation forces flush of traces at the end of traces. We did recently add an auth-flush on app shutdown - but this probably never gets called on notebooks.

nirga:
Fixes #65 

The solution is to switch to a non-batching processor for notebooks, which will "flush" every trace directly to the exporter. We should probably do something like this for local runs as well at some point.

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/utils/__init__.py
+++ b/packages/traceloop-sdk/traceloop/sdk/utils/__init__.py
@@ -12,3 +12,15 @@ def camel_to_snake(s):
         return s.lower()
 
     return cameltosnake(s[0].lower() + s[1:])
+
+
+def is_notebook():
+    try:
+        from IPython import get_ipython
+
+        ip = get_ipython()
+        if ip is None:
+            return False
+        return True
+    except Exception:
+        return False
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -15,6 +15,7 @@ body:
         - "Bedrock Instrumentation"
         - "Chromadb Instrumentation"
         - "Cohere Instrumentation"
+        - "Haystack Instrumentation"
         - "OpenAI Instrumentation"
         - "Pinecone Instrumentation"
         - "VertexAI Instrumentation"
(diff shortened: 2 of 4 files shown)


packages/traceloop-sdk/traceloop/sdk/utils/__init__.py
```diff
@@ -12,3 +12,15 @@ def camel_to_snake(s):
         return s.lower()
 
     return cameltosnake(s[0].lower() + s[1:])
+
+
+def is_notebook():
```
galkleinman, line 17: [naming]

```suggestion
def should_process_single_spans():
```

galkleinman: you can then have another method of `is_notebook` inside which is one case of single span processing. I just don't want to tie up the tracing code with notebook or not notebooks and other cases.

galkleinman (approved):

packages/traceloop-sdk/traceloop/sdk/utils/__init__.py
nirga: I get it, but it shouldn't be in utils anyway IMO. I just created a separate method for initializing `span_processor`

PR #154, opened 2023-11-07
fix: new OpenAI API

nirga:


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/__init__.py
```diff
@@ -23,26 +24,47 @@
...
+        "span_name": "openai.chat",
+    },
+    {
+        "module": "openai.resources.completions",
+        "object": "Completions",
+        "method": "create",
+        "span_name": "openai.completion",
+    },
+]
+
```
galkleinman, line 56: Consider preserving one variable, a dict though instead of an array where each key represents a major version and value an array of wrapped methods.

Might help when there will be another version, we'll be able to pick the right wrapped methods array by comparing the key with the sdk version

galkleinman (approved):
LGTM

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/__init__.py
nirga: Hope that they won't do any huge breaking changes in the future 😅

PR #202, opened 2023-11-21
add OpenAI User & Headers args

cmpxchg16:
Add tracking of the user & headers arguments, e.g:

```
response = openai.ChatCompletion.create(
    model="gpt-3.5-turbo",
    messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Who won the world series in 2020?"},
        {"role": "assistant", "content": "The Los Angeles Dodgers won the World Series in 2020."},
        {"role": "user", "content": "Where was it played shlomi@vibranium-dome.com?"},
    ],
    temperature=1,
    request_timeout=5,
    user="user-123456",
    headers={"x-session-id": "abcd-1234-cdef"},
)
```


nirga:
Thanks @cmpxchg16! Can you add a test for that under `traceloop-sdk`?

nirga:
hmm not sure why the tests are failing. Seems like a problem with OpenAI SDK. Upgrading it to 1.3.4 seems to resolve it (#203). I'll rebase here.

cmpxchg16:
> Thanks @cmpxchg16! Can you add a test for that under `traceloop-sdk`?

@nirga when I tried to run tests, it requires some api key (Incorrect API key provided for OpenAI); why the tests are depends in api key..?

nirga:
Yes @cmpxchg16 our current tests are more like [integration tests](https://www.traceloop.com/docs/openllmetry/contributing/developing) where we actually make calls to OpenAI, Anthropic, etc and make sure that the requests and responses are logged correctly.

Yep, it's not perfect but unfortunately that was the easiest way to make sure nothing breaks with changes providers are making to their SDKs. I'll run the tests for you.

nirga (approved):

PR #359, opened 2024-01-24
fix: `resource_attributes` always being None

reiyw:
This pull request fixes a problem where `app_name` and `resource_attributes` passed to the `Traceloop.init()` method were not being added to the Trace attributes because `resource_attributes` was always `None`.

The `update()` method of the `dict` is an in-place method with a return value of `None`, so there is no need to reassign the original variable.

### DIFF
--- a/packages/traceloop-sdk/tests/conftest.py
+++ b/packages/traceloop-sdk/tests/conftest.py
@@ -9,7 +9,12 @@ pytest_plugins = []
 @pytest.fixture(scope="session")
 def exporter():
     exporter = InMemorySpanExporter()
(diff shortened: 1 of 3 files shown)


nirga:
Thanks for this @reiyw! Mind adding a test to make sure this doesn't happen again?

nirga:
Actually nvm I'm adding it myself 😅

nirga (approved):

reiyw:
Thanks for the quick fix!

gyliu513:
Found same issue just now and trying to create a PR, found the code has been updated, good job @reiyw !

PR #362, opened 2024-01-24
Add more semantic conventions to query embeddings / query result events

paolorechia:


packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv/ai/__init__.py
```diff
@@ -46,6 +46,18 @@ class SpanAttributes:
 
 class EventName(Enum):
     VECTOR_DB_QUERY_EMBEDDINGS = "vector_db.query.embeddings"
+    VECTOR_DB_QUERY_RESULT = "vector_db.query.result"
+
+
+class EmbeddingsEventAttributes(Enum):
+    VECTOR = "vector_db.query.embeddings.vector"
```
nirga, line 53: Let's use the same notation we had for SpanAttributes - all are in the same enum

paolorechia:
@nirga  is this what you mean? I merged all of them into a single `EventAttributes` enum

paolorechia:
OK, also standardized it a bit more.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv/ai/__init__.py
```diff
@@ -44,8 +44,17 @@ class SpanAttributes:
     LLM_REPETITION_PENALTY = "llm.watsonx.repetition_penalty"
 
 
-class EventName(Enum):
+class EventAttributes(Enum):
+    # Query Embeddings
     VECTOR_DB_QUERY_EMBEDDINGS = "vector_db.query.embeddings"
```
nirga, line 49: What's the difference between `VECTOR_DB_QUERY_EMBEDDINGS` and `VECTOR_DB_QUERY_EMBEDDINGS_VECTOR`

paolorechia:
The first is the event name and the second is the attribute vector (where
the embeddings vector is stored)

Nir Gazit ***@***.***> schrieb am Do. 25. Jan. 2024 um 16:49:

nirga:
@paolorechia ahh got it! So let's put the event name in a separate `enum` and then have an `enum` for all event attributes, WDYT?

paolorechia:
Sounds good.

Nir Gazit ***@***.***> schrieb am Do. 25. Jan. 2024 um 17:08:

paolorechia:
@nirga There we go :)

nirga (approved):

Issue #258, opened 2023-12-16
🐛 Bug Report: cannot override auto-instrumented packages


nirga:
### Which component is this bug for?

Traceloop SDK

### 📜 Description

We need a way to override and specify the packages to auto-instrument. Right now we [do this automatically](https://github.com/traceloop/openllmetry/blob/main/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L251C14-L251C14) based on installed packages, which might not fit everyone.

### 👟 Reproduction steps

Issues with transformer package we saw with litellm maintainer - https://github.com/BerriAI/litellm/issues/1160

### 👍 Expected behavior

-

### 👎 Actual Behavior with Screenshots

-

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

Mizba-Anjum:
Hi I tried looking at this and I made a PR #363 that adds an option to specify the packages used in an array argument in the init() function.
ex: `Traceloop.init(instruments=['openai'])`
 Let me know if it helps or could be improved!

PR #363, opened 2024-01-24
Create manual instrumentation as well as auto instrumentation

Mizba-Anjum:
Related to #258, provides an option to choose packages (referred to as 'instruments') in an array in the init function of Traceloop.
Example:
`Traceloop.init(instruments=['openai'])`
can include:
`openai
anthropic
cohere
pinecone
chroma
langchain
llama_index
transformers
requests
urllib3
pymysql
bedrock
replicate
vertexai
watsonx`

includes a warning if no packages were chosen
image
warning currently does not throw error or stop the application so the LLM may still be called

It cannot determine if the chosen package is the correct one. ex: choosing the anthropic instrument while using openai will result in no warning and no LLM data despite generating a trace:
With correct instrument usage:
image
With incorrect instrument usage:
image
Not sure how to go about fixing that but I'm sure it would be useful if users don't want the LLM to work if the trace isn't going to work


packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -55,6 +55,7 @@ def __new__(
         processor: SpanProcessor = None,
         propagator: TextMapPropagator = None,
         exporter: SpanExporter = None,
+        instruments=[]
```
nirga, line 58: Default value should be `None` otherwise you'll never go into the if at Line 113

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -108,7 +109,64 @@ def __new__(
             if propagator:
                 set_global_textmap(propagator)
 
-            init_instrumentations()
+            instrument_set = False
+            if not instruments:
+                init_instrumentations()
+                instrument_set = True
+            else:
+                if 'openai' in instruments:
```
nirga, line 117: I suggest we use an enum for this (it can be an enum with string values), so the user can choose values from an enum instead of guessing whether we called it `openai` or `OpenAI`.

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -108,7 +109,64 @@ def __new__(
...
+                    init_openai_instrumentor()
+                    instrument_set = True
+                if 'anthropic' in instruments:
+                    init_anthropic_instrumentor()
```
nirga, line 121: Potentially, the `init_` calls can return a boolean specifying whether they were able to instrument (for example, whether the `anthropic` library exists) and if not - log a warning.

nirga (commented):
Thanks so much for this! This looks overall great, I made a couple of comments and also responded to your questions.

Also, nake sure to run formatter + linter on your code.

Mizba-Anjum:
I made some changes and added an enum, now the format requires an extra import for the instruments like this:
```
from traceloop.sdk impot Traceloop, Instruments
Traceloop.init(instruments=[Instruments.OPENAI.value, Instruments.ANTHROPIC.value])
```
This will initialize the chosen instruments. To initialize all, leave the array empty or leave out the instruments argument. Not importing the Instruments class also causes it to initialize all.

I did have the option of using a for loop to go through them like `for instrument in instruments: instruments.value == etc` which would involve having to check for duplicates and would also throw an error for invalid inputs unlike the current solution. Let me know if that's something you'd like or if you prefer it this way.

I also added booleans to each of the init_ functions and if they don't return True, a warning will appear "x library does not exist".

packages/traceloop-sdk/traceloop/sdk/__init__.py
```diff
@@ -47,6 +48,7 @@ def init(
         propagator: TextMapPropagator = None,
         traceloop_sync_enabled: bool = True,
         resource_attributes: dict = {},
+        instruments=None
```
nirga, line 51: Add type to the argument

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -108,7 +110,153 @@ def __new__(
...
+                if 'anthropic' in instruments:
+                    if not init_anthropic_instrumentor():
+                        print(
+                            Fore.RED
+                            + "Warning: Anthropic library does not exist."
+                        )
+                        print(Fore.RESET)
+                    else:
+                        instrument_set = True
+                if 'cohere' in instruments:
```
nirga, line 135: Use the enum values instead of the string

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -272,6 +420,7 @@ def init_instrumentations():
     init_replicate_instrumentor()
     init_vertexai_instrumentor()
     init_watsonx_instrumentor()
+    return True
```
nirga, line 423: Why is this needed?

nirga (commented):
Thanks! Left a couple of more small comments

packages/traceloop-sdk/traceloop/sdk/__init__.py
Mizba-Anjum: not sure about this one, should it be something like `array = None or []`? sorry if it's a dumb question

nirga: No worries :) 
I mean `instruments: Set[Instruments] = {}`

nirga:
Sorry given your `if` sentence the default value should be `None`

Mizba-Anjum:
I added a for loop instead of multiple 'if x in y' statements for time complexity and to catch invalid arguments.

I also created a separate instruments file instead of importing it from `__init__` because when trying to import it into `tracing.py` it created a circular import error where `tracing` was also imported into `__init__`. I hope that's alright, if not, let me know.

The `return True` in the _init function for all packages was initially for verification like all the separate packages, but I removed it anyway.

Currently, implementation is like this:
 ```
from traceloop.sdk.instruments import Instruments

Traceloop.int(instruments=[Instruments.OPENAI, Instruments.ANTHROPIC])
```

Mizba-Anjum:
> Sorry given your `if` sentence the default value should be `None`

I do believe that an empty set or array would still trigger the `if not` statement. The testing worked on my end when using empty array/set. Or is there another reason it needs to be `None`?

nirga:
I've triggered the tests here and looks like they're failing so there's an issue in your code I think.

Regarding the code itself - it's really a nit but I think the previous way was actually more readable - even if a bit less efficient. Python isn't an efficient language anyway so I'd prioritize readability. But that's really a nit so don't feel obligated to revert it.

Reg the None - IIRC python returns true if you test an empty set because it's a value. I also think it makes more sense that None just instruments everything but an empty array instruments nothing, but again - personal preference.

Mizba-Anjum:
I changed the set back to a None, let me know if there are any other changes needed or if you'd like, I could revert the for loop back into `in` checks (though that would also take away the check for garbage inputs).

Mizba-Anjum:
> I also think it makes more sense that None just instruments everything but an empty array instruments nothing, but again - personal preference.

I noticed that you said this and currently the code does instrument everything on an empty array/set as well but I can make a change to make it do as you said.

nirga:
@Mizba-Anjum no that's ok. I just added the typing annotations I mentioned and I'll merge. Thanks so much for this contribution!

nirga (approved):

Issue #250, opened 2023-12-15
🚀 Feature: embedding vectors in attributes of vector DB calls


nirga:
### Which component is this feature for?

All Packages

### 🔖 Feature description

We need to extract and report embedding vectors that are sent and returned for `query` calls to vector DBs like Chroma, Weaviate, Pinecone, etc.

### 🎤 Why is this feature needed ?

More visiblity into what's happening when calling a vector DB

### ✌️ How do you aim to achieve this?

This isn't straightforward - these might be large, so sending them as attributes might not make sense and we'll need to think of the proper OTEL format to use (logs? [span events](https://opentelemetry.io/docs/concepts/signals/traces/#span-events)?)

### 🔄️ Additional Information

Consult on [slack](traceloop.com/slack) before working on it.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

nirga:
@quest-bot stash 100

paolorechia:
@quest-bot embark

paolorechia:
Hi, @nirga 

I started working on this (still figuring my way around the codebase :)), I opened a draft PR where I made some questions. Could you please check it out? https://github.com/traceloop/openllmetry/pull/344

Thanks!

PR #368, opened 2024-01-26
feat(pinecone): instrument pinecone query embeddings

paolorechia:
Second PR for issue: https://github.com/traceloop/openllmetry/issues/250
Implements embeddings instrumentation for pinecone database.

* [full trace example](https://gist.github.com/paolorechia/853d767dfd84bfd5618412744595b08a)

Useful additions:
* Adds new attributes to pinecone.query span, for instance:
```json
        "pinecone.query.top_k": 3,
        "pinecone.query.include_values": true,
        "pinecone.query.include_metadata": true
```

* Adds 3 new events to the query trace:
1. `query.embeddings.i`:
```json
    "events": [
        {
            "name": "vector_db.query.embeddings.0",
            "timestamp": "2024-01-26T23:10:52.677797Z",
            "attributes": {
                "vector_db.query.embeddings.0.vector": [
                    -0.08285148441791534,
                    0.0602964386343956,
                    (...)
```
2. `query.usage`
```json
        {
            "name": "pinecone.query.usage",
            "timestamp": "2024-01-26T23:10:52.904263Z",
            "attributes": {
                "readUnits": 6
            }
        },
```
3. `query.result.i`:
```json
        {
            "name": "vector_db.query.result.0",
            "timestamp": "2024-01-26T23:10:52.904605Z",
            "attributes": {
                "vector_db.query.result.0.id": "id1",
                "vector_db.query.result.0.score": 0.922191083,
                "vector_db.query.result.0.metadata": "{'description': 'This is for text of index 1'}",
                "vector_db.query.result.0.vector": [
                    -0.0902868807,
                    0.0328526832,
                    0.00385046285,
                    0.0246606264,
                   (...)
```


paolorechia:
Added three new event attributes that may consider for semantic conventions:
```txt
                "vector_db.query.result.0.id": "id1",
                "vector_db.query.result.0.score": 0.922191083,
                "vector_db.query.result.0.vector"
```

paolorechia:
Note: I have not yet executed the unit tests as I don't have an OpenAI API Key handy. Will do this in the following days.

paolorechia:
@nirga if you have time, please take a first look at this PR 
(and if are able to run the pinecone tests with just a click, maybe try it out - I doubt I'd get it right on the first try though :))

paolorechia:
@nirga I didn't find any other VectorDBs that are currently implemented for me to continue the work. I could work on Weviate instrumentation, but since there's nothing there it seems like a separate issue to me.

I updated the unit test with the correct key (`pinecone.query.result.usage`), but I'm still lacking credits in my OpenAI account. I could purchase some next week, but in the meantime I don't have an easy way to test it.

I could also implement a test using sentence transformers embeddings, but the library is quite heavy on dependencies, so I'm not sure if it's a good idea.

paolorechia:
@quest-bot loot #250

paolorechia:
@nirga, I had another go at this today.

I updated my OpenAI account, so I was able to run the tests and the original sample app locally, so I made the tests pass.

Additionally, I decided to change again (:sweat_smile: ) the attributes here.  I made it so it matches what I'm doing here: https://github.com/traceloop/openllmetry/pull/370

In summary, the changes consist in removing the index from the semantic event attributes and using `pinecone` namespace for everything that is specific to Pinecone, as it was requested in the other PR. 

Here's the [full exported trace](https://gist.github.com/paolorechia/b9b1fb8acc6664218cef2f8c4f991036)

Some interesting parts:

```json
    "events": [
        {
            "name": "db.query.embeddings",
            "timestamp": "2024-01-31T17:02:01.483501Z",
            "attributes": {
                "db.query.embeddings.vector": [
                    -0.027747908607125282,
                    -0.02155691757798195,
                    0.03326960280537605,
                    -0.0021926425397396088,
```

```json
                -0.0012348864693194628,
                    -0.025084666907787323,
                    -0.03232143446803093
                ]
            }
        },
        {
            "name": "pinecone.query.usage",
            "timestamp": "2024-01-31T17:02:01.745058Z",
            "attributes": {
                "readUnits": 6
            }
        },
        {
            "name": "db.pinecone.query.result",
            "timestamp": "2024-01-31T17:02:01.745328Z",
            "attributes": {
                "db.pinecone.query.result.id": "WS1uVMGhlWQ-t747.92",
                "db.pinecone.query.result.score": 0.863803566,
                "db.pinecone.query.result.metadata": "{'channel_id': 'UCv83tO5cePwHMt1952IVVHw', 'end': 906.0, 'published': datetime.datetime(2021, 10, 20, 17, 6, 20, tzinfo=tzutc()), 'start': 747.0, 'text': \"pooling approach. Or we can't use it in its current form. Now the solution to this problem was introduced by two people in 2019 Nils Reimers and Irenia Gurevich. They introduced what is the first sentence transformer or sentence BERT. And it was found that sentence BERT or S BERT outformed all of the previous Save the Art models on pretty much all benchmarks. Not all of them but most of them. And it did it in a very quick time. So if we compare it to BERT, if we wanted to find the most similar sentence pair from 10,000 sentences in that 2019 paper they found that with BERT that took 65 hours. With S BERT embeddings they could create all the embeddings in just around five seconds. And then they could compare all those with cosine similarity in 0.01 seconds. So it's a lot faster. We go from 65 hours to just over five seconds which is I think pretty incredible. Now I think that's pretty much all the context we need behind sentence transformers. And what we do now is dive into a little bit of how they actually work. Now we said before we have the core transform models and what S BERT does is fine tunes on sentence pairs using what is called a Siamese architecture or Siamese network. What we mean by a Siamese network is that we have what we can see, what can view as two BERT models that are identical and the weights between those two models are tied. Now in reality when implementing this we just use a single BERT model. And what we do is we process one sentence, a sentence A through the model and then we process another sentence, sentence B through the model. And that's the sentence pair. So with our cross-linked we were processing the sentence pair together. We were putting them both together, processing them all at once. This time we process them separately. And during training what happens is the weights within BERT are optimized to reduce the difference between two vector embeddings or two sentence\", 'title': 'Intro to Sentence Embeddings with Transformers', 'url': 'https://youtu.be/WS1uVMGhlWQ'}",
                "db.pinecone.query.result.vector": [
                    -0.0165825691,
                    -0.0046767923,
                    0.0124068623,
```

Also tested briefly on the frontend:
image

paolorechia:
@nirga I've applied our last discussion from #370, modifying both Pinecone and ChromaDB into an unified format. I had to copy the new semantic classes across the packages, but we can delete the duplicates and import on a follow-up PR (once semantic-conventions is bumped).

nirga (approved):

PR #370, opened 2024-01-27
chore(semconv): add new conventions from Pinecone query result

paolorechia:
Proposes to extend semantic-conventions-ai with event attributes determined by https://github.com/traceloop/openllmetry/pull/368


paolorechia:
@nirga please take a look at this one as well :)

nirga:
Thanks @paolorechia! I wonder if we should have complete distinct attributes for chroma and pinecone. I'd rather we have the same attributes for both of them, and then potentially if there's something specific for chroma add just one attribute for that (which is similar to what I'm trying to do [here](https://github.com/open-telemetry/semantic-conventions/pull/639/files#diff-35817cb0d942a77be23554e66f9efbd67146f9f0b1307f1bdad66e778b44f034) for LLMs). WDYT?

paolorechia:
@nirga In principle, I agree.

As we add support for more databases, what is common between different implementations starts to emerge.

Maybe a more efficient development process is to start using the vendor namespace, and then once common attributes are identified, refactor them.

From the implementation, perspective, would you put these vendor specific attributes inside the same package/file currently used? E.g., semantic-conventions-ai?

I’d gladly update this PR along the week.

On a somewhat related topic, playing with different conventions in the code level is cumbersome (because we need to always merge / bump / release).

Any idea how we could simplify this for local development?

nirga:
Reg. Simplifying the dev process - yes. You just need to depend on the semantic convention package directly in the poetry file (instead of on a published version as it is today). 

I'm fine with refactoring later, whatever works for you best. But if there are things that for sure are the same - maybe we can use the same semantic convention from day 1

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv/ai/__init__.py
```diff
@@ -53,12 +53,17 @@ class EventAttributes(Enum):
     # Query Embeddings
     VECTOR_DB_QUERY_EMBEDDINGS_VECTOR = "vector_db.query.embeddings.{i}.vector"
 
-    # Query Result
+    # Query Result (from ChromeDB)
```
galkleinman, line 56:
```suggestion
    # Query Result (from ChromaDB)
```

galkleinman (commented):
Hey @paolorechia,

Gal from Traceloop here.

Attaching here the [database semantic conventions](https://opentelemetry.io/docs/specs/semconv/database) of opentelemetry for the inspiration, we would like to stick to their patterns and do as closer implementation as we can for the vector dbs.

My suggestions after going over it:

1. Every attribute should start with `db` rather than `vector_db`
2. The shared attributes of vector databases should be prefixed by `db.vector` (see the sql databases example in the attached docs). Regarding finding these ones - I totally get your iterative approach for merging the shared ones later - it's a great agile development practice, having said that it introduce some sort of breaking changes, therefore I agree with @nirga and suggest a small research of let's say ChromaDB, Pinecone and Weaviate, to extract to common attrs.
3. The DB specific ones should be prefixed with `db.chromadb`, `db.pinecone` ....
4. Make sure to report the generic DB attributes introduced by opentelemetry in the docs I attached also for the vector db.

Glad to discuss everything, either here or in our community slack. Feel free to reach out. 💪 💪

paolorechia:
@galkleinman thanks for the review! Once I get some time I’ll get back to this topic this week. 

We’re just missing weaviate example here to decide on the first set of common attributes.

Maybe it makes sense to first implement the basic weviate instrumentation, and also have a test setup for it? Do we have an open issue for that?

paolorechia:
@galkleinman please check if the new format is better now:

1. I'd applied the changes you requested (as I understood it)
2. I removed the indexing, since I think it was a bad idea I had, and looks cleaner without indexing.

Challenge here: this will cause breaking changes in code, so requires changes in other packages. I couldn't get `semantic-conventions-ai` package installed locally to be correctly imported, I end up getting a `ModuleNotFound`  error :/

Regarding Weaviate, like I mentioned, I'll play a bit with it's instrumentation and then propose some attributes for it here.

paolorechia:
@nirga, @galkleinman 

The best way I could update the ChromaDB with working tests in the same PR is to avoid using the Enum classes directly, and duplicate the strings there.

It would require a second PR to replace the hardcoded attributes with the Enum attributes, but at least no tests will be in a broken state.
  
Let me know what you think.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv/ai/__init__.py
```diff
@@ -45,19 +45,25 @@ class SpanAttributes:
 
 
 class Events(Enum):
-    VECTOR_DB_QUERY_EMBEDDINGS = "vector_db.query.embeddings"
-    VECTOR_DB_QUERY_RESULT = "vector_db.query.result"
+    DB_QUERY_EMBEDDINGS = "db.query.embeddings"
+    DB_PINECONE_QUERY_RESULT = "db.pinecone.query.result"
```
nirga, line 49: why not use the same name for both of them?

nirga: How do the differ?

paolorechia: I mean these attributes:
```
    # Query Result (from ChromaDB)
    DB_CHROMADB_QUERY_RESULT_IDS = "db.chromadb.query.result.ids"
    DB_CHROMADB_QUERY_RESULT_DISTANCES = "db.chromadb.query.result.distances"
    DB_CHROMADB_QUERY_RESULT_METADATA = "db.chromadb.query.result.metadata"
    DB_CHROMADB_QUERY_RESULT_DOCUMENTS = "db.chromadb.query.result.documents"

    # Query Result (from Pinecone)
    DB_PINECONE_QUERY_RESULT_ID = "db.pinecone.query.result.id"
    DB_PINECONE_QUERY_RESULT_SCORE = "db.pinecone.query.result.score"
    DB_PINECONE_QUERY_RESULT_VECTOR = "db.pinecone.query.result.vector"

```

In summary, ChromaDB returns attributes of type list, and Pinecone returns a different type per each attribute (str, float, list).

paolorechia: We could take some actions to reduce the specificity, e.g., make pinecone also use lists. Then we can reuse some fields (ids, metadata).

In theory, we could also unify score / distance somehow, but it might be misleading, as score is the complement of the distance, I think this is the rule we could apply to pinecone:

`distance = 1 - score`

Or this to ChromaDB:

`score = 1 - distance`

But it can be confusing as it’s then a derived metric instead of a raw one.

nirga: I'd keep the difference between the `distance` and `score`, but I like the rest.

paolorechia: I mean, for consistency reasons we could always use type list here. Even for attributes that are only present in a specific DB. What do you think?

nirga: Maybe I'm missing something. Can you explain what does each attribute mean in each context?

paolorechia: Hey, @nirga sorry for the delay answering you. I needed a short break this topic :)

Let's go over this in detail we have these attributes in chroma:

```
    DB_CHROMADB_QUERY_RESULT_IDS = "db.chromadb.query.result.ids"
    DB_CHROMADB_QUERY_RESULT_DISTANCES = "db.chromadb.query.result.distances"
    DB_CHROMADB_QUERY_RESULT_METADATA = "db.chromadb.query.result.metadata"
    DB_CHROMADB_QUERY_RESULT_DOCUMENTS = "db.chromadb.query.result.documents"
```

The response format of Chroma comes as an object of lists
```
{
   ids: [1, 2, 3...],
   distances: [0.3, 0.5, 0.6...],
   metadata: ["some metadata text", "another metadata text",...],
   documents: ["retrieved text", "retrieved text2", ...]
}
```
Therefore all these fields are stored as lists. Either of numbers or strings. 

ID represents the stored document ID in ChromaDB.
Distance is how close the retrieved document is assessed to be to the query text.
Metadata is well, metadata.
And document is what the user originally stored beforehand in the Database.

Now let's take a look at Pinecone:
```
    # Query Result (from Pinecone)
    DB_PINECONE_QUERY_RESULT_ID = "db.pinecone.query.result.id"
    DB_PINECONE_QUERY_RESULT_SCORE = "db.pinecone.query.result.score"
    DB_PINECONE_QUERY_RESULT_VECTOR = "db.pinecone.query.result.vector"
```

This would come like this:

```
[
    {"id": 1, "score": 0.9, vector: [0.1, 0.2, 0.3, 0.9, -0.3,...]} "metadata": {"field2": "value1"},
    {"id": 2, "score": 0.8, vector: [0.2, 0.1, 0.4, 0.7, -0.4,...]}, "metadata": {"field3": "value2"},
    {"id": 3, "score": 0.7, vector: [0.3, 0.0, -0.3, -0.9, 0.3,...], "metadata": {"field1": "value1"}},
]
```
As you can see, ID is basically the same thing as in ChromaDB, but the query comes back in slightly different format.
The way I built the original Pinecone PR is to follow the original response structure from Pinecone query result. What we could do instead, is force the trace to follow a specific convention, e.g., make Pinecone trace result look a bit like ChromaDB result.

For instance, if we convert Pinecone result into Chroma format:
```
{
   ids: [1, 2, 3...],
   score: [0.9, 0.8, 0.6...],
   metadata: ["{\"field2\": \"value1\"}",  {\"field3\": \"value2\"},...],
   vectors: [[0.1, 0.2, 0.3, 0.9, -0.3,...], [0.2, 0.1, 0.4, 0.7, -0.4,...], [0.3, 0.0, -0.3, -0.9, 0.3,...]]
}
```

We could also go into the other direction and make Chroma look like Pinecone:

```
[
    {"id": 1, "distance": 0.3,  "document": "retrieved text", "metadata": "some metadata text",
    {"id": 2, "distance" 0.5, , "document": "retrieved text2": "another metadata text",
    {"id": 3, "distance": 0.6, "document": ..., "metadata": ...
]
```
I don't know if that helps or makes things more confusing for users! But if we do this, then we can reuse more fields across different databases. We just need to define which is the canonical format.

paolorechia: Forgot to explain score and vectors attributes from Pinecone, but I think these should be clear by now.

Score is the complement of distance, e.g, how good the query matches the document, but here higher is better.
And vectors are the resulting vectors. These should be the same idea across databases, the only difference here is how the result is presented (again, a list in ChromaDB, or an attribute in Pinecone).

nirga:
I think it's already covered by #368, right?

PR #413, opened 2024-02-10
chore: refactor testing to use mock HTTP responses

nirga:
Move tests for specific instrumentations to be under that instrumentation directory, and refactor to use [VCR.py](https://vcrpy.readthedocs.io/en/latest/index.html)

- [x] OpenAI
- [x] Anthropic
- [x] Cohere
- [x] VertexAI
- [x] Replicate
- [x] Langchain
- [x] Haystack
- [x] Chroma
- [x] Pinecone


galkleinman (approved):
LGTM.

While I believe mocking is a much better option here, it will be far more complex to implement, and the suggested solution will introduce tremendous improvement in the short term.

PR #421, opened 2024-02-13
fix(openai): handle async streaming responses for openai v1 client

najork:


packages/opentelemetry-instrumentation-openai/tests/test_completions.py
```diff
@@ -20,6 +20,25 @@ def test_completion(exporter, openai_client):
...
+    spans = exporter.get_finished_spans()
+    assert [span.name for span in spans] == [
+        "openai.completion",
+    ]
+    open_ai_span = spans[0]
+    assert (
+        open_ai_span.attributes["llm.prompts.0.user"]
+        == "Tell me a joke about opentelemetry"
+    )
+    assert open_ai_span.attributes.get("llm.completions.0.content")
```
najork, line 39: I'm happy to split this into it's own PR, figured some test coverage for non-streaming async calls would also be beneficial.

nirga:
Thanks @najork! Looks like the tests aren't actually running since we're missing some pytest async plugin (see test run logs). 
When they do run - you'll need to run them locally. Here's how to do that:
1. Set the environment variable `OPENAI_API_KEY` with an actual API key.
2. Comment out [this line](https://github.com/traceloop/openllmetry/blob/4dab1142636b4e75a82e5e51bf1491fefed193a2/packages/opentelemetry-instrumentation-openai/tests/conftest.py#L36)
3. Run `poetry run pytest -v --record-mode=once` so that we'll have a recording of the requests + responses for future tests. Don't worry, this does not log your API key in any way, just the response completion you got from OpenAI so that the test can be run without actually calling OpenAI.

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/completion_wrappers.py
```diff
@@ -60,22 +60,23 @@ async def acompletion_wrapper(tracer, wrapped, instance, args, kwargs):
     if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
         return wrapped(*args, **kwargs)
 
-    async with start_as_current_span_async(
```
najork, line 63: This was causing `span.is_recording()` to return false in `_abuild_from_streaming_response`, in turn causing the response attribute setter to short-circuit, and the test assertion for span content to fail

nirga (approved):

PR #472, opened 2024-02-22
fix(sdk): fail gracefully in case input/output serialization failure

nirga:


philipesteiff:
hey @nirga, I'm trying to use `@atask()` and getting `TypeError: Object of type Foo is not JSON serializable`.

It seems the same `pass` logic need to be applied on the async task as well.

nirga:
Right! https://github.com/traceloop/openllmetry/pull/525

PR #523, opened 2024-02-26
Only accumulate text if it exists

anjor:
I was getting this error:

```
> complete_choice["text"] += choice.get("text") E TypeError: can only concatenate str (not "NoneType") to str
```


nirga:
Thanks @anjor nice catch :)
There's an equivalent line in the `chat_wrappers.py` for streaming API there so maybe we can fix it there as well

anjor:
Thanks! :)

https://github.com/traceloop/openllmetry/blob/ac3a41832e775a2dc8654b2c19078c52e5477394/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L197-L198

That already seems to have the check?

nirga (approved):

nirga:
@anjor released v0.12.3

PR #526, opened 2024-02-26
chore: temporarily disabled watsonx tests

nirga:


nirga:
FYI @gyliu513 I needed to disable watson's tests as the requirement of Python >= 3.10 is conflicting with the SDK allowing Python >= 3.8.1

Is there a version of watson that supports Python 3.8?

PR #533, opened 2024-02-26
fix(llamaindex): instrument agents & tools

nirga:


packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/base_agent_instrumentor.py
```diff
@@ -0,0 +1,62 @@
...
+from wrapt import wrap_function_wrapper
+
+from opentelemetry.instrumentation.llamaindex.utils import (
+    _with_tracer_wrapper,
+    start_as_current_span_async,
+)
+from opentelemetry.semconv.ai import SpanAttributes, TraceloopSpanKindValues
+
+
+V9_MODULE_NAME = "llama_index.agent.types"
```
paolorechia, line 12: :question: Sorry for my ignorance, what are these versions? Llama Index versions?

packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/base_tool_instrumentor.py
```diff
@@ -0,0 +1,62 @@
+from importlib.metadata import version as package_version, PackageNotFoundError
```
paolorechia, line 1: :upside_down_face: (NIT) There is a lot of duplication here. Maybe it could be refactored sometime.

paolorechia (approved):
Some nitpicks / questions, but looks good otherwise.

paolorechia:
Just noticed there's a failing test on 3.11 because of pydantic

packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/base_agent_instrumentor.py
nirga: Yes, they re-arrenged their packages and I want to support both the old version, the new version, and another location of the module in the new version (this is the legacy module).

packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/base_tool_instrumentor.py
nirga: Yeah I know :/ I need to refactor this

Issue #542, opened 2024-02-28
🐛 Bug Report: Watsonx traceloop flow has many http request spans which is distracting


huang-cn:
### Which component is this bug for?

Watsonx Instrumentation

### 📜 Description

when using traceloop flow with watsonx, the trace data include many http request spans, which is distracting.  

### 👟 Reproduction steps

following the [traceloop openllmetry example](https://www.traceloop.com/openllmetry/integrations/observability-for-openai-with-traceloop) to instrument watsonx LLM, and output otel data to jaeger.

### 👍 Expected behavior

plan for improvement:

1. adding these domain names to traceloop-sdk tracing `EXCLUDED_URLS` to avoid generating http request spans
2. create a watson model init span to show the model initialization response time.

### 👎 Actual Behavior with Screenshots

![image](https://github.com/traceloop/openllmetry/assets/58250676/84886351-b29f-4cf0-be75-f7ec5346132f)

### 🤖 Python Version

3.11

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

paolorechia:
My first impression is that this is not a bug but intended behavior. That being said, making it more configurable is always a good idea

huang-cn:
@paolorechia It appears there's no simple switch to turn these http request spans off,  in the sdk tracing `EXCLUDED_URLS` we can see URLs of many api have been excluded already, we're planning to add the Watsonx api to the `EXCLUDED_URLS` to address this.

paolorechia:
I see, I found it (https://github.com/traceloop/openllmetry/blob/3eba03e21b99cfd9b09bc46cdd1d009634f43072/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L35)

Makes sense to me, I agree you can consider it a bug then 🙂

paolorechia:
@nirga While researching for semantic conventions I’ve seen some OpenTelemetry  instrumentations define trace level.

Maybe it’s a feature to consider for traceloop, so users can adjust these kind of things?

nirga:
@huang-cn yep, sounds good!
@paolorechia that's a good question, which span kind to use. We're starting a semantic convention working group to discuss that - https://github.com/open-telemetry/community/blob/main/projects/llm-semconv.md

Issue #529, opened 2024-02-26
🐛 Bug Report: Validate if watsonx can run with python 3.8


gyliu513:
### Which component is this bug for?

Watsonx Instrumentation

### 📜 Description

This is a follow up for https://github.com/traceloop/openllmetry/pull/526, need check if watsonx support python 3.8

### 👟 Reproduction steps

Check https://github.com/traceloop/openllmetry/pull/526 for detail

### 👍 Expected behavior

Watsonx need to work with python 3.8

### 👎 Actual Behavior with Screenshots

Check https://github.com/traceloop/openllmetry/pull/526

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

huang-cn:
@gyliu513 the `ibm-watsonx-ai` package requires ["Python >=3.10"](https://pypi.org/project/ibm-watsonx-ai/0.1.8/) since version `0.1.1`, I think that's the only reason preventing us from using Python 3.8.

gyliu513:
@nirga with above statement, I think Watsonx cannot support python 3.8. So for now, I think we have to disable the test until the SDK want to bump up python to 3.10, does that make sense? Thanks

nirga:
@gyliu513 @huang-cn can we maybe use v0.1.0 in the tests then?

huang-cn:
@nirga @gyliu513 there's only a v0.0.5 which is the earliest version can support Python 3.9, unfortunately that version doesn't has the `ModelInference` yet, we can not use it.

nirga:
@huang-cn since we run tests on multiple python versions, maybe we can run then on Python 3.9+ only?

huang-cn:
@nirga unfortunately the v0.0.5 is a very old version we are not able to use, for all the other versions of `ibm-watsonx-ai` they all require [Python >= 3.10](https://pypi.org/project/ibm-watsonx-ai/#history).  
is it possible we only run tests on Python 3.10+?

PR #543, opened 2024-02-28
fix(watsonx): exclude http request, adding span for model initialization

huang-cn:
  
issue related discussions in https://github.com/traceloop/openllmetry/issues/542
- exclude watsonx api URLs in `EXCLUDED_URLS` in traceloop-sdk tracing
- add a watson model_init span for model initialization response time.
- update test to enable local test, but will skip in CI for now, due to issue https://github.com/traceloop/openllmetry/issues/529

### DIFF
--- a/packages/opentelemetry-instrumentation-watsonx/tests/conftest.py
+++ b/packages/opentelemetry-instrumentation-watsonx/tests/conftest.py
@@ -20,7 +20,14 @@ def exporter():
     provider.add_span_processor(processor)
     trace.set_tracer_provider(provider)
 
-    WatsonxInstrumentor().instrument()
+    try:
+        # Check for required package in the env, skip test if could not found
+        from ibm_watsonx_ai.foundation_models import ModelInference
+        # to avoid lint error
+        del ModelInference
+        WatsonxInstrumentor().instrument()
+    except ImportError:
+        print("no supported ibm_watsonx_ai package found, Watsonx instrumentation skipped.")
 
     return exporter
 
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-instrumentation-watsonx/tests/conftest.py
```diff
@@ -20,7 +20,14 @@ def exporter():
     provider.add_span_processor(processor)
     trace.set_tracer_provider(provider)
 
-    WatsonxInstrumentor().instrument()
+    try:
+        # Check for required package in the env, skip test if could not found
+        from ibm_watsonx_ai.foundation_models import ModelInference
```
nirga, line 25: don't we need to add them to poetry otherwise they won't get installed in the test?

huang-cn: @nirga I'm not sure if we add the `ibm-watsonx-ai` import in poetry, would it cause issue for test on 3.8 and 3.9? is there a way to set it to only run test on 3.10+?

packages/opentelemetry-instrumentation-watsonx/tests/test_generate.py
```diff
@@ -1,17 +1,17 @@
 import pytest
+import sys
 
 
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="ibm-watson-ai requires python3.10")
```
paolorechia, line 5: :+1: nice way to skip it

packages/opentelemetry-instrumentation-watsonx/tests/conftest.py
paolorechia: @nirga test execution in CI looks good to me. It skips watsonx test on Python < 3.10, but runs normally otherwise. Traceloop SDK tests are running for all versions.
I think this is a good solution.

paolorechia (approved):

packages/opentelemetry-instrumentation-watsonx/tests/conftest.py
nirga: @paolorechia but how does it run if the Watson package is not installed? 🤔

paolorechia: Aha! It's only running the try/catch, and exiting on the except then? I missed that.

nirga: Yep 😅

paolorechia: OK, but at least the @huang-cn is able to run tests locally now. 

Do you think we might need to revive the optional dependencies issue I created but closed? Maybe it's a way forward for these issues. I don't think it's reasonable to request the PR author to implement a fix for this though.

nirga: Yeah let's merge it, open an issue and figure out if there's a way around it within poetry

Issue #551, opened 2024-03-01
🚀 Feature: Watsonx streaming generate support


huang-cn:
### Which component is this feature for?

Watsonx Instrumentation

### 🔖 Feature description

support Watsonx streaming generate.

### 🎤 Why is this feature needed ?

so the Watsonx streaming generation/completion span could include LLM response data.

### ✌️ How do you aim to achieve this?

parsing streaming response data to span attributes.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #552, opened 2024-03-01
fix(watsonx): Watsonx stream generate support

huang-cn:
- [X] (If applicable) I have updated the documentation accordingly.

Fixes https://github.com/traceloop/openllmetry/issues/551


huang-cn:
Here is a screenshot in Jaeger UI:  
image

nirga:
Thanks @huang-cn! Can you try and run the tests locally by manually installing IBM's packages?

huang-cn:
@nirga Sure. I'd run tests locally for both unit test and E2E test before raising the PR, that's the reason we construct the test scripts that way, to allow local tests to work.

nirga (approved):

PR #585, opened 2024-03-06
fix(openai): streaming metrics

### LINKED ISSUE #584: 🚀 Feature: Log time between first and last token of Open ai streaming response

OpenAI Instrumentation

### 🔖 Feature description

 Have a way to Log time b/w first token and last token for each OpenAI streaming API call.
 Also good to have one for "time to first token".

The primary requirement would be to show it along with the other params like duration, etc.
Worth wile to see if we have a param that show if the request is of type streaming or not.

To track the time spent while streaming response.

Have a field along with the duration field. That shows:
1. Time difference b/w first and last token.
2. Time to generate first token.

Also it would be good to have a metric for it.

### 🔄️ Additional Information

_No response_

None

nirga:
Towards #584


mohitk1995 (commented):
Will this change automatically add a variable as part of the trace as well?
Apart from that LGTM.

nirga:
> Will this change automatically add a variable as part of the trace as well? Apart from that LGTM.

No, I will do it in a separate PR

PR #661, opened 2024-03-19
fix(openai): Support report token usage in stream mode

### LINKED ISSUE #627: 🐛 Bug Report: Number of tokens not reported when streaming OpenAI/AOAI

OpenAI Instrumentation

### 📜 Description

`llm.usage.prompt_tokens` and `llm.usage.total_tokens` are not reported when streaming is used. Per discussion in Slack, this may be because this information is not included in the response from OpenAI/Azure OpenAI. Especially when `TRACELOOP_TRACE_CONTENT` is `false`, this is an important statistic and would be required for derived fields like token throughput.

### 👟 Reproduction steps

The following code correctly reports the number of tokens:
```
prompt = "Say Hi!"

response_non_streaming = openai.ChatCompletion.create(
    model="gpt-3.5-turbo",
    messages=[{"role": "user", "content": prompt}]
)
```

The following code does not report the number of tokens:
```
prompt = "Say Hi!"

response_streaming = openai.ChatCompletion.create(
    model="gpt-3.5-turbo",
    messages=[{"role": "user", "content": prompt}],
    stream=True
)
```

### 👍 Expected behavior

`llm.usage.prompt_tokens` and `llm.usage.total_tokens` should be reported regardless of `stream` setting.

Example with non-streamed ada:
image

Example with streaming gpt-3.5:
image

### 🤖 Python Version

_No response_

If this value is not reported by OpenAI/Azure OpenAI directly, then it may be a design decision of whether it's worth counting in this library, or whether only values returned by the APIs should be reported.

None

Humbertzhang:
Fixes #627 

* For Traces:
image

* For Metrics:
image

Related doc PR: https://github.com/traceloop/docs/pull/24


Humbertzhang:
CI error:
```
  E               vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/home/runner/work/openllmetry/openllmetry/packages/opentelemetry-instrumentation-openai/tests/traces/cassettes/test_chat/test_chat_streaming.yaml') in your current record mode ('none').
  E               No match for the request (<Request (GET) https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken>) was found.
```

tiktoken need to download file at it first run, so vcr's none record mode maybe not suite for it...

nirga:
@Humbertzhang you can tell VCR to ignore certain hostnames:
https://github.com/traceloop/openllmetry/blob/373f5c72269362422d86e4a1f4644661fe95401c/packages/opentelemetry-instrumentation-llamaindex/tests/conftest.py#L50

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/__init__.py
```diff
@@ -205,3 +229,29 @@ def model_as_dict(model):
         return model_as_dict(model.parse())
     else:
         return model
+
+
+def should_record_stream_token_usage():
```
nirga, line 234: Let's put it as an optional parameter to the `__init__` of the instrumentation (I'd default it to `true`)? I think it's cleaner (and I want to move the `should_trace_content` to be like that as well)

Humbertzhang: You mean move it to `Traceloop.init()`'s arg, right?

nirga: No, the instrumentation init. Then traceloop sdk can pass it yes

Humbertzhang:
Hello @nirga, I have resolved the VCR issue by adding ignore_hosts. However, I'm not entirely sure what you mean by `"put it as an optional parameter to the __init__ of the instrumentation."`

Could you provide me an example? Alternatively, we could just merge this, and I can make modifications in a later PR after you've updated `should_trace_content`.

nirga (approved):

nirga:
@Humbertzhang sure, thanks!

PR #662, opened 2024-03-21
fix: auto-instrumentation support

### LINKED ISSUE #591: 🚀 Feature: Make instrumentations compatible with "opentelemetry-instrument" automatic instrumentation

All Packages

### 🔖 Feature description

When starting a python app via "opentelemetry-instrument python myapp.py" available instrumentors are automatically discovered and used for tracing. This works with official opentelemetry instrumentors, but not with the instrumentors created by traceloop. Instrumentors created by traceloop should behave the same way as the opentelemetry instrumentors so that they can be used in automatic instrumentation without having to change the code of the python app which is being observed.

Being able to use automatic instrumentation would remove the requirement of changing code to add observability to apps.

The "official" instrumentations by opentelemetry have the following lines in their [pyproject.toml](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml):
```
[project.entry-points.opentelemetry_instrumentor]
httpx = "opentelemetry.instrumentation.httpx:HTTPXClientInstrumentor"
```
Instrumentations with the "opentelemetry_instrumentor" entry-point are automatically loaded by the [auto-instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py#L69C5-L69C71):
```
for entry_point in iter_entry_points("opentelemetry_instrumentor"):
  ...
  distro.load_instrumentor(entry_point, skip_dep_check=True)
```

### 🔄️ Additional Information

_No response_

None

nirga:
Attempt to fix #591 
Didn't enable for the following instrumentations since it fails the build for some reason:
* `bedrock` (`boto3`)
* `watsonx`
* `tranformers`


jinsongo:
@nirga The build problem has been fixed for Watsonx by the PR https://github.com/traceloop/openllmetry/pull/885

PR #695, opened 2024-03-26
fix(traceloop-sdk): custom span processor's on_start is honored

### LINKED ISSUE #686: 🐛 Bug Report: Sentry.io integration breaks

Traceloop SDK

### 📜 Description

According to https://www.traceloop.com/docs/openllmetry/integrations/sentry, Traceloop SDK should work with Sentry SDK by using OpenTelemetry as its instrumenter. However the traces are never sent out.

### 👟 Reproduction steps

1. `pip install --upgrade 'sentry-sdk[opentelemetry]'`
2. Add `sentry_sdk.init(dsn=<Your DSN>, enable_tracing=True, instrumenter="hotel",debug=True)`
3. Add `Traceloop.init(processor=SentrySpanProcessor(), propagator=SentryPropagator())`
4. Run my langchain app.

### 👍 Expected behavior

As langchain invokes LLM and other chains, traces should be sent to Sentry.io, and the console should print:

```sh
 [sentry] DEBUG: Setting up integrations (with default = True)
...
Traceloop exporting traces to a custom exporter
...
 [sentry] DEBUG: [Tracing] Create new propagation context: {'trace_id': '123456789abcdef', 'span_id': '12341234', 'parent_span_id': None, 'dynamic_sampling_context': None}
 [sentry] DEBUG: Sending envelope [envelope with 1 items (transaction)] project:123412341234 host:<DSN>.ingest.us.sentry.io
 [sentry] DEBUG: 1 event(s) pending on flush
Sentry is attempting to send 1 pending events
Waiting up to 2 seconds
Press Ctrl-C to quit
 [sentry] DEBUG: background worker flushed
...
```

However, the traces are never sent out. I only see the following in console:

```sh
 [sentry] DEBUG: Setting up integrations (with default = True)
...
Traceloop exporting traces to a custom exporter
...
 [sentry] DEBUG: [Tracing] Create new propagation context: {'trace_id': '123456789abcdef', 'span_id': '12341234', 'parent_span_id': None, 'dynamic_sampling_context': None}
...
 [sentry] DEBUG: background worker got flush request
 [sentry] DEBUG: background worker flushed
...
```

### 🤖 Python Version

3.11.8

I troubleshoot this issue by replacing`Traceloop.init` with `LangchainInstrumentor().instrument()`:

```python
import sentry_sdk
from opentelemetry.propagate import set_global_textmap
from sentry_sdk.integrations.opentelemetry import SentrySpanProcessor, SentryPropagator

def setup_sentry():
    sentry_sdk.init(
        dsn=<DSN>,
        enable_tracing=True,
        instrumenter="otel",
        debug=True,
    )

    provider = TracerProvider()
    provider.add_span_processor(SentrySpanProcessor())
    set_global_textmap(SentryPropagator())

    LangchainInstrumentor().instrument()

def my_langchain_app():
    setup_sentry()
    # ...
``` 

Both `LangchainInstrumentor` and `sentry_sdk` work as expected: langchain workflow traces are sent and can be observed on Sentry.io.

After switching back to `Traceloop.init(processor=SentrySpanProcessor(), propagator=SentryPropagator())`, it doesn't work anymore.

I've looked into the source code of both Traceloop SDK and Sentry python SDK, I find Sentry's custom `SentrySpanProcessor` relies heavily on the `on_start` method:

https://github.com/getsentry/sentry-python/blob/05d1e5ca94cc4fffcd01c46ceda6713459308404/sentry_sdk/integrations/opentelemetry/span_processor.py#L117-L178

But `TracerWrapper` replaces spans_processor's `on_start` with its own implementation `_span_processor_on_start`:

https://github.com/traceloop/openllmetry/blob/630c7b3d90cd8db8610a3c58c5979a03ce54d348/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L111

This breaks Sentry's logic. 

I'm willing to provide a fix but I'm not sure how the 2 `on_start` should coexist. Could you give some suggestions? Thanks.

None

evisong:
Fixes #686.

If a custom span processor is specified, its `on_start` will be called after setting LLM-specific attributes.

E.g. (https://www.traceloop.com/docs/openllmetry/integrations/sentry):
```python
Traceloop.init(processor=SentrySpanProcessor(), propagator=SentryPropagator())
```

### DIFF
--- a/packages/traceloop-sdk/tests/conftest.py
+++ b/packages/traceloop-sdk/tests/conftest.py
@@ -3,6 +3,8 @@
 import os
 import pytest
 from traceloop.sdk import Traceloop
+from traceloop.sdk.tracing.tracing import TracerWrapper
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 
 pytest_plugins = []
@@ -33,3 +35,27 @@ def environment():
 @pytest.fixture(scope="module")
 def vcr_config():
     return {"filter_headers": ["authorization"]}
+
+
+@pytest.fixture
+def exporter_with_custom_span_processor():
+    # Clear singleton if existed
+    if hasattr(TracerWrapper, "instance"):
+        _trace_wrapper_instance = TracerWrapper.instance
+        del TracerWrapper.instance
+
+    class CustomSpanProcessor(SimpleSpanProcessor):
+        def on_start(self, span, parent_context=None):
+            span.set_attribute("custom_span", "yes")
+
+    exporter = InMemorySpanExporter()
+    Traceloop.init(
+        exporter=exporter,
+        processor=CustomSpanProcessor(exporter),
+    )
+
+    yield exporter
+
+    # Restore singleton if any
+    if _trace_wrapper_instance:
+        TracerWrapper.instance = _trace_wrapper_instance
--- a/packages/traceloop-sdk/tests/test_sdk_initialization.py
+++ b/packages/traceloop-sdk/tests/test_sdk_initialization.py
@@ -18,3 +18,15 @@ def test_resource_attributes(exporter, openai_client):
     open_ai_span = spans[0]
     assert open_ai_span.resource.attributes["something"] == "yes"
     assert open_ai_span.resource.attributes["service.name"] == "test"
+
+
+@pytest.mark.vcr("test_resource_attributes.yaml")
+def test_custom_span_processor(exporter_with_custom_span_processor, openai_client):
+    openai_client.chat.completions.create(
+        model="gpt-3.5-turbo",
+        messages=[{"role": "user", "content": "Tell me a joke about opentelemetry"}],
+    )
+
+    spans = exporter_with_custom_span_processor.get_finished_spans()
+    open_ai_span = spans[0]
+    assert open_ai_span.attributes["custom_span"] == "yes"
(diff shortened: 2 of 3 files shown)


evisong:
I've tested the changes with Sentry SDK in my local environment, it works as expected (btw it's indeed impressive!)

I haven't yet found a suitable place to add tests for this use case. Could you give some guidance? Thanks.

nirga:
Nice work @evisong, thanks! Reg. the tests - should be pretty easy. You can add a test to the SDK where you set a custom span processor with an `on_start` callback that simply adds some attribute to all spans, and then make sure that that attribute exists once you get the finished spans. 
You don't even need to add it to a separate test - you can add the span processor in the `conftest.py` file where we initialize the SDK and then just test for the attribute in one of the existing tests.

evisong:
> Nice work @evisong, thanks! Reg. the tests - should be pretty easy. You can add a test to the SDK where you set a custom span processor with an `on_start` callback that simply adds some attribute to all spans, and then make sure that that attribute exists once you get the finished spans. You don't even need to add it to a separate test - you can add the span processor in the `conftest.py` file where we initialize the SDK and then just test for the attribute in one of the existing tests.

Thanks for the guidance @nirga.

I started by adding the logic to the existing fixture `exporter`, but soon I realized the impact may be too big. So instead I created a new fixture `exporter_with_custom_span_processor`. The singleton in `TracerWrapper` adds some complexity but I guess the test is still straightforward.

Please let me know if any adjustments are needed. Thanks.

nirga (approved):
Nice :) I made a small fix to not couple your new test with the mock data of the other test but other than that LGTM

PR #710, opened 2024-03-30
Haystack V2 Version

### LINKED ISSUE #640: 🚀 Feature: Support Haystack v2

Haystack Instrumentation

### 🔖 Feature description

Upgrade our instrumentation to support the new Haystack v2 APIs

-

-

### 🔄️ Additional Information

_No response_

None

mayankagarwals:
Fixes #640

### DIFF
--- a/packages/opentelemetry-instrumentation-haystack/opentelemetry/instrumentation/haystack/__init__.py
+++ b/packages/opentelemetry-instrumentation-haystack/opentelemetry/instrumentation/haystack/__init__.py
@@ -18,7 +18,7 @@ from opentelemetry.instrumentation.haystack.version import __version__
 
 logger = logging.getLogger(__name__)
 
-_instruments = ("farm-haystack >= 1.20.1",)
+_instruments = ("haystack-ai >= 2.0.0",)
 
 WRAPPED_METHODS = [
     {
--- a/packages/opentelemetry-instrumentation-haystack/pyproject.toml
+++ b/packages/opentelemetry-instrumentation-haystack/pyproject.toml
@@ -36,7 +36,7 @@ pytest = "8.1.0"
 pytest-sugar = "1.0.0"
 
 [tool.poetry.group.test.dependencies]
-farm-haystack = "~1.20.1"
+haystack-ai = "~2.0.0"
 pytest = "8.1.0"
 pytest-sugar = "1.0.0"
 vcrpy = "^6.0.1"
@@ -48,7 +48,7 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.extras]
-instruments = ["farm-haystack"]
+instruments = ["haystack-ai"]
 
 [tool.poetry.plugins."opentelemetry_instrumentor"]
-farm_haystack = "opentelemetry.instrumentation.haystack:HaystackInstrumentor"
+haystack-ai = "opentelemetry.instrumentation.haystack:HaystackInstrumentor"


nirga:
@mayankagarwals can you update the sample app as well and share a screenshot showing this working?

nirga:
Sorry @mayankagarwals I missed that 🤦 
I don't have one, but it should look like the other samples - you can just use the test you've created and turn it into a working program. This is how the screenshot should look like - https://github.com/traceloop/openllmetry/pull/706

mayankagarwals:
Attaching Traceloop Screenshot with haystack V2

image

@nirga Can you help me with the lint command for the repo?

packages/sample-app/sample_app/haystack_app.py
nirga, line 1: I think you need to update the sample app poetry file as well, no?

packages/sample-app/sample_app/haystack_app.py
```diff
@@ -0,0 +1,28 @@
...
+Traceloop.init(app_name="haystack_example")
+
+def haystack_app():
+    
+    prompt_builder = DynamicChatPromptBuilder()
+    api_key=os.getenv("OPENAI_API_KEY")
+    llm = OpenAIChatGenerator(api_key=Secret.from_token(api_key), model="gpt-4")
+
+
+    pipe = Pipeline()
```
nirga, line 19: So this is a really simple pipeline. I wonder if we should support more complex types of workflows like the ones Haystack publishes here - https://github.com/deepset-ai/haystack-cookbook/tree/main/notebooks

nirga (commented):
Why does the OpenAI span appear twice? 🤔

packages/sample-app/sample_app/haystack_app.py
nirga: Like this one - https://haystack.deepset.ai/tutorials/27_first_rag_pipeline

packages/sample-app/sample_app/haystack_app.py
mayankagarwals: Yes! My bad. Done

mayankagarwals:
> Why does the OpenAI span appear twice? 🤔

I think that's because I see an OpenAI client specific instrumentation with same name : https://github.com/traceloop/openllmetry/blob/9f6048089c4de70cf6e43d19992fc088464ac8fe/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L34 

We can potentially change the span name to haystack.openai.chat?

packages/sample-app/sample_app/haystack_app.py
mayankagarwals: Makes sense but do you think we should consider making this a separate PR?

nirga:
> > Why does the OpenAI span appear twice? 🤔
> 
> I think that's because I see an OpenAI client specific instrumentation with same name :
> 
> https://github.com/traceloop/openllmetry/blob/9f6048089c4de70cf6e43d19992fc088464ac8fe/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L34
> 
> We can potentially change the span name to haystack.openai.chat? Let me know if you have any other naming ideas

Yep, that sounds good!

packages/sample-app/sample_app/haystack_app.py
nirga: @mayankagarwals hmm. It just that this is a better example of what people do with Haystack. Otherwise I worry that when people will use this it may not work for them.

mayankagarwals:
image

Updated screenshot after changing name

packages/sample-app/sample_app/haystack_app.py
mayankagarwals: Got it. Makes sense.

nirga (approved):

Issue #251, opened 2023-12-15
🚀 Feature: OpenTelemetry Metrics


nirga:
### Which component is this feature for?

All Packages

### 🔖 Feature description

Some values, like token usage, are better reported (also) as [otel metrics](https://opentelemetry.io/docs/specs/otel/metrics/). We should update all instrumentations to send metrics as well as traces.

### 🎤 Why is this feature needed ?

More ways to use the outputted data. Observability platforms can use metrics to set alerts on or to create dashboards.

### ✌️ How do you aim to achieve this?

This needs to be researched and defined. Please consult on [slack](traceloop.com/slack) before starting to work on this.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

nirga:
@quest-bot stash 300

Humbertzhang:
@quest-bot embark

hanchchch:
@quest-bot embark

naghost-dev:
@quest-bot embark

codefromthecrypt:
some upstream progress on metric definition https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md @naghost-dev if it helps your work in progress

nitin302:
hello, looking for metrics, is this available?

nirga:
Supported on some instrumentations, we're adding more everyday - specifically langchain isn't supported yet unfortunately.

Rutik7066:
@quest-bot embark

Mayank77maruti:
@quest-bot embark

Mayank77maruti:
Hey @nirga , It would be really great if you can share some additional context regarding the issue.

nirga:
@Mayank77maruti sorry I missed it - you should look at the existing instrumentations and see where we implemented metrics (like the OpenAI one). We're missing this in some other instrumentations (unfortunately I don't have an inventory checklist for that :/ )

ash01825:
@quest-bot embark

PR #711, opened 2024-03-30
fix(traceloop-sdk): default value for metrics endpoint

hanchchch:
Relates to #251

Current
- The metrics endpoint does not have default value, it must be set by env var for now.
- It also have to be set with otlp metrics path `/v1/metrics`, unlikely to the [traces endpoint](https://github.com/traceloop/openllmetry/tree/main/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L386).
    ```python
    def init_spans_exporter(api_endpoint: str, headers: Dict[str, str]) -> SpanExporter:
        if "http" in api_endpoint.lower() or "https" in api_endpoint.lower():
            return HTTPExporter(endpoint=f"{api_endpoint}/v1/traces", headers=headers)
        else:
            return GRPCExporter(endpoint=f"{api_endpoint}", headers=headers)
    ```

Proposed
- Use `api_endpoint` as default of the metrics endpoint.
- Append metrics path, just like the traces endpoint.

I didn't open new issue for this just because it's a very small and simple change. Please feel free to comment. 

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/__init__.py
+++ b/packages/traceloop-sdk/traceloop/sdk/__init__.py
@@ -174,7 +174,7 @@ class Traceloop:
         if metrics_exporter:
             print(Fore.GREEN + "Traceloop exporting metrics to a custom exporter")
 
-        metrics_endpoint = os.getenv("TRACELOOP_METRICS_ENDPOINT")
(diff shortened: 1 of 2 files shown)


nirga (approved):
You're right! It makes much more sense, thanks!

PR #745, opened 2024-04-03
fix(openai): switch to init flag for token usage enrichment

nirga:


packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -450,7 +450,9 @@ def init_openai_instrumentor():
         Telemetry().capture("instrumentation:openai:init")
         from opentelemetry.instrumentation.openai import OpenAIInstrumentor
 
-        instrumentor = OpenAIInstrumentor(enrich_assistant=True)
+        instrumentor = OpenAIInstrumentor(
+            enrich_assistant=True, enrich_token_usage=True
```
galkleinman, line 454: `os.getenv("OPENAI_SOMETHING")`  if exist and otherwise True? otherwise how can it be disabled when using the sdk?

galkleinman (approved):

Issue #302, opened 2024-01-04
🐛 Bug Report: Async and streaming Anthropic is not supported


nirga:
### Which component is this bug for?

Anthropic Instrumentation

### 📜 Description

Anthropic provides an async API as well as a sync API: 
https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#async-usage

Both should be supported as part of the instrumentation

### 👟 Reproduction steps

-

### 👍 Expected behavior

-

### 👎 Actual Behavior with Screenshots

-

### 📃 Provide any additional context for the Bug.

This can be supported similarily to our OpenAI instrumentation

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

AromaticHydrocarbon:
Hello. Can I work on this issue?

nirga:
Sure @AromaticHydrocarbon!

PR #750, opened 2024-04-05
fix(anthropic): Async and streaming Anthropic  (#302)

AromaticHydrocarbon:

image
image
image

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -16,6 +16,8 @@ from opentelemetry.instrumentation.utils import (
     unwrap,
 )
 
+from anthropic._streaming import Stream, AsyncStream
+
 from opentelemetry.semconv.ai import SpanAttributes, LLMRequestTypeValues
 from opentelemetry.instrumentation.anthropic.version import __version__
 
@@ -36,7 +38,37 @@ WRAPPED_METHODS = [
         "method": "create",
         "span_name": "anthropic.completion",
     },
+    {
+        "package": "anthropic.resources.messages",
+        "object": "Messages",
+        "method": "stream",
+        "span_name": "anthropic.completion.stream",
+    },
 ]
+WRAPPED_AMETHODS = [
+    {
+        "package": "anthropic.resources.completions",
+        "object": "AsyncCompletions",
+        "method": "create",
+        "span_name": "anthropic.completion",
+    },
+    {
+        "package": "anthropic.resources.messages",
+        "object": "AsyncMessages",
+        "method": "create",
+        "span_name": "anthropic.completion",
+    },
+    {
+        "package": "anthropic.resources.messages",
+        "object": "AsyncMessages",
+        "method": "stream",
+        "span_name": "anthropic.completion.stream",
+    },
+]
+
+
+def is_streaming_response(response):
+    return isinstance(response, Stream) or isinstance(response, AsyncStream)
 
 
 def should_send_prompts():
@@ -86,6 +118,7 @@ def _set_input_attributes(span, kwargs):
     _set_span_attribute(
         span, SpanAttributes.LLM_PRESENCE_PENALTY, kwargs.get("presence_penalty")
     )
+    _set_span_attribute(span, SpanAttributes.LLM_IS_STREAMING, kwargs.get("stream"))
 
     if should_send_prompts():
         if kwargs.get("prompt") is not None:
@@ -102,6 +135,22 @@ def _set_input_attributes(span, kwargs):
                 )
 
 
+def _set_completions(span, choices):
+    if not span.is_recording() or not choices:
+        return
+
+    try:
+        for choice in choices:
+            index = choice.get("index")
+            prefix = f"{SpanAttributes.LLM_COMPLETIONS}.{index}"
+            _set_span_attribute(
+                span, f"{prefix}.finish_reason", choice.get("finish_reason")
+            )
+            _set_span_attribute(span, f"{prefix}.content", choice.get("text"))
+    except Exception as e:
+        logger.warning("Failed to set completion attributes, error: %s", str(e))
+
+
 def _set_span_completions(span, response):
     index = 0
     prefix = f"{SpanAttributes.LLM_COMPLETIONS}.{index}"
@@ -115,6 +164,33 @@ def _set_span_completions(span, response):
                 f"{SpanAttributes.LLM_COMPLETIONS}.{i}.content",
                 content.text,
             )
+
+
+async def _set_token_usage_a(span, anthropic, request, response):
+    if not isinstance(response, dict):
+        response = response.__dict__
+
+    prompt_tokens = 0
+    if request.get("prompt"):
+        prompt_tokens = await anthropic.count_tokens(request.get("prompt"))
+    elif request.get("messages"):
+        prompt_tokens = sum(
+            [await anthropic.count_tokens(m.get("content")) for m in request.get("messages")]
+        )
+
+    completion_tokens = 0
+    if response.get("completion"):
+        completion_tokens = await anthropic.count_tokens(response.get("completion"))
+    elif response.get("content"):
+        completion_tokens = await anthropic.count_tokens(response.get("content")[0].text)
+
+    total_tokens = prompt_tokens + completion_tokens
+
+    _set_span_attribute(span, SpanAttributes.LLM_USAGE_PROMPT_TOKENS, prompt_tokens)
+    _set_span_attribute(
+        span, SpanAttributes.LLM_USAGE_COMPLETION_TOKENS, completion_tokens
+    )
+    _set_span_attribute(span, SpanAttributes.LLM_USAGE_TOTAL_TOKENS, total_tokens)
 
 
 def _set_token_usage(span, anthropic, request, response):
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -178,47 +254,213 @@ def wrapper(wrapped, instance, args, kwargs):
...
 
-        response = wrapped(*args, **kwargs)
 
-        if response:
-            try:
-                if span.is_recording():
-                    _set_response_attributes(span, response)
-                    _set_token_usage(span, instance._client, kwargs, response)
+@_with_tracer_wrapper
+async def _a_wrap(tracer, to_wrap, wrapped, instance, args, kwargs):
```
nirga, line 422:
```suggestion
async def _awrap(tracer, to_wrap, wrapped, instance, args, kwargs):
```

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -117,6 +166,33 @@ def _set_span_completions(span, response):
             )
 
 
+async def _set_token_usage_a(span, anthropic, request, response):
```
nirga, line 169: why do you need this function? `count_tokens` isn't an async method

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -178,47 +254,213 @@ def wrapper(wrapped, instance, args, kwargs):
...
+            complete_response.get("events")[index]["text"] += item.delta.text
+        elif item.type == 'message_delta':
+            for event in complete_response.get("events", []):
+                event["finish_reason"] = item.delta.stop_reason
+        yield item_to_yield
+
+    # prompt_usage
+    prompt_tokens = 0
+    if kwargs.get("prompt"):
+        prompt_tokens = instance.count_tokens(kwargs.get("prompt"))
```
nirga, line 284: This call can be costly and increase latency. I'd add a flag for the instrumentation controlling whether this is enabled or not - see the OpenAI instrumentation for an example

packages/opentelemetry-instrumentation-anthropic/tests/test_completion.py
```diff
@@ -107,12 +107,121 @@ def test_anthropic_multi_modal(exporter):
...
+    spans = exporter.get_finished_spans()
+    assert [span.name for span in spans] == [
+        "anthropic.completion",
+    ]
+    anthropic_span = spans[0]
+    assert (
+            anthropic_span.attributes["llm.prompts.0.user"]
+            == "Tell me a joke about OpenTelemetry"
+    )
+    assert (anthropic_span.attributes.get("llm.completions.0.content"))
```
nirga, line 221: Can we test that this attribute is actually the response we accumulated in line 209?

nirga (commented):
Thanks so much for this @AromaticHydrocarbon! I've left a couple of comments. 

Note that you should upload the recordings of your new tests - set an Anthropic Key in your environment, comment out the line with the mock anthropic key in `conftest.py` and then run them locally with `poetry run pytest -v --record-mode=once -vv`

nirga:
@AromaticHydrocarbon have you seen my previous comments?

AromaticHydrocarbon:
> @AromaticHydrocarbon have you seen my previous comments?

@nirga Thank you very much for your feedback on the code I submitted. I have made modifications and improvements based on your suggestions. 

1. Regarding uploading the recordings of new tests: I apologize for not understanding how GitHub tests are executed. I have now uoload my cassettes.

2. Regarding avoiding code duplication: I have extracted the repeated lines in the stream section and moved the common code into specific methods to reduce duplication.

3. Regarding putting the streaming handling code in a separate file: I have moved the streaming handling code to a separate file.

4.  Regarding the count_tokens function: I carefully reviewed the code and noticed that the count_tokens function in the AsyncAnthropic class is indeed already defined as an asynchronous method. 

5. Regarding adding a flag to control whether to enable costly calls: I referred to the example in the OpenAI instrumentation and used the should_record_stream_token_usage() function in the streaming handling code. 
 
6.Regarding testing whether the accumulated response on line 209 is actually set as an attribute: I have added corresponding content in the tests

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/config.py
```diff
@@ -0,0 +1,2 @@
+class Config:
+    enrich_token_usage = True
```
nirga, line 2: A few comment on this Config:
1. I think it's best if its `False` by default. 
2. You also need to have an option to set it during `__init__` of the instrumentation (see the OpenAI instrumentation for an example).
3. You can then pass `True` to it when `traceloop_sdk` initialize it (see how we did it with the OpenAI instrumentation).

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+            prefix = f"{SpanAttributes.LLM_COMPLETIONS}.{index}"
+            _set_span_attribute(
+                span, f"{prefix}.finish_reason", choice.get("finish_reason")
+            )
+            _set_span_attribute(span, f"{prefix}.content", choice.get("text"))
+    except Exception as e:
+        logger.warning("Failed to set completion attributes, error: %s", str(e))
+
+
+def should_record_stream_token_usage():
```
nirga, line 54: nit: not needed, you can just check the config directly

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _build_from_streaming_response(
+        span,
+        response,
+        instance,
+        kwargs
+):
+    complete_response = {"events": [], "model": "", "usage": {}}
+    for item in response:
+        process_response_item(item, complete_response)
```
nirga, line 66: nit: first yield, then process

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+
+async def _abuild_from_streaming_response(
+        span,
+        response,
+        instance,
+        kwargs
+):
+    complete_response = {"events": [], "model": ""}
+    async for item in response:
+        process_response_item(item, complete_response)
```
nirga, line 110: same here - first yield then process

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+        instance,
+        kwargs
+):
+    complete_response = {"events": [], "model": "", "usage": {}}
+    for item in response:
+        process_response_item(item, complete_response)
+        yield item
+
+    # calculate token usage
+    if should_record_stream_token_usage():
```
nirga, line 70: this is risky and may fail - better to wrap it with `try` - `except` like we have in `set_completions`

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -0,0 +1,15 @@
+import os
+from opentelemetry import context as context_api
+
+
+def _set_span_attribute(span, name, value):
```
nirga, line 5: nit
```suggestion
def set_span_attribute(span, name, value):
```

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+from opentelemetry.instrumentation.anthropic.utils import _set_span_attribute, should_send_prompts
+from opentelemetry.trace.status import Status, StatusCode
+from opentelemetry.semconv.ai import SpanAttributes
+
+from opentelemetry.instrumentation.anthropic.config import Config
+
+logger = logging.getLogger(__name__)
+
+
+def process_response_item(item, complete_response):
```
nirga, line 11: nit
```suggestion
def _process_response_item(item, complete_response):
```

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -0,0 +1,143 @@
...
+    _set_span_attribute(span, SpanAttributes.LLM_USAGE_PROMPT_TOKENS, prompt_tokens)
+    _set_span_attribute(
+        span, SpanAttributes.LLM_USAGE_COMPLETION_TOKENS, completion_tokens
+    )
+    _set_span_attribute(span, SpanAttributes.LLM_USAGE_TOTAL_TOKENS, total_tokens)
+
+    _set_span_attribute(span, SpanAttributes.LLM_RESPONSE_MODEL, complete_response.get("model"))
+
+
+def _set_completions(span, choices):
```
nirga, line 38: if this is a copy paste from the `__init__` - better to just re-use it from there (or use this method there). I want to avoid duplications as much as possible cause these methods change often and people tend to forget to look for all the places where a certain attribute is set.

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
nirga, line 11: this is risky and may fail - can we rap it in `try except` and log an error?

nirga (commented):
Thanks @AromaticHydrocarbon! I left a couple of more comments. Sorry for the nit-picking, these instrumentations are delicate and we need to make sure we're not breaking anything for our users.

AromaticHydrocarbon:
> Thanks @AromaticHydrocarbon! I left a couple of more comments. Sorry for the nit-picking, these instrumentations are delicate and we need to make sure we're not breaking anything for our users.

Thank you for your detailed comments and suggestions. Regarding the issues you raised, I have made the following responses and modifications:

I have completed the corresponding modifications with reference to OpenAI instrumentation, setting enrich_token_usage to False by default and adding a parameter in the __init__ method of AnthropicInstrumentor to control it. 

I have renamed the process_response_item function to _process_response_item.

I have added a try-except block in the _process_response_item function to capture possible exceptions and log errors.

Regarding _set_completions(), I would like to clarify that this part of the code was copied from the streaming handle section of OpenAI instrumentation, not from __init__. 

I have adjusted the order of the code to yield first and then process.

I have renamed the _set_span_attribute function to set_span_attribute.

Thank you again for your feedback. I will continue to work on improving code quality. If you have any other suggestions or opinions, I am very willing to listen and make corresponding adjustments.

PR #763, opened 2024-04-08
fix(langchain): BedrockChat model name should be model_id

evisong:
Fixes #762

- [x] I have added tests that cover my changes.
- [x] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change.
- [x] PR name follows conventional commits format: `feat(instrumentation): ...` or `fix(instrumentation): ...`.
- [ ] (If applicable) I have updated the documentation accordingly.

PR #770, opened 2024-04-09
fix(bedrock): support anthropic v3

nirga:


evisong:
In case you run into a UT failure caused by `http://169.254.169.254/latest/api/token` request, most likely AWS boto3 client is trying to read credentials from EC2 metadata, you may refer to this commit: https://github.com/traceloop/openllmetry/pull/763/commits/a700ced251d0ed81acdbcf8c45e331491e67c8a4 for workaround this.

nirga:
Thanks @evisong! :)

PR #804, opened 2024-04-13
fix(langchain): use chain names if exist

nirga:

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/task_wrapper.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/task_wrapper.py
@@ -16,12 +16,10 @@ def task_wrapper(tracer, to_wrap, wrapped, instance, args, kwargs):
     if instance.__class__.__name__ in ("AgentExecutor"):
         return wrapped(*args, **kwargs)
 
-    if hasattr(instance, "name") and instance.name:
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/task_wrapper.py
```diff
@@ -16,12 +16,10 @@ def task_wrapper(tracer, to_wrap, wrapped, instance, args, kwargs):
...
         return wrapped(*args, **kwargs)
 
-    if hasattr(instance, "name") and instance.name:
-        name = f"{to_wrap.get('span_name')}.{instance.name.lower()}"
-    elif to_wrap.get("span_name"):
-        name = to_wrap.get("span_name")
-    else:
-        name = f"langchain.task.{instance.__class__.__name__}"
+    config = args[1] if len(args) > 1 else {}
+    run_name = config.get("run_name") or instance.get_name()
```
galkleinman, line 20: worth failsafing `instance.get_name()` maybe? i mean, checking whether it exists before accessing it.

galkleinman (approved):

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/task_wrapper.py
nirga: Will do with #748

PR #893, opened 2024-04-25
fix(sdk): reset the color of messages in the custom metrics exporter

reiyw:
This pull request adds a reset code that was missing from the messages output when using the custom metric exporter. Prior to making this change, there was no reset output after this message, so all subsequent messages were green.


nirga (approved):
Thanks!

Issue #855, opened 2024-04-19
🚀 Feature: log content filter results in proper attributes (Azure OpenAI)


nirga:
### Which component is this feature for?

OpenAI Instrumentation

### 🔖 Feature description

Azure OpenAI has an option to [filter inappropriate content](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cpython-new), which returns a slightly different response. While #854 handles this partially, we should explicitly log all the reasons in separate attributes.

### 🎤 Why is this feature needed ?

-

### ✌️ How do you aim to achieve this?

Create one or more attributes to log the reasons content was redacted. 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

dtee1:
Interested

sukidesuka:
This change definitely has bugs. I updated to the latest version of traceloop-sdk today, 0.16.6, but even though gpt4's responses are normally output to the console without any filtering issues, it still shows FILTERED on traceloop.

![image](https://github.com/traceloop/openllmetry/assets/63297045/cf5019e5-6af0-47bd-828c-29f02585ed52)

![image](https://github.com/traceloop/openllmetry/assets/63297045/70d9b941-df34-496c-b986-d13c94602411)

sukidesuka:
According to the Azure official documentation https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cpython-new

```
data: {"id":"","object":"","created":0,"model":"","prompt_annotations":[{"prompt_index":0,"content_filter_results":{"hate":{"filtered":false,"severity":"safe"},"self_harm":{"filtered":false,"severity":"safe"},"sexual":{"filtered":false,"severity":"safe"},"violence":{"filtered":false,"severity":"safe"}}}],"choices":[],"usage":null}

data: {"id":"chatcmpl-7rCNsVeZy0PGnX3H6jK8STps5nZUY","object":"chat.completion.chunk","created":1692913344,"model":"gpt-35-turbo","choices":[{"index":0,"finish_reason":null,"delta":{"role":"assistant"}}],"usage":null}

[...]

PR #901, opened 2024-04-27
Added logging for content filter

dtee1:

### DIFF
--- a/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
+++ b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
@@ -549,7 +549,11 @@ def _accumulate_stream_items(item, complete_response):
 
         delta = choice.get("delta")
 
-        if delta and delta.get("content"):
-            complete_choice["message"]["content"] += delta.get("content")
-        if delta and delta.get("role"):
-            complete_choice["message"]["role"] = delta.get("role")
+    try:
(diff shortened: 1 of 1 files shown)


nirga:
Thanks @dtee1 but I think we already wrap everything with a try-catch and log an exception (see the `@dont_throw` annotation)

dtee1:
> Thanks @dtee1 but I think we already wrap everything with a try-catch and log an exception (see the `@dont_throw` annotation)

Ok sounds good. I can take that out. Would you rather I just log the warning in an else?

nirga:
> > Thanks @dtee1 but I think we already wrap everything with a try-catch and log an exception (see the `@dont_throw` annotation)
> 
> 
> 
> Ok sounds good. I can take that out. Would you rather I just log the warning in an else?

We also log there 😀

dtee1:
> > > Thanks @dtee1 but I think we already wrap everything with a try-catch and log an exception (see the `@dont_throw` annotation)
> > 
> > 
> > Ok sounds good. I can take that out. Would you rather I just log the warning in an else?
> 
> We also log there 😀

Ok, how is the fix #855 supposed to be implemented? Maybe I missed something

nirga:
@dtee1 so when calling azure it may filter the response returned from the model if it contains harmful text, see https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cpython-new

We should add attributes to the span according to those returned from azure in this case.

nirga:
Superseded by #1539

PR #1068, opened 2024-05-19
feat(milvus): add Milvus instrumentation

### LINKED ISSUE #244: 🚀 Feature: Milvus Instrumentation

All Packages

### 🔖 Feature description

Instrument calls to [Milvus](https://milvus.io/docs/search.md), including adding attributes, similarly to our [Chroma](https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-chroma) instrumentation. The instrumentation should support all types of calls - streaming, non streaming, async, etc.

Completness of OpenLLMetry

Similarily to other instrumentations we have in this repo.

### 🔄️ Additional Information

* Make sure to initialize this in [Traceloop SDK](https://github.com/traceloop/openllmetry/blob/main/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L241).
* Add proper [tests](https://github.com/traceloop/openllmetry/tree/main/packages/traceloop-sdk/tests).

None

Josecespedesant:
Fixes #244

Testing instrumentation in Traceloop:

image

- [x] I have added tests that cover my changes. (Same as chromadb package, some test cases can still be added)


Josecespedesant:
Also mentioning #244 the issue here as well

nirga (commented):
Thanks so much @Josecespedesant! Looks like there's an error when trying to install the package, can you take a look?

nirga:
@Josecespedesant looks like there's a conflict during installation:

https://github.com/traceloop/openllmetry/actions/runs/9147471985/job/25149747974

Josecespedesant:
Will take a look at it!

packages/opentelemetry-instrumentation-milvus/opentelemetry/instrumentation/milvus/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+WRAPPED_METHODS = [
+    {
+        "package": pymilvus,
+        "object": "MilvusClient",
+        "method": "insert",
+        "span_name": "milvus.insert"
+    },
+    {
+        "package": pymilvus,
```
nirga, line 30: I think this needs to be a string otherwise this may fail

nirga (commented):
Overall looks good. Left a small comment, and we need to make the test pass

packages/opentelemetry-instrumentation-milvus/opentelemetry/instrumentation/milvus/__init__.py
Josecespedesant: Hi! From what I saw in the chromadb instrumentation the package is added directly as this, I don't know if for this case it's different but it seems to work as this?

nirga: You're right! I was only looking at Pinecone

nirga (approved):
Thanks so much @Josecespedesant, great work!

Josecespedesant:
Thank you very much! Looking forward to work on any other issue that arises 🙌

PR #1134, opened 2024-05-24
fix(weaviate): support v4.6.3

tibor-reiss:
Related to #1123

- Bump weaviate from 3.26.2 to 4.6.3
- Update instrumentation to be compatible with v4
- Update the examples
- Keep v3-specific code so that 3.26.2 still works


nirga:
Thanks @tibor-reiss, did you try it to make sure it works with the new version?

tibor-reiss:
Hi @nirga, yes, 9 passed, 1 skipped.
Took me a while to figure out that it was just a bump in the version in `__init__` - but this was a good learning curve :)

nirga:
@tibor-reiss can you run one of the sample apps that use weaviate and make sure that you see traces, and add a screenshot here? I vaguely remember that there was a reason why @paolorechia who wrote this instrumentation didn't enable it for v3

tibor-reiss:
@nirga Took me some time to set up a fully functioning environment locally - consequently, I have updated the sample apps to accommodate for local work. I adjusted v4 so it resembles v3.
It seems v3 still runs with weaviate v4, but it will be deprecated at some point.
Screenshot attached from traceloop.

nirga:
@tibor-reiss actually no :/ the GET / DELETE spans you're seeing are instrumented by otel's URLLIB3 that we're installing :/ Looks like the weaviate spans aren't there. You can easily see it by disabling all instrumentations except for weaviate, like this:

```
from traceloop.sdk.instruments import Instruments

Traceloop.init(instruments={Instruments.WEAVIATE})
```

From that, I'm guessing that if we update the test files to use the v4 syntax they will also start failing

tibor-reiss:
Thanks for your patience @nirga! Here the updated screenshot.
image

nirga:
Thanks @tibor-reiss! Don't update version numbers, they get bumped automatically when we release 😃

So now it works both with old and new versions of weaviate? Or did we drop support for old versions?

tibor-reiss:
@nirga Oh, sorry, I did not think that you would like to keep support for v3 - installing 0.19.0 with weaviate==3.26.2 would have still worked.

Anyway, I have put it back - with minor changes:
- _GraphQLInstrumentor is the only overlap
- marked the v3 so hopefully it can be easily deleted in the future
- moved the previous tests to *_v3.py

So now both versions work - tested both the weaviate_v3.py and weaviate_v4.py.

nirga:
@tibor-reiss the recording of the cassettes don't work for you?

nirga:
Hey @tibor-reis! What VCR does is just recording and replaying of HTTP requests, so as long as the database is sending back responses this should work. I can try and look into this as well today. 

During CI, I think spinning off a local weaviate might be more cumbersome than recording and replaying HTTP responses but if you think it's easier and stable enough - let's do it. Can you assist in fixing the CI yaml though? (it's under the .github folder)

tibor-reiss:
The reason for some test failures is that grpc is built into weaviate-v4. grpc, but this is not supported in vcrpy. As a solution for these tests, I have removed the vcr markers, and added a command line flag (`with_grpc`), i.e. the tests which use grpc (e.g. fetching) need a running weaviate instance.

packages/sample-app/sample_app/weaviate_v3.py
```diff
@@ -1,40 +1,37 @@
 # Tested with weaviate-client==3.26.0
+# Weaviate instrumentation with opentelemetry-instrumentation-weaviate==0.19.0
```
nirga, line 2: this can be removed, right? cause we support both

nirga (commented):
Thanks @tibor-reiss! This looks overall good, 2 small comments:
1. Can you fix the lint issues?
2. Can you remove the cassettes that are no longer used?

packages/sample-app/sample_app/weaviate_v3.py
tibor-reiss: Are there any cassettes left? I removed both "fetch" cassettes, the rest is needed afais.

nirga (approved):

PR #1140, opened 2024-05-25
fix(langchain): stringification fails for lists of LangChain `Documents`

rbellamy:
Code fails when trying to stringify lists of LangChain `Documents` e.g., when processing response from a retriever.

This should fix that problem.

Basically following the guidance found here (for the same reason): https://github.com/langchain-ai/langchain/issues/2222#issuecomment-1911839856


nirga:
@rbellamy thanks! I think this will require updating the tests as well (approved them for running now and I see they're failing)

rbellamy:
Removed stringification and leveraged a custom `JSONEncoder` for
serializing function input/output to traces.

Super sorry for not running the tests before submitting the PR.

I updated the tests I could run - neither the Bedrock nor Anthropic
worked for me - failing with vcr cassette failures. All tests I could
run now pass.

These changes radically alters how messages are represented - note
the changes made to the `test_simple_lcel` test. Specifically, where
before the messages were output in their stringified version, now
they are the json serialized form.

While this does produce more verbose output I think that's outweighed by
the a more robust approach to serializing the output.

nirga (approved):
Thanks so much for this @rbellamy I wanted to fix this for so long

PR #1141, opened 2024-05-25
feat(vertexai): `vertexai.generative_models` / `llm_model` detection

rbellamy:
- `GenerativeModel` promoted out of preview.
- `llm_model` detection for PaLM and Gemini via protected `instance` members
  - PaLM: `_model_id`
  - Gemini: `_model_name`

Prior `llm_model` detection used `global llm_model` which caused a race when tracing calls to multiple models in async code paths.

As well, given that detecting the `model_id`/`model_name` was the sole reason for wrapping `from_pretrained`/`__init__` this seems like it has a lighter footprint.


nirga:
Thanks @rbellamy! we don't have tests for vertex AI unfortunately since it doesn't play well with vcr.py
Have you tested this locally with a sample app? can you link a screenshot?

rbellamy:
Attached is a screenshot of a trace with spans from two models in the same LCEL. Hope this helps!

Screenshot 2024-05-26 at 20 44 32

nirga (approved):
Nice work, thank you @rbellamy!

Issue #541, opened 2024-02-27
🚀 Feature: re-write Langchain instrumentation to use Langchain Callbacks


nirga:
### Which component is this feature for?

Langchain Instrumentation

### 🔖 Feature description

Right now, we monkey-patch classes and methods in LlamaIndex which requires endless work and constant maintenance. Langchain has a system for [callbacks](https://python.langchain.com/docs/modules/callbacks/) that can potentially be used to create/end spans without being too coupled with with the framework's inner structure.

### 🎤 Why is this feature needed ?

Support Langchain entirely and be future-proof to internal API changes

### ✌️ How do you aim to achieve this?

Look into Langchain callbacks and how other frameworks are using it.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

maciejwie:
Chainlit uses langchain's [callbacks](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/langchain/callbacks.py) for their instrumentation, and it seems to work well enough. They inherit from langchain's `BaseTracer` class as well and do their observability through callbacks, as well as some front-end functions (ex: updating the user-facing message on each token). Langchain does allow for multiple independent callbacks to be specified, so doing it this way doesn't exclude any user-created callbacks.

midhun1998:
Hi @nirga ,
I have a rough idea on the implementation now and I am willing to pick up this issue.
After reading your comments in https://github.com/traceloop/openllmetry-js/issues/133#issuecomment-1999619802 here is my understanding:

1. Create a call handler similar to `StdOutCallbackHandler` in the same directory as `opentelemetry-instrumentation-langchain` 
2. Modify the `task_wrapper`, `atask_wrapper`, `workflow_wrapper`, and `aworkflow_wrapper` to inject the callback handler to the instance and use the callback to set/unset span. A small doubt here is that do we still need a `task_wrapper.py` and `workflow_wrapper.py`? Can I add the injecting logic to `__init__.py` itself. Which one's the right approach?
3. The callback handler should be injected for Langchain classes instances which support callbacks.

Please correct me if my understanding is incorrect. 
Thanks,
Midhun

nirga:
Right @midhun1998! And indeed we probably don't need them all

midhun1998:
Thanks for the confirmation, Nir! I will keep the issue updated with the progress. 🙂

midhun1998:
[...]

PR #902, opened 2024-04-28
feat(instrumentation): Updated Langchain instrumentation to use Langchain Callbacks

midhun1998:
This is a draft PR to record my progress and collaborate with the team to get suggestions. 

Supposed to fix #541 

I will check the below checkbox when they are ready:

- [ ] I have added tests that cover my changes.
- [ ] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change.
- [x] PR name follows conventional commits format: `feat(instrumentation): ...` or `fix(instrumentation): ...`.
- [ ] (If applicable) I have updated the documentation accordingly.

PR #1170, opened 2024-05-28
feat(langchain): use callbacks

### LINKED ISSUE #1101: 🚀 Feature: Add support for streaming in Langchain instrumentation

Langchain Instrumentation

### 🔖 Feature description

Today, for Langchain, the solution will automatically instrument invoke and subsequently generate methods.  It would be great if this functionality could be extended to include the stream / _stream methods of Runnable as well, see https://python.langchain.com/v0.2/docs/concepts/#runnable-interface 

With the move to more complex language models and subsequent increases in response times, a significant number of LLM providers are aggressively pushing streaming, both sync and async as the primary interface to their solutions.

I'm sorry, but I don't understand the internals of traceloop SDK. From an external perspective, I'd like to see tasks extended so that calls to the stream endpoint would be visible in a workflow.

### 🔄️ Additional Information

_No response_

None

tibor-reiss:
Some work on #541 (which will then allow to fix #1101 in a more maintainable way)

ToDo:
- [ ] Add async callback -> according to the docs, sync callbacks work


tibor-reiss:
There is quite some duplication, maybe there is a better way to do this?

nirga:
@tibor-reiss I think a better way will actually be something like #902. I think the contributor stopped working on it there so potentially we can pick this up (I was planning to do so sometime next week)

packages/opentelemetry-instrumentation-langchain/tests/test_chains.py
```diff
@@ -39,12 +46,8 @@ def test_sequential_chain(exporter):
     )
 
     spans = exporter.get_finished_spans()
-
```
nirga, line 42: I'm guessing you want to keep the assertions here?

nirga (commented):
Thanks @tibor-reiss this looks like a great step in the right direction! We should do something similar to our llamaindex instrumentation as well (we have an issue for that). I think this is still more robust than the way we do it today. I've been in touch with the folks from Langchain and hopefully we'll found a way for the new Runnable syntax.
I'll dig into their codebase as well to try to come up with a solution but I think your proposal here works well! (at least for now).

Mind fixing the lint / tests?

nirga:
Potentially what we can do (maybe outside the scope for this PR) - wrap a call to `Runnable.invoke()` and then create a `RunnableBinding` that wraps the original `Runnable` and invoke it

packages/opentelemetry-instrumentation-langchain/tests/test_chains.py
tibor-reiss: Yes, ofc, cleaned up the PR now.

tibor-reiss:
> Potentially what we can do (maybe outside the scope for this PR) - wrap a call to `Runnable.invoke()` and then create a `RunnableBinding` that wraps the original `Runnable` and invoke it

Sounds like a good idea.

I have now edited the scope/description of the PR. As you said, it's probably better to first implement callbacks as far as possible, and do later RunnableSequence.

For this PR, what might still make sense is to add BaseChatModel. It's not trivial due to 2 reasons:
1. custom_chat_wrapper.py has quite some additional span attributes (shouldn't be a big issue)
2. There is no `on_chat_model_end` (yet). So one might have to call `self.span_end()` already in `on_chat_model_start`.

nirga:
@tibor-reiss so let's do it gradually. Begin with this PR, and think about `BaseChatModel` later. The issue with no `on_chat_model_end` is that we won't have the right timing for the span, so might be better maybe to keep it as it is for now

I see you marked this as WIP, let me know when this is ready and I'll do a full review :)

tibor-reiss:
@nirga please check
Note: according to the docs, sync callbacks work also in the async case.

nirga (commented):
Looks good overall @tibor-reiss! Great work :) 
(I'm guessing we'll need to do it for other types of chains as well, right?)

midhun1998:
> @tibor-reiss I think a better way will actually be something like #902. I think the contributor stopped working on it there so potentially we can pick this up (I was planning to do so sometime next week)

Hi @nirga ,

I was away due to some personal reasons. I have some changes parked in my local and wanted to review them, but it looks like there is some progress in this PR and I'm not sure if I should continue with my PR 😞 .  Let me know if you want me to close that one.

nirga:
Maybe we can combine your work here as well @midhun1998 ?

midhun1998:
> Maybe we can combine your work here as well @midhun1998 ?

Works for me. I can add in my changes here or get my PR finished up quickly.

packages/opentelemetry-instrumentation-langchain/tests/test_chains.py
```diff
@@ -45,6 +45,51 @@ def test_sequential_chain(exporter):
...
+        output_variables=["synopsis", "review"],
+        verbose=True,
+    )
+    await overall_chain.ainvoke(
+        {"title": "Tragedy at sunset on the beach", "era": "Victorian England"}
+    )
+
+    spans = exporter.get_finished_spans()
+
+    assert [
```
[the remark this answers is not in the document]
tibor-reiss: Added more checks.
Re name: the span name is set in the callback handler, so it can't be changed (there might be occasions where it can be, e.g. with_config)

tibor-reiss:
> Looks good overall @tibor-reiss! Great work :) (I'm guessing we'll need to do it for other types of chains as well, right?)

Yes, chains can be added. I added now StuffDocumentsChain (with a test). Maybe there is a better way to add all children of the base classes since Chain is already added and here is the callbacks attribute defined...

@midhun1998 feel free to edit

nirga (approved):
@tibor-reiss @midhun1998 I'm going to merge this so we can continue this work against `main` branch - I think it will be easier. Great work @tibor-reiss :) Thanks so much!

PR #1234, opened 2024-06-04
fix(vertexai): change the span names to match method calls

rbellamy:
Without this change it's impossible to tell from the traces which execution path was used (sync, async, streamed).

Screenshot 2024-06-04 at 06 08 39


nirga:
Nvm just saw the PR comment 🤦 
I wonder - should we do this for all instrumentations?

nirga (approved):

PR #1256, opened 2024-06-05
fix(instrumentation): Handle OpenAI run polling

michael-hhai:

# Context
When you run `openai_client.beta.threads.runs.create_and_poll`, this calls [the poll method](https://github.com/openai/openai-python/blob/main/src/openai/resources/beta/threads/runs/runs.py#L1083) which ultimately calls `self.with_raw_response.retrieve`. This returns a `LegacyAPIReponse` object instead of a `Run` object like the `runs_retrieve_wrapper` expects. However, if you have a `LegacyAPIResponse` object, then you can convert it to a `Run` object by just calling `.parse()`. This PR simply checks if the returned response object if a `LegacyAPIResponse` and calls `.parse()` if so. It also asserts that the parsed response is a `Run` object. Hopefully, this assert will make things clearer for anyone who might be working with this code somewhere down the line.


nirga:
Thanks @michael-hhai! Mind adding a test for this?

nirga:
@michael-hhai go inside the openai package and run:
`poetry run pytest --record-mode=once -vv`

nirga (approved):

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
```diff
@@ -62,7 +65,14 @@ def runs_retrieve_wrapper(tracer, wrapped, instance, args, kwargs):
 
     response = wrapped(*args, **kwargs)
 
-    if response.id in runs:
+    if type(response) is LegacyAPIResponse:
+        parsed_response = response.parse()
+    else:
+        parsed_response = response
+
+    assert type(parsed_response) is Run
```
nirga, line 73: @michael-hhai why do you need this?

nirga: @michael-hhai the problem is that we don't want to throw exceptions in the instrumentations

nirga: Good to know! We should add `@dont_throw` annotation anyway then

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/utils.py
```diff
@@ -124,7 +125,7 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
         except Exception as e:
             logger.debug(
-                "OpenLLMetry failed to trace in %s, error: %s", func.__name__, str(e)
+                "OpenLLMetry failed to trace in %s, error: %s", func.__name__, traceback.format_exc()
```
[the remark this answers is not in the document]
nirga: Yep that's indeed better! We should do it everywhere

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
```diff
@@ -55,15 +58,23 @@ def runs_create_wrapper(tracer, wrapped, instance, args, kwargs):
 
 @_with_tracer_wrapper
 def runs_retrieve_wrapper(tracer, wrapped, instance, args, kwargs):
+    @dont_throw
```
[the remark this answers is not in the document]
nirga: That's ok. You can't put it in the wrapper anyway since you'll be catching OpenAI's exceptions too which is something we don't want to do.

nirga (approved):

PR #1293, opened 2024-06-10
feat(gemini): basic support in generate_content API

nirga:


galkleinman (approved):
Consider changing the naming to `opentelemetry-instrumentation-google-generativeai` IMO.

I know it's a long and disgusting name and it's a lot of work changing it, but that's the package name of the package being wrapped which may include other models than Gemini in the future.

PR #1303, opened 2024-06-12
Added cost metrics for openai in init.py and chat_wrappers.py

apuwk:
Added cost metrics for OpenAI. Confused because the tests aren't running properly and I'm not sure where I am going wrong.


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/__init__.py
```diff
@@ -66,7 +67,14 @@ def _instrument(self, **kwargs):
                 unit="s",
                 description="Duration of chat completion operation",
             )
-
+            
+            # Adding Price Metric
+            chat_price_counter = meter.create_counter(
+                name="llm.open.ai.chat_completions.price",
+                unit="cost", # dollar?
```
nirga, line 74: yep, `usd`

PR #1341, opened 2024-06-17
fix: input/output serialization issue for langchain

galkleinman:


packages/sample-app/sample_app/methods_decorated_app.py
```diff
@@ -67,7 +67,5 @@ def joke_workflow():
     signature = generate_signature(pirate_joke)
     print(pirate_joke + "\n\n" + signature)
 
-    Traceloop.report_score("chat_id", "chat_1234", 1)
```
nirga, line 70: :(

nirga (approved):

PR #1372, opened 2024-06-21
feat(anthropic): add instrumentation for Anthropic tool calling

### LINKED ISSUE #879: 🚀 Feature: Support Anthropic's Tools

Anthropic Instrumentation

### 🔖 Feature description

Anthropic released support for calling tools with Claude: https://docs.anthropic.com/claude/docs/tool-use
We should support adding this to spans, similar to OpenAI

Completeness of Anthropic instrumentation

Similar to this:
https://github.com/traceloop/openllmetry/blob/9c32de87ddde8c16d653efcf0f23308acd661fb7/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/__init__.py#L88

### 🔄️ Additional Information

_No response_

None

peachypeachyy:
Fixes #879 


nirga (commented):
Thanks for this @peachypeachyy! Can you also add instrumentation for the [tool use](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#chain-of-thought) that anthropic returns?

peachypeachyy:
@nirga The response of tool_use as the finish reason is already captured [here](https://github.com/traceloop/openllmetry/pull/1372/commits/79ee02ada803da986c31dc51cd79ca57da7cf8b5#diff-f47bd14f210d9764e30d470ea4c66c8c406df4888014309bdb152b1efb975bc3R664):

When you say instrumentation of tool_use, do you mean setting the response attributes of the [Chain of Thought](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#chain-of-thought)? i.e setting response attributes to something like this:

```
{
      "type": "tool_use",
      "id": "toolu_01A09q90qw90lq917835lq9",
      "name": "get_weather",
      "input": {"location": "San Francisco, CA", "unit": "celsius"}
}
```

nirga:
@peachypeachyy this should be [similar to OpenAI](https://github.com/traceloop/openllmetry/blob/747cf1887831591892a83fdbd4cbce681234f697/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L353). You're right btw, we're missing these from the semantic conventions.

peachypeachyy:
@nirga done, I have captured spans for tool_use in response as well, I have used SpanAttributes.LLM_COMPLETIONS as that is what was used in the OpenAI code.

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -320,6 +337,39 @@ def _set_response_attributes(span, response):
             prompt_tokens + completion_tokens,
         )
 
+    if response.get("role"):
+        set_span_attribute(span, f"{SpanAttributes.LLM_COMPLETIONS}.role", response.get("role"))
+
+    if response.get("stop_reason"):
+        set_span_attribute(span, f"{SpanAttributes.LLM_COMPLETIONS}.stop_reason", response.get("stop_reason"))
+
+    if response.get("content"):
```
nirga, line 346: Hey @peachypeachyy this is already happening in `_set_span_completions`. What I meant is editing the code there to specify the tool selection (if a tool was selected) - and following the similar conventions [we have in the OpenAI instrumentation](https://github.com/traceloop/openllmetry/blob/bf0007e2b199c0a67fd99ea6d24a7e83a4ebabdc/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/__init__.py#L73)

peachypeachyy: Hey @nirga let me step you through my thought process.

Following is the request being sent:
```
response = client.messages.create(
    model="claude-3-opus-20240229",
    max_tokens=1024,
    tools=[
        {
            "name": "get_weather",
            "description": "Get the current weather in a given location",
            "input_schema": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g. San Francisco, CA"
                    },
                    "unit": {
                        "type": "string",
                        "enum": ["celsius", "fahrenheit"],
                        "description": "The unit of temperature, either 'celsius' or 'fahrenheit'"
                    }
                },
                "required": ["location"]
            }
        },
        {
            "name": "get_time",
            "description": "Get the current time in a given time zone",
            "input_schema": {
                "type": "object",
                "properties": {
                    "timezone": {
                        "type": "string",
                        "description": "The IANA time zone name, e.g. America/Los_Angeles"
                    }
                },
                "required": ["timezone"]
            }
        }
    ],
    messages=[
        {
            "role": "user",
            "content": "What is the weather like right now in New York City, NY?"
        }
    ]
)
```

The corresponding formatted response is:
```
{'content': [TextBlock(text='<thinking>\nThe get_weather function is relevant to answer this question, as it provides the current weather for a given location.\n\nIt requires the "location" parameter, which in this case is provided by the user as "New York City, NY". \n\nThe "unit" parameter is optional. The user did not specify a preference, so we can use the default.\n\nSince all required parameters are available, we can proceed with calling the get_weather function to get the current weather in New York City.\n</thinking>', type='text'),
             ToolUseBlock(id='toolu_013gp9gPBaYhPVLMeq4N34yU', input={'location': 'New York City, NY'}, name='get_weather', type='tool_use')],
 'id': 'msg_01H62BKGLJBD8zstmvW21Cp4',
 'model': 'claude-3-opus-20240229',
 'role': 'assistant',
 'stop_reason': 'tool_use',
 'stop_sequence': None,
 'type': 'message',
 'usage': Usage(input_tokens=748, output_tokens=165)}
```

In this case, even though we have 2 functions passed in tools, `get_weather` and `get_time`, claude's opus determined to use get_weather as seen in the `ToolUseBlock`.

I have captured this tool `get_weather` which is being used within `ToolUseBlock` using `f"{SpanAttributes.LLM_COMPLETIONS}.{i}.name"`.

Now, which tool has to be used is governed by the parameter `[tool_choice](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#forcing-tool-use)` which needs to be set in the request. I have not set it in this example, by default the value is `auto`. Do you need me to set spans for this within the request? something like `f"{SpanAttributes.LLM_REQUEST_FUNCTIONS}.{i}.tool_choice == auto`

nirga: @peachypeachyy can you look at `set_span_completions`? It's also setting the same attributes you're setting here. You should only add the tool choice logging, similar to [OpenAI](https://github.com/traceloop/openllmetry/blob/a3a7d4e4e54f51e6d8e814f36853adb9a0bfea23/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/__init__.py#L86)

nirga:
@dinmukhamedm happy for a contribution! It's a small one and I think @peachypeachyy is no longer working on it

dinmukhamedm:
Awesome, opened https://github.com/traceloop/openllmetry/pull/2150.

PR #1373, opened 2024-06-21
feat(marqo): Add marqo instrumentation

Benedikt-Wolf:

Supports the basic operations of adding/deleting documents and querying an index.


nirga (commented):
Thanks so much for this @Benedikt-Wolf! Left a small comment

Benedikt-Wolf:
Hey @nirga, thanks for reviewing my PR! I can't seem to find the comment you mentioned - I'm starting to doubt myself here 😅. Could you confirm if the comment was actually added successfully, or am I just missing something obvious? Thanks!

packages/opentelemetry-instrumentation-marqo/opentelemetry/instrumentation/marqo/wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+
+@dont_throw
+def _set_search_attributes(span, kwargs):
+    _set_span_attribute(span, "db.marqo.search.query", kwargs.get("q"))
+
+
+@dont_throw
+def _set_delete_documents_attributes(span, kwargs):
+    _set_span_attribute(
+        span, "db.marqo.delete_documents.ids_count", count_or_none(kwargs.get("ids"))
```
nirga, line 85: Can you use the semantic conventions for these? If there are common ones you can use - it'd be best. Otherwise, just extract them to the semantic conventions package.

nirga (commented):
Oops! Added the comment now

packages/opentelemetry-instrumentation-marqo/opentelemetry/instrumentation/marqo/wrapper.py
Benedikt-Wolf: With common ones, do you mean using f.e. `CHROMADB_DELETE_IDS_COUNT` for this example?

Benedikt-Wolf: I added the new semantic conventions I need to the corresponding file, but I could not use them in the wrapper since it does not import the files from my local changes. I'd do another PR after this changing `'db.marqo.search.query'` etc. to `AISpanAttributes.MARQO_SEARCH_QUERY` to fix that. Or is there another way to resolve this?

nirga: Yes, that's a limitation of Poetry with nested local deps. I've released a version for the semconv so it should work for you now @Benedikt-Wolf

nirga (approved):

Issue #1428, opened 2024-07-01
🐛 Bug Report: OpenAI message doesn't account for `tool_calls`


achang97:
### Which component is this bug for?

OpenAI Instrumentation

### 📜 Description

The current implementation assumes that every message has the `content` field. However, this is an unsafe assumption, as there can also be messages with just a `tool_calls` array.

### 👟 Reproduction steps

1. Add an input message with  `tool_calls` and no `content`

### 👍 Expected behavior

It should include `tool_calls` as a separate JSONified field.

### 👎 Actual Behavior with Screenshots

```
        "gen_ai.prompt.1.role": "user",
        "gen_ai.prompt.1.content": "Can you tell me the weather in San Francisco, CA?",

        # This should be a tool call. Instead, it just references the previous message.
        "gen_ai.prompt.2.role": "assistant",
        "gen_ai.prompt.2.content": "Can you tell me the weather in San Francisco, CA?",
```

### 🤖 Python Version

3.11.6

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
Fixed with #1429

PR #1429, opened 2024-07-01
fix(openai): Handle `tool_calls` assistant messages

achang97:
https://github.com/traceloop/openllmetry/issues/1428


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -318,13 +318,16 @@ def _set_prompts(span, messages):
 
     for i, msg in enumerate(messages):
         prefix = f"{SpanAttributes.LLM_PROMPTS}.{i}"
-        if isinstance(msg.get("content"), str):
-            content = msg.get("content")
-        elif isinstance(msg.get("content"), list):
```
achang97, line 323: I removed this check, since according to OpenAI docs, `content` should always be a `string` or `null`

nirga (commented):
Thanks @achang97! Can you rebase this?

nirga (approved):

Issue #1297, opened 2024-06-11
🐛 Bug Report: Not getting token counts for Databricks (both foundation & external models)


tkanhe:
### Which component is this bug for?

Langchain Instrumentation

### 📜 Description

Databricks supports the OpenAI Client for querying LLM models (foundation and external models). I am using it with Langchain. I am getting the traces but not the token count.

Ref. https://docs.databricks.com/en/machine-learning/model-serving/score-foundation-models.html

### 👟 Reproduction steps

Code:

```
import asyncio

from fastapi import FastAPI
from fastapi.responses import StreamingResponse
from langchain.callbacks import AsyncIteratorCallbackHandler
from langchain.chains.question_answering import load_qa_chain
from langchain.docstore.document import Document
from langchain_openai import ChatOpenAI
from pydantic import BaseModel
from traceloop.sdk import Traceloop
from traceloop.sdk.decorators import workflow

from prompt import prompt_template, variables

Traceloop.init(app_name="tk test", disable_batch=True, api_endpoint="http://localhost:4318")


app = FastAPI()


class Message(BaseModel):
    content: str


@workflow(name="send_message")
async def send_message(question: str):
    callback = AsyncIteratorCallbackHandler()

    model = ChatOpenAI(
        model_name="databricks-dbrx-instruct",
        api_key="dapi51b9336b04**********************e",
        base_url="https://dbc-***********.cloud.databricks.com/serving-endpoints",
        streaming=True,
        callbacks=[callback],
    )

    chain = load_qa_chain(model, chain_type="stuff", prompt=prompt_template, verbose=False)

    task = asyncio.create_task(chain.ainvoke({"input_documents": [Document(page_content=variables["context"])], "question": question}))

    try:
        async for token in callback.aiter():
            yield token
    except Exception as e:
        print(f"Caught exception: {e}")
    finally:
        callback.done.set()
    await task


@app.post("/stream_chat")
def stream_chat(message: Message):
    generator = send_message(message.content)
    return StreamingResponse(generator, media_type="text/event-stream")
```

### 👍 Expected behavior

It should give the token count along with the traces...

### 👎 Actual Behavior with Screenshots

Traces I'm getting:
```
{
        "Timestamp": datetime.datetime(2024, 6, 11, 11, 41, 7, 762861),
        "TraceId": "81d12193976454e2d9bdbdd0944c3c06",
        "SpanId": "aba567ee86a445e8",
        "ParentSpanId": "55c0ac37410e2496",
[...]

PR #1452, opened 2024-07-04
fix(langchain): report llm spans (instead of normal instrumentations)

### LINKED ISSUE #1395: 🐛Bug: Gemini streaming with Langchain spans are not logged

Langchain Instrumentation

### 🔖 Feature description

Consider intercepting existing callback handlers since they might have metrics to be published as well

In my current scenario, I have a RAG chain that already utilizes a callback handler while sending the request to the Vertex AI. The custom callback internally is called more than once and the callback calculates latencies of each internal requests to the LLM. Once these latencies are calculated, I am sending them using `Traceloop.set_association_properties(info.model_dump())` in the handler but it is never part of the span. I think it would be really great if traceloop can also intercept existing callback handlers to ensure that any Traceloop related metrics that are expected to be published from these customer handlers, are also considered.

Aspect oriented programming should be leveraged to introduce cross cutting concerns. Or, we may think of introducing custom decorators for these callbacks where such metrics are added.

### 🔄️ Additional Information

_No response_

None

nirga:
Fixes #1395, #1297

Instead of relying on the SDKs instrumentations when using LangChain, we should use the fact that frameworks like LangChain and LlamaIndex know about the LLM calls and all the details and already provide them in a standard format. This PR starts reporting LLM calls here and suppresses those reports in the LLM-specific instrumentations.

cc: @tibor-reiss

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -512,7 +512,7 @@ async def _awrap(
     if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY) or context_api.get_value(
         SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY
     ):
-        return wrapped(*args, **kwargs)
+        return await wrapped(*args, **kwargs)
 
     name = to_wrap.get("span_name")
     span = tracer.start_span(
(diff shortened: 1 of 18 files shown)


tibor-reiss:
100%

Something to consider regarding the tests: remove the llm-specific instrumentors when langchain (probably same for llama-index) is used, e.g. https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-langchain/tests/conftest.py#L25

nirga:
The problem is that when folks use the SDK, it installs these instrumentations anyway, so what you're suggesting will just hide the issue. Parallel to this, I'm working on #1453 which will turn them off in langchain traces. But, this still relays on the context being propagated correctly which doesn't seem to always be the case from your other PR.

PR #1494, opened 2024-07-09
fix: association properties and workflow / task on metrics

nirga:


galkleinman (approved):
Much cleaner than I thought. 

I would omit the "common" from "common_attributes" everywhere, as it doesn't provide any additional info imo.

PR #1504, opened 2024-07-09
refactor(openai): rename functions to tools

achang97:


packages/opentelemetry-instrumentation-openai/tests/traces/test_functions.py
```diff
@@ -138,10 +183,6 @@ async def test_open_ai_function_calls_tools_streaming(
         open_ai_span.attributes[f"{SpanAttributes.LLM_COMPLETIONS}.0.tool_calls.0.id"],
         str,
     )
-    assert (
```
achang97, line 141: I dropped these checks because these tests are checking the LLM_COMPLETIONS output and the functionality for checking LLM_TOOLS shouldn't differ

nirga:
This introduces a breaking change, and given that we have [another breaking change](https://github.com/open-telemetry/semantic-conventions/pull/980) coming soon which is the approved semantic convention I don't think we should merge this as of now.

nirga:
This introduces a breaking change, and given that we have [another breaking change](https://github.com/open-telemetry/semantic-conventions/pull/980) coming soon which is the approved semantic convention I don't think we should merge this as of now.

PR #1520, opened 2024-07-10
refactor(openai): add nested `function` key to `tool_calls`

achang97:


nirga:
This introduces a breaking change, and given that we have [another breaking change](https://github.com/open-telemetry/semantic-conventions/pull/980) coming soon which is the approved semantic convention I don't think we should merge this as of now.

PR #1521, opened 2024-07-10
feat(openai): add embeddings data

achang97:


packages/opentelemetry-instrumentation-openai/tests/traces/test_embeddings.py
```diff
@@ -51,6 +54,9 @@ def test_embeddings_with_raw_response(exporter, openai_client):
         == "text-embedding-ada-002"
     )
     assert open_ai_span.attributes[SpanAttributes.LLM_USAGE_PROMPT_TOKENS] == 8
+    assert isinstance(
```
nirga, line 57: I think it's better to send those in events (see an [example from Pinecone](https://github.com/traceloop/openllmetry/blob/3f6d978a30516c0d1684ef86f2d0e1c8c7a7ab7b/packages/opentelemetry-instrumentation-pinecone/opentelemetry/instrumentation/pinecone/query_handlers.py#L60)

nirga (commented):
Thanks! Note that tests seem to be failing

PR #1539, opened 2024-07-11
fix(openai): log content filter results in proper attributes

### LINKED ISSUE #855: 🚀 Feature: log content filter results in proper attributes (Azure OpenAI)

jonhsma:
Fixes #855 


jonhsma:
The lint failure was reproduced locally. I fixed it and was able to run lint locally with no issue with the latest commit

nirga (commented):
Thanks @jinsongo! can you add a test for that?

jinsongo:
> Thanks @jinsongo! can you add a test for that?

Thanks @jonhsma! can you add a test for that?  ^_^

nirga:
🙈😅

jonhsma:
> > Thanks @jinsongo! can you add a test for that?
> 
> Thanks @jonhsma! can you add a test for that? ^_^

Sure. Is there any guide on writing test? If not I will try to figure it out..

jonhsma:
I tried to test it locally. It seems that this type of response could only come from Azure OpenAI and it's impossible for me to get an API key for that. I think I saw in the docs that simulated responses are used for these tests but I can't find that page anymore. Is there any suggestions? Thanks!

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -777,11 +777,16 @@ def _accumulate_stream_items(item, complete_response):
...
+            complete_choice["content_filter_results"] = choice.get("content_filter_results")
 
         delta = choice.get("delta")
 
         if delta and delta.get("content"):
-            complete_choice["message"]["content"] += delta.get("content")
+            if not (choice.get("finish_reason") and complete_choice["finish_reason"] == "content_filter"):
+                complete_choice["message"]["content"] += delta.get("content")
+            else:
+                complete_choice["message"]["content"] += "FILTERED"
```
nirga, line 789: This is duplication, see method `_set_completions`

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -777,11 +777,16 @@ def _accumulate_stream_items(item, complete_response):
         complete_choice = complete_response.get("choices")[index]
         if choice.get("finish_reason"):
             complete_choice["finish_reason"] = choice.get("finish_reason")
+        if choice.get("content_filter_results"):
```
nirga, line 780: this is only part of the solution - as it will only work for streaming results. You need to update this in more places - see for example `_set_completions`. 

Can you also verify me that this is how this field is returned in streaming? Can you link to the docs from Azure?

nirga (commented):
Reg. the tests - we already have a test for this, you just need to update it with noew assertions - see test_chat_content_filtering in `tests/traces/test_azure`. You can basically add your assertions there

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
jonhsma: I see. I'll go through the rest of chat_wrappers and let you know if I have other questions.

Here is an example of how the filed is returned in streaming [link](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#sample-response-stream-blocked-by-filters). I think the field is _content_filter_results_. Also it seems that _content_filter_results_ won't be returned for non-streaming prompts [link](https://platform.openai.com/docs/guides/chat-completions/response-format) but there is no harm to preempt that.

I do notice that it might be  AzureAI specific. In open AI docs, _content_filter_ is one of the finish reasons but there is no mentioning of _content_filter_results [link](https://platform.openai.com/docs/guides/chat-completions/response-format). I don't think it's gonna cause a problem though

nirga: I think this is the example for non-streaming:
https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#example-scenario-an-input-prompt-containing-content-that-is-classified-at-a-filtered-category-and-severity-level-is-sent-to-the-completions-api

jonhsma: You're right 😅 I pasted the wrong link. Sounds like we want to handle filtered prompts as well. I think I have an idea how to take care of that as well..

jonhsma:
This is still a work in progress. It passed the tests locally.

The latest commit should be able to (🤞) handle both streaming and non-streaming filtered results like these
- https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#sample-response-stream-blocked-by-filters
- https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#scenario-your-api-call-asks-for-multiple-responses-n1-and-at-least-one-of-the-responses-is-filtered

but it won't handle situations like this where the input is filtered
- https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#example-scenario-an-input-prompt-containing-content-that-is-classified-at-a-filtered-category-and-severity-level-is-sent-to-the-completions-api

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -371,9 +371,26 @@ def _set_completions(span, choices):
...
+                print(object)
+                for key, item in object.items():
+                    if isinstance(item, dict):
+                        _cast_into_span(span, prefix + f".{key}", item)
+                    else:
+                        _set_span_attribute(span, f"{prefix}.{key}", item)
+
+            for category, item in choice.get("content_filter_results").items():
+                # This handels known possible outcomes including filtering errors
+                _cast_into_span(span, f"{prefix}.content_filter_results.{category}", item)
```
nirga, line 388: I think it's better to just send everything in a stringified JSON in the `content_filter_results` attribute

nirga (commented):
Thanks! I think if we're doing this we might as well handle all cases - including input filtering and streaming filtering (which you don't handle yet IIUC)

jonhsma:
It seems that both streaming and non-streaming calls would eventually run _set_completions()? So in that case as long as I collect content_filter_results in _accumulate_stream_items(), it will be passed in to _set_completions() in _build_from_streaming_response()?

Here is what I have in _accumulate_stream_items()
https://github.com/traceloop/openllmetry/blob/de2c882e5111dc5bb64326345775c0ef3db89800/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L797-L798

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
jonhsma: now it is and the tests are updated

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -687,6 +701,9 @@ def _build_from_streaming_response(
 
     _set_response_attributes(span, complete_response)
 
+    # wprompt_filter_results
```
nirga, line 704: comment not needed

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -270,6 +272,8 @@ def _handle_response(
 
     # span attributes
     _set_response_attributes(span, response_dict)
+    # prompt_filter_results
+    _log_prompt_filter(span, response_dict)
```
nirga, line 276: why is this a separate method and not part of `_set_response_attributes`?

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -617,6 +628,9 @@ def _close_span(self):
 
         _set_response_attributes(self._span, self._complete_response)
 
+        # prompt_filter_results
```
nirga, line 631: comment not needed

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
```diff
@@ -371,9 +375,16 @@ def _set_completions(span, choices):
             span, f"{prefix}.finish_reason", choice.get("finish_reason")
         )
 
+        if choice.get("content_filter_results"):
+            _set_span_attribute(
+                span, f"{prefix}.content_filter_results",
```
nirga, line 380: let's extract this to a const as well

packages/opentelemetry-instrumentation-openai/tests/traces/.test_chat.py.swp
nirga, line 1: file added to PR by mistake?

nirga (commented):
Thanks @jonhsma! Left a couple of small comments

packages/opentelemetry-instrumentation-openai/tests/traces/.test_chat.py.swp
jonhsma: 🙈 let me drop it...

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
jonhsma: Did that, but the constant is now in the scope of the file. Are we considering moving it somewhere else?

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
jonhsma: 🤣 Just realize they're like twins. Moved it into _set_response_attributes() in __init__

nirga (approved):

Issue #540, opened 2024-02-27
🚀 Feature: re-write LlamaIndex instrumentation to use LlamaIndex `CallbackManager`


nirga:
### Which component is this feature for?

LlamaIndex Instrumentation

### 🔖 Feature description

Right now, we monkey-patch classes and methods in LlamaIndex which requires endless work and constant maintenance. LlamaIndex has a system for callbacks that can potentially be used to create/end spans without being too coupled with with the framework's inner structure.

### 🎤 Why is this feature needed ?

Support LlamaIndex entirely and be future-proof to internal API changes

### ✌️ How do you aim to achieve this?

Look into LlamaIndex `callback_manager` and how other frameworks are using it.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

varaarul:
I can start looking into this!

larinam:
@nirga , could you fix the link to documentation in the description and give a hint what pieces of this project's code should be touched by this change?

nirga:
Done @larinam! 
Basically, right now we wrap LlamaIndex calls to log and create spans. This means we have to keep up with their APIs and update our instrumentations every time something changes on LlamaIndex's end. 

A better solution would be to register ourselves to the callbacks of the LlamaIndex engine and just open and close spans when we get called. This is much more robust and future proof.

larinam:
All right @nirga , thanks. But I'll try to ask one more time with the question. 
The code base is not that little for the project, could you give a hint what pieces of this project's code should be touched by this change? Could you just attach a couple of links?

nirga:
In our project it's just under the LlamaIndex instrumentation. Do you want to join our [slack](https://traceloop.com/slack) and discuss there? Might be easier :)

Issue #1541, opened 2024-07-11
🐛 Bug Report: Spans being detached from Langchain parent traces


maciejwie:
### Which component is this bug for?

Langchain Instrumentation

### 📜 Description

Calls to functions instrumented by other libraries get detached from the parent trace. This happens for example in Langchain chains which call a retriever which is instrumented by opentelemetry.instrumentation.sqlalchemy. The nominally child spans are created but are independent of each other and of the actual parent.

Occurs for both sync and async calls.

### 👟 Reproduction steps

This code reproduces the issue given a local postgres server:
```
import os

from langchain_core.prompts import PromptTemplate
from langchain_core.output_parsers import StrOutputParser
from langchain_core.runnables import RunnablePassthrough
from langchain_openai import ChatOpenAI, OpenAIEmbeddings
from langchain_postgres import PGVector
from opentelemetry.sdk.trace.export import ConsoleSpanExporter
from traceloop.sdk import Traceloop


Traceloop.init(
    exporter=ConsoleSpanExporter(),
)

PROMPT_TEMPLATE = """
You are an assistant for question-answering tasks. Use the following pieces of retrieved context to answer the question.

Question: {question} 

Context: {context} 

Answer:
"""

qa_prompt = PromptTemplate.from_template(PROMPT_TEMPLATE)

llm = AzureChatOpenAI()

embeddings = OpenAIEmbeddings()

vectorstore = PGVector(
    connection=PGVector.connection_string_from_db_params(
        driver="psycopg",
        host="localhost",
        port=5432,
        database="vectors",
        user="postgres",
        password="localpassword"
    ),
    embeddings=embeddings,
)

def format_docs(docs):
    return "\n\n".join(doc.page_content for doc in docs)

qa_chain = (
    {
        "context": vectorstore.as_retriever() | format_docs,
        "question": RunnablePassthrough(),
    }
    | qa_prompt | llm | StrOutputParser()
)

qa_chain.invoke("What are autonomous agents?")

```

### 👍 Expected behavior

All spans in the chain should have the same `trace_id` and appropriate `parent_id`.

### 👎 Actual Behavior with Screenshots

Example:
```
{
    "name": "RunnableAssign<filter>.langchain.task",
    "context": {
        "trace_id": "**0x2b1a7979f5725daf1a624b67a66c0d65**",
        "span_id": "0xff0ed485230f3387",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
[...]

PR #1546, opened 2024-07-13
feat(llama-index): Use callbacks

tibor-reiss:
Related to #540

https://docs.llamaindex.ai/en/stable/module_guides/observability/callbacks/

Todo:
- [ ] Add more tests
- [ ] Check all previous instrumentors and move them to callbacks

Checks:


tibor-reiss:
@nirga This is similar to the previous work on langchain-callbacks. Let me know please if you agree with the direction!

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
```diff
@@ -23,17 +24,49 @@ def multiply(a: int, b: int) -> int:
...
-        "FunctionTool.tool",
-        "openai.chat",
-    } == set([span.name for span in spans])
-
-    agent_span = next(span for span in spans if span.name == "ReActAgent.agent")
-    function_tool = next(span for span in spans if span.name == "FunctionTool.tool")
-    openai_span = next(span for span in spans if span.name == "openai.chat")
+        "agent_step.llama_index.workflow",
+        "llm.llama_index.task",
+        "function_call.llama_index.task",
```
nirga, line 29: looks like openai is missing here?

nirga (commented):
Nice! I also started working on this from a different angle - #1524 

The main thing I worry about this (and this is what I also commented about) - is that this messes up the traces. See also #1541 on a new bug we have with callbacks on LangChain

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
tibor-reiss: I am not sure if that is the case: the llm calls are under llm.llama_index.task, e.g. see L40: `assert llm_span_1.attributes[SpanAttributes.LLM_REQUEST_MODEL] == "gpt-3.5-turbo-0613"`.

Imo, there is the question if the OpenAi instrumentor is needed if one uses langchain. On one side not because langchain is supposed to take care of that. On the other hand, if one explicitly uses the OpenAI instrumentor, then the expectation is to see these traces.

tibor-reiss:
> Nice! I also started working on this from a different angle - #1524
> 
> The main thing I worry about this (and this is what I also commented about) - is that this messes up the traces. See also #1541 on a new bug we have with callbacks on LangChain

I see what you mean. It seems to be a question how langchain calls the LLMs (including OpenAI) - see also comment on the specific test.

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
tibor-reiss: After some digging: openai/shared/chat_wrappers gets called, but context_api.get_value(SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY) is True - this was added recently, wasn't it? I wonder if it's just a new env variable which needs to be set in conftest...

tibor-reiss: This is it - and I copied it: https://github.com/traceloop/openllmetry/pull/1452/files#diff-b5474fc40399d3e9320e536a98b121519743b68e5b7271884a8b853052a5ffe8R273

@nirga Do you remember why you are setting this value?

tibor-reiss: I replaced it now with current_context. Now the LLM spans look ok. There is even the test_chroma_vector_store where there are 2 spans outside of the engine - which is correct imho.

nirga: The reason I did that is because we report an OpenAI span now within LangChain so I didn't want to report the same data twice (if using LangChain). It's only turned on within LangChain but otherwise defaults to off. Not sure I understood what you've done differently to make it work?

tibor-reiss: This is only about your comment "looks like openai is missing here?". My understanding is that langchain and llama are both frameworks, but apparently they have some differences, e.g. how they handle openai calls.

Regarding the bigger issue in #1541: that's still not solved.

nirga (commented):
Thanks @tibor-reiss this looks good now! One thing I'd change is the way you instrument - I'd do something similar to [my PR](https://github.com/traceloop/openllmetry/pull/1524) and use the [new Instrumentation modules for new LlamaIndex versions](https://docs.llamaindex.ai/en/stable/module_guides/observability/instrumentation/) (and just keep using what we have now for older versions). WDYT?

tibor-reiss:
> Thanks @tibor-reiss this looks good now! One thing I'd change is the way you instrument - I'd do something similar to [my PR](https://github.com/traceloop/openllmetry/pull/1524) and use the [new Instrumentation modules for new LlamaIndex versions](https://docs.llamaindex.ai/en/stable/module_guides/observability/instrumentation/) (and just keep using what we have now for older versions). WDYT?

Hi @nirga, thanks for the tip, I was not aware of this change from callbacks to instrumentation. I started it now, and ended up with something very similar to yours. However, I still have the issue that the order of the spans is weird - need to look into this further...

tibor-reiss:
> > Thanks @tibor-reiss this looks good now! One thing I'd change is the way you instrument - I'd do something similar to [my PR](https://github.com/traceloop/openllmetry/pull/1524) and use the [new Instrumentation modules for new LlamaIndex versions](https://docs.llamaindex.ai/en/stable/module_guides/observability/instrumentation/) (and just keep using what we have now for older versions). WDYT?
> 
> Hi @nirga, thanks for the tip, I was not aware of this change from callbacks to instrumentation. I started it now, and ended up with something very similar to yours. However, I still have the issue that the order of the spans is weird - need to look into this further...

I have created https://github.com/run-llama/llama_index/pull/14997 which will hopefully make the spans in test_query_pipeline fall in line. The test will fail once that is merged, but I think that's good and we can fix it then.

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
```diff
@@ -23,17 +24,57 @@ def multiply(a: int, b: int) -> int:
...
-    openai_span = next(span for span in spans if span.name == "openai.chat")
+    } == {span.name for span in spans}
+
+    agent_workflow_span = next(span for span in spans if span.name == "AgentRunner.llama_index.workflow")
+    function_tool_span = next(span for span in spans if span.name == "FunctionTool.llama_index.task")
+    llm_span_1, llm_span_2 = [span for span in spans if span.name == "OpenAI.llama_index.task"]
+    openai_span_1, openai_span_2 = [span for span in spans if span.name == "openai.chat"]
+
+    assert agent_workflow_span.parent is None
+    assert function_tool_span.parent is not None
```
nirga, line 42: can we assert on the specific structure here? for example that the function tool span's parent is the agent workflow? (or whatever it actually is)

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
```diff
@@ -90,19 +131,42 @@ def test_agent_with_query_tool(exporter):
...
+        span for span in spans if span.name == "OpenAIAssistantAgent.llama_index.workflow"
     )
-    query_engine_tool = next(
-        span for span in spans if span.name == "QueryEngineTool.tool"
+    synthesize_span = next(span for span in spans if span.name == "BaseSynthesizer.llama_index.task")
+    llm_span_1, llm_span_2 = [span for span in spans if span.name == "OpenAI.llama_index.task"]
+    openai_span_1, openai_span_2 = [span for span in spans if span.name == "openai.chat"]
+
+    assert agent_span.parent is None
+    assert synthesize_span.parent is not None
```
nirga, line 149: same for this - assert actual parents

packages/opentelemetry-instrumentation-llamaindex/tests/test_query_pipeline.py
```diff
@@ -44,18 +45,52 @@ def test_query_pipeline(exporter):
     spans = exporter.get_finished_spans()
 
     assert {
-        "llama_index_query_pipeline.workflow",
-        "retrieve.task",
-        "synthesize.task",
+        # "QueryPipeline.llama_index.workflow",
```
nirga, line 48: the comment is because the change you pushed to llamaindex? I'd rather write this test as if your PR is already in, and disable this entire test temporarily with a comment as to why (+link to your PR)

nirga (approved):
Thanks so much for this @tibor-reiss! I've approved the PR since all of my comments are nits but appreciate if you have the time to take a look :) 

Thanks again!

packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/dispatcher_wrapper.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def prepare_to_exit_span(
+        self, id_: str, result: Optional[Any] = None, **kwargs
+    ) -> SpanHolder:
+        """Logic for preparing to drop a span."""
+        span_holder = self.open_spans[id_]
+        span_holder.span.end()
+        context_api.detach(span_holder.token)
+        with self.lock:
+            self.completed_spans += [span_holder]
```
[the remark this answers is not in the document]
tibor-reiss: Technically you are right, can be deleted. However, logically it makes sense (similarly to SimpleSpanHandler from llama-index): it opens up possibilities for some more sophisticated logic. Please feel free to delete it!

packages/opentelemetry-instrumentation-llamaindex/tests/test_agents.py
tibor-reiss: Yes, I wanted to do this. But the test would become quite brittle. Below is the span structure for this test. It's very verbose, lot of same named spans. E.g. FunctionTool, the one you mentioned, has ReActAgentWorker as parent, but that pops up 4 times. I am still undecided what is the best approach here - maybe there is a third better option? So please feel free to change it to your liking.

name span_id parent.span_id
ReActAgentWorker.llama_index.task 319464811280482099 7923477847642805764
AgentRunner.llama_index.task 7923477847642805764 5348156688623149641
openai.chat 10833227178492975942 787045099616181983
OpenAI.llama_index.task 787045099616181983 13446663969238565076
ReActOutputParser.llama_index.task 1599754872828344697 13446663969238565076
FunctionTool.llama_index.task 10958318651987115747 13446663969238565076
ReActAgentWorker.llama_index.task 13446663969238565076 15738578304360324053
AgentRunner.llama_index.task 15738578304360324053 5348156688623149641
openai.chat 12351434778324108193 14394044961538896963
OpenAI.llama_index.task 14394044961538896963 18062227453713671516
ReActOutputParser.llama_index.task 17662053697512896373 18062227453713671516
ReActAgentWorker.llama_index.task 18062227453713671516 8217273878331094304
AgentRunner.llama_index.task 8217273878331094304 5348156688623149641
ReActAgentWorker.llama_index.task 7685564445797887478 11969873770185106285
AgentRunner.llama_index.task 11969873770185106285 5348156688623149641
AgentRunner.llama_index.task 5348156688623149641 10450133951577012365
AgentRunner.llama_index.workflow 10450133951577012365 None

packages/opentelemetry-instrumentation-llamaindex/tests/test_query_pipeline.py
tibor-reiss: Wow, PR has been already merged yesterday, so it should be available soon hopefully. I have adjusted to test.

packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/dispatcher_wrapper.py
```diff
@@ -0,0 +1,144 @@
...
+        span.set_attribute(SpanAttributes.TRACELOOP_ENTITY_NAME, span_name)
+        return SpanHolder(span, token, current_context)
+
+    def prepare_to_exit_span(
+        self, id_: str, result: Optional[Any] = None, **kwargs
+    ) -> SpanHolder:
+        """Logic for preparing to drop a span."""
+        span_holder = self.open_spans[id_]
+        span_holder.span.end()
+        context_api.detach(span_holder.token)
```
tibor-reiss, line 114: @nirga I added this line as an experiment to see if the spans behave well, but I don't understand it fully. Could you please explain why this is needed?
Without this, the previous span - even if no connection (and `span.end()` already called) - will become the parent when executing `span = self._tracer.start_span(`, which is due to `opentelemetry.sdk.trace.__init__`:
```
parent_span_context = trace_api.get_current_span(
            context
        ).get_span_context()
```
There is a `opentelemetry.context.__init__.detach` call in span.end(), but it's not enough. Is this maybe related to your comment of different spans in opentelemetry and openllmetry? https://github.com/traceloop/openllmetry/issues/1541#issuecomment-2230378905

nirga:
@tibor-reiss just ran this against the sample app `poetry run python sample_app/llama_index_chroma_app.py` and got an exception:
```
`  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/sdk/trace/export/__init__.py", line 367, in _export_batch
    self.span_exporter.export(self.spans_list[:idx])  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 167, in export
    serialized_data = self._serialize_spans(spans)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 130, in _serialize_spans
    return encode_spans(spans).SerializePartialToString()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 56, in encode_spans
    resource_spans=_encode_resource_spans(sdk_spans)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 79, in _encode_resource_spans
    pb2_span = _encode_span(sdk_span)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 120, in _encode_span
    kind=_SPAN_KIND_MAP[sdk_span.kind],
         ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'task'
```

tibor-reiss:
> @tibor-reiss just ran this against the sample app `poetry run python sample_app/llama_index_chroma_app.py` and got an exception:
> 
> ```
> `  File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/sdk/trace/export/__init__.py", line 367, in _export_batch
>     self.span_exporter.export(self.spans_list[:idx])  # type: ignore
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 167, in export
>     serialized_data = self._serialize_spans(spans)
>                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py", line 130, in _serialize_spans
>     return encode_spans(spans).SerializePartialToString()
>            ^^^^^^^^^^^^^^^^^^^
>   File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 56, in encode_spans
>     resource_spans=_encode_resource_spans(sdk_spans)
>                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 79, in _encode_resource_spans
>     pb2_span = _encode_span(sdk_span)
>                ^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/nirga/vecinity/openllmetry/packages/sample-app/.venv/lib/python3.11/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/trace_encoder/__init__.py", line 120, in _encode_span
>     kind=_SPAN_KIND_MAP[sdk_span.kind],
>          ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
> KeyError: 'task'
> ```

I think it should be fixed with the last commit - could you please try @nirga?
I added a test which only uses free models - this is what I tested and there were no errors. I also added the new llama-index Settings (instead of ServiceContext).

nirga:
Thanks @tibor-reiss I think we're almost there with this PR, so thank you for your great job :) 
Right now I see 3 main obstacles for merging this:

1. We need to log token usage and model data
Screenshot 2024-08-02 at 16 53 45

2. We need to turn off the specific instrumentations of packages like OpenAI, etc. ([like we do in Langchain](https://github.com/traceloop/openllmetry/blob/247b72b7e9896a39dc45c53ff138b9aef9e095c0/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_wrapper.py#L311))

3. We need to log input and outputs of tasks - [as we used to do in the old way](https://github.com/traceloop/openllmetry/blob/247b72b7e9896a39dc45c53ff138b9aef9e095c0/packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/utils.py#L72).

tibor-reiss:
> Thanks @tibor-reiss I think we're almost there with this PR, so thank you for your great job :) Right now I see 3 main obstacles for merging this:
> 
>     1. We need to log token usage and model data
>        Screenshot 2024-08-02 at 16 53 45
> 
>     2. We need to turn off the specific instrumentations of packages like OpenAI, etc. ([like we do in Langchain](https://github.com/traceloop/openllmetry/blob/247b72b7e9896a39dc45c53ff138b9aef9e095c0/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_wrapper.py#L311))
> 
>     3. We need to log input and outputs of tasks - [as we used to do in the old way](https://github.com/traceloop/openllmetry/blob/247b72b7e9896a39dc45c53ff138b9aef9e095c0/packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/utils.py#L72).

1. It seems that the usage/model is dependent on the package versions - I added some more code (e.g. see `# raw can be Any, not just ChatCompletion`). I also added more asserts to the tests. Additionally, the new sample app (with the LLM HuggingFaceH4/zephyr-7b-alpha) does not trigger the chat-like events and I have not found these metrics on the other events. 
2. I added the same logic
3. Still todo...

tibor-reiss:
Hi @nirga, did you have the chance to check the usage and model data with openai using the latest version, please? Is there anything else left? If not, I would recommend to merge this, and I would then start working on a new PR to extract even more information from the various events.

nirga:
Sorry for the delay @tibor-reiss I was traveling. I'm a bit hesitant because the missing inputs and outputs which might be crucial, let me see if I can fix it.

tibor-reiss:
> Sorry for the delay @tibor-reiss I was traveling. I'm a bit hesitant because the missing inputs and outputs which might be crucial, let me see if I can fix it.

@nirga I am not sure what you mean. I have added a first version earlier, here the output from the new sample app. Could you please check the old sample app which uses OpenAI how it looks now?

image

PR #1620, opened 2024-07-21
feat(sdk): allow disablign prompt sending as argument to Traceloop.init()

liamjdavis:

### DIFF
--- a/packages/traceloop-sdk/tests/test_disabling_prompt_feature.py
+++ b/packages/traceloop-sdk/tests/test_disabling_prompt_feature.py
@@ -1,0 +1,51 @@
+import os
+
+import pytest
+
+from traceloop.sdk import Traceloop
+from openai import OpenAI
+from opentelemetry.semconv.ai import SpanAttributes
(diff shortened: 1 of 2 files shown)


nirga (commented):
Thanks @liamjdavis! I have to say I'm less of a fan of changing an environment for a user - this may have some unexpected results. I think a better solution is to add this to the configuration of all instrumentations and then pass the value during initialization, like we do here:
https://github.com/traceloop/openllmetry/blob/2b337d3600863e15f5cc5b37de149f60f11b5a30/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L546

liamjdavis:
@nirga I have started changing it to changing the configuration of all instrumentation, but I am unsure of what to change in the configurations, particularly where the only parameter beyond what's in the BaseInstrumentor is the exception_logger. Can you help me here?

nirga:
@liamjdavis you should add a new parameter to the instrumentations which controls whether prompt sending is enabled or disabled.

liamjdavis:
Thank you, I'll start working on it.

PR #1957, opened 2024-09-12
feat(bedrock): support metrics for bedrock

jinsongo:


jinsongo:
@nirga Could you please advise why I always get the following error, even though I tried changing the model ID? Thanks.

https://github.com/traceloop/openllmetry/actions/runs/10912918948/job/30288462269?pr=1957
`botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the InvokeModel operation: You don't have access to the model with the specified model ID.`    cc @gyliu513

jinsongo:
I already know the root cause why I always getting the error `botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the InvokeModel operation: You don't have access to the model with the specified model ID.`. Thanks.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -29,6 +29,8 @@ class Meters:
     LLM_WATSONX_COMPLETIONS_RESPONSES = "llm.watsonx.completions.responses"
     LLM_WATSONX_COMPLETIONS_TOKENS = "llm.watsonx.completions.tokens"
 
+    LLM_BEDROCK_COMPLETIONS_EXCEPTIONS = "llm.bedrock.completions.exceptions"
```
gyliu513, line 32: Is this required?

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -62,36 +89,72 @@ def _set_span_attribute(span, name, value):
 def _with_tracer_wrapper(func):
     """Helper for providing tracer for wrapper functions."""
 
-    def _with_tracer(tracer, to_wrap):
+    def _with_tracer(
+        tracer,
+        metricParams,
```
gyliu513, line 94: If you are adding a new parameter, I prefer put it as last one.

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -62,36 +89,72 @@ def _set_span_attribute(span, name, value):
...
         return wrapper
 
     return _with_tracer
 
 
 @_with_tracer_wrapper
-def _wrap(tracer, to_wrap, wrapped, instance, args, kwargs):
+def _wrap(
+    tracer,
+    metricParams,
```
gyliu513, line 116: ditto

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -476,6 +589,34 @@ def _set_amazon_span_attributes(span, request_body, response_body):
             )
 
 
+def _create_metrics(meter: Meter):
```
gyliu513, line 592: There are duplicate code in anthropic and groq, we may need a clean up in future, like creating a common shared lib for the LLM instrumentation. @nirga ^^

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
jinsongo: It's an open issue. Actually, I think it's better to use a common name, for example `llm.completions.exceptions` for all ai system to count exceptions. If yes, I would like to use another PR to fix all related code.

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
jinsongo: I just want to put the metric parameters near by tracer and it's an internal function only.

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
nirga: Yep! I agree
https://github.com/traceloop/openllmetry/issues/1997

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
nirga: @jinsongo if we're doing this here I'd try to use the new genAI conventions - https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics
I think they're already released so you can even use them directly from the original otel semconv package

nirga: (in this case - it means exceptions are counted in the duration metric)

jinsongo: @nirga I cannot find an appropriate metric name for the exceptions counter of Bedrock. As you know, version 0.4.1 of opentelemetry-semantic-conventions-ai does not include `LLM_BEDROCK_COMPLETIONS_EXCEPTIONS` although it's already included in this current PR commits, which causes the build to fail when I use it in the Bedrock instrumentation code. As you know, that's the same approach for OpenAI, Authropic and Watsonx.

nirga: @jinsongo this should be in the standard `opentelemetry-semantic-conventions` package. And you should count exceptions on the duration metrics as specified in the spec.

gyliu513: @jinsongo If this parameter is not required, I think we can ignore this for now, and we can fix this in future as a consolidation for semantic convention.

jinsongo: @gyliu513 I have removed `LLM_BEDROCK_COMPLETIONS_EXCEPTIONS` from `semconv_ai`, and added a `TODO` comment about fixing in future as a consolidation for semantic convention.

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
jinsongo: Will fix with another PR once the code for https://github.com/traceloop/openllmetry/issues/1997 is ready.

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -27,9 +29,30 @@
...
+    def __init__(
+        self,
+        token_histogram: Histogram,
+        choice_counter: Counter,
+        duration_histogram: Histogram,
+        exception_counter: Counter,
+    ):
+        self.vendor = ""
+        self.model = ""
+        self.isStream = False
```
nirga, line 48: nit: `is_stream`

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
nirga, line 94: nit: `metric_params`

nirga (commented):
nice @jinsongo! a couple of styling comments and we're ready to go

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
jinsongo: Done

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
jinsongo: Done

gyliu513:
/lgtm

@nirga can you help review? Thanks

jinsongo:
@nirga Do you have more comments? Thanks.

nirga (approved):

PR #1958, opened 2024-09-13
fix(sdk): support a "block-list" of things to not instrument

codehruv:
Need help here. There are no specific tests checking for `requests`. Where should I add the test?

Could you give a recommendation on how to do this with our `HoneyHiveTracer`? 

Is there a particular example script that would work?

TBD, once the tests are running


codehruv:
One more note, if someone specifically asks for requests package instrument, I'm not considering the disable behavior.

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -62,6 +62,7 @@ class TracerWrapper(object):
     def __new__(
         cls,
         disable_batch=False,
+        disable_requests_instrumentation=True,
```
nirga, line 65: this seems highly specific, don't you think? I think I'd rather add something like the "instruments" parameter but that can be sort of a "block list" where any of the libraries in that parameter **won't** be instrumented.

nirga (commented):
Thanks! I think there's a better way to do it

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
codehruv: Block list sounds like a better idea. The reason I went specific was because the tracer is normally expected to instrument GenAI-related packages, but has `requests` as a special additional scenario. Block list would be broad enough for future developments as well.

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -525,35 +533,65 @@ def init_tracer_provider(resource: Resource) -> TracerProvider:
...
-    init_bedrock_instrumentor(should_enrich_metrics)
-    init_replicate_instrumentor()
-    init_vertexai_instrumentor()
-    init_watsonx_instrumentor()
-    init_weaviate_instrumentor()
-    init_alephalpha_instrumentor()
-    init_marqo_instrumentor()
-    init_lancedb_instrumentor()
-    init_groq_instrumentor()
+def init_instrumentations(should_enrich_metrics: bool, block_instruments: Optional[Set[Instruments]] = None):
```
nirga, line 536: ideally this might be an opportunity to combine these 2 ways and just have `init_instrumentations` work the same as the `if instruments is not None` clause - just set all instrumentations as the value of `instruments`. WDYT?

packages/traceloop-sdk/traceloop/sdk/__init__.py
```diff
@@ -44,6 +44,7 @@ def init(
         api_key: str = None,
         headers: Dict[str, str] = {},
         disable_batch=False,
+        disable_requests_instrumentation: bool = True,
```
nirga, line 47: remove?

nirga (approved):

PR #1996, opened 2024-09-19
Structured output for openai

AnsariZayd:
Removed the testenv file and now using pytest.


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/utils.py
```diff
@@ -88,6 +88,10 @@ def wrapper(wrapped, instance, args, kwargs):
                 args,
                 kwargs,
             )
+            span = tracer.start_span(name="openai.response")
```
nirga, line 91: this should be on the original span and not on a new one

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/unitTest.py
```diff
@@ -0,0 +1,69 @@
+import unittest
+import logging
+from unittest.mock import Mock
+from utils import handle_structured_output  # Ensure this import is correct
+
+# Configure logging to write to a file
+logging.basicConfig(
```
nirga, line 7: no need for this

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/unitTest.py
```diff
@@ -0,0 +1,69 @@
...
+    filename='C:\\openai new feature on onpensource\\openllmetry\\packages\\opentelemetry-instrumentation-openai\\opentelemetry\\instrumentation\\openai\\app.log',  # Name of the log file
+    level=logging.INFO,  # Set the logging level
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+)
+
+logging.info("Logging configuration set up.")
+
+class TestHandleStructuredOutput(unittest.TestCase):
```
nirga, line 15: this is not pytest - please take a look at existing tests and extend

nirga (commented):
hey @AnsariZayd thanks but we should try to stay as close as we can to the current code. Left some comments :)

nirga:
Fixed with #2111

PR #2000, opened 2024-09-22
feat(sdk+openai): support base64 images upload

galkleinman:


packages/sample-app/sample_app/openai_vision_base64_example.py
```diff
@@ -0,0 +1,84 @@
...
+
+api_key = os.environ.get("OPENAI_API_KEY")
+
+
+def encode_image(image_path):
+    with open(image_path, "rb") as image_file:
+        return base64.b64encode(image_file.read()).decode('utf-8')
+
+
+image_path = "/Users/galklm/Downloads/elephant.jpeg"
```
nirga, line 16: nit: maybe add an image together with the sample app and point to it here so people can actually run the sample app and see it in action?

packages/traceloop-sdk/tests/test_image_uploader.py
```diff
@@ -0,0 +1,25 @@
+import nest_asyncio
```
nirga, line 1: fyi: I think it's an unmaintained repo that has been archived recently:
https://github.com/erdewit/nest_asyncio

nirga (approved):

packages/traceloop-sdk/tests/test_image_uploader.py
galkleinman: the issue is that is solves an issue with a test i added... look at the last commit (or the one before), I specifically added it to the tests to resolve an issue.

packages/sample-app/sample_app/openai_vision_base64_example.py
galkleinman: will do

Issue #389, opened 2024-02-02
🚀 Feature: 


mohitk1995:
### Which component is this feature for?

OpenAI Instrumentation

### 🔖 Feature description

 Want to log a variable's value in the trace of llm. Where llm is called from a async task. Currently the association_properties are shared by all the tasks. 
So basically a way to report a variable's value. 

### 🎤 Why is this feature needed ?

To add more details to the LLM trace that are specific to each llm call made in the same context.

### ✌️ How do you aim to achieve this?

Give a way to log variables inside a llm traces. This should only be set for local scope of the function or the linked functions/processes. Should not be updated globally.

### 🔄️ Additional Information

I tried setting the association_properties within each task those were called from a common parent. But the variable's value was getting overwritten globally. Each time I was getting the most recent value.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

varaarul:
I'll look into this

PR #2085, opened 2024-10-03
Update to Issue #389- Added ContextVars variable "association_properties_var"

Daniel-Plops:
<-- Thanks for submitting a PR!  -->

Added a ContextVars variable to the TracerWrapper object such that it updates every time set_association_properties is called such that it stores a copy of the properties in a variable called association_properties_var, so changes are only stored when using the set_association_properties function. This allows for the properties to be viewed at certain instances in between trace updates so other functions could be made possible using this backup of association_properties.


nirga (commented):
Thanks @Daniel-Plops! I'm not sure what is the use case you're looking to resolve with this?

PR #2086, opened 2024-10-03
fix(anthropic): Replace count_tokens with usage for newer models

### LINKED ISSUE #2061: 🚀 Feature: Use usage attribute instead of anthropic.count_tokens()

Anthropic Instrumentation

### 🔖 Feature description

Getting more accurate numbers on token usage, including the token cost of images, by using the usage attribute from the anthropic response.

The description of the function says that the function is only accurate for older models such as Claude 2.1. I think the majority of people use the later 3.X models for which this function gives a very rough estimate. Anthropic say that the usage attribute on the response should be used instead.

```python
def count_tokens(
        self,
        text: str,
    ) -> int:
        """Count the number of tokens in a given string.

        Note that this is only accurate for older models, e.g. `claude-2.1`. For newer
        models this can only be used as a _very_ rough estimate, instead you should rely
        on the `usage` property in the response for exact counts.
        """
        # Note: tokenizer is untyped
        tokenizer = self.get_tokenizer()
        encoded_text = tokenizer.encode(text)  # type: ignore
        return len(encoded_text.ids)  # type: ignore
```

Update the anthropic instrumentation to use the token count from the usage attribute of the response instead of calling the count_token function.

### 🔄️ Additional Information

_No response_

None

tibor-reiss:
Fixes #2061

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -182,23 +182,26 @@ async def _aset_token_usage(
     if not isinstance(response, dict):
         response = response.__dict__
 
-    prompt_tokens = 0
-    if hasattr(anthropic, "count_tokens"):
-        if request.get("prompt"):
-            prompt_tokens = await anthropic.count_tokens(request.get("prompt"))
-        elif request.get("messages"):
-            prompt_tokens = 0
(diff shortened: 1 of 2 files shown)


tibor-reiss:
@nirga What do you think, for older models, shall we drop the `count_tokens` completely and return 0 (with/without some warning), or keep the old code as a fallback (as is now in the commit)?

nirga (approved):
I think this is the best way. Thanks @tibor-reiss!

Issue #1815, opened 2024-08-06
🚀 Feature: support structured outputs for OpenAI


nirga:
### Which component is this feature for?

OpenAI Instrumentation

### 🔖 Feature description

Add support to OpenAI's recently announced structured outputs:
https://openai.com/index/introducing-structured-outputs-in-the-api/
For example, support `refusal` responses

### 🎤 Why is this feature needed ?

-

### ✌️ How do you aim to achieve this?

-

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

Benedikt-Wolf:
I'll have a look at it

iarslankhalid:
> I'll have a look at it, if no one is working on this currently?

hey @Benedikt-Wolf, is there any update regarding this feature?

Benedikt-Wolf:
Hey @iarslankhalid no, not at all😅I am in a bit of a stress rn (last exam, job interviews, ...), so if u want to take over, please do so 👍 I will probably not have time in the next 2 weeks

iarslankhalid:
sure @Benedikt-Wolf .. Good luck with your exams and job interviews

vinit5112:
@nirga I will look at it, if no one is working on this currently?

nirga:
Sure @vinit5112 :)

vinit5112:
@nirga, I understood the issue. But can you help me where we need to structure openai output.

nirga:
Hey @vinit5112 sorry for thte delay I was OOO!
We need to log the data returned by the structured output as OpenAI describes in this link:
https://openai.com/index/introducing-structured-outputs-in-the-api/

For example, they return something like `refusal` which we don't log on the span currently

nirga:
Thanks @AnsariZayd! Not sure where @vinit5112 stands with this

AnsariZayd:
Hi Nir, Thanks for giving me this opportunity. I am a little confused about
what exactly the new feature is intended to do, it would be great if you
can provide me some details related to the feature. Also I was looking at
the code in the packages/opentelemetry-instrumentation-openai, I am not
sure where do you want to add the new feature. Do you want to add a new
folder or file for this feature and wanted to store the refusal in a new
log ? Please let me know so that I can start working on it.
regards,
Zayd

nirga:
[...]

PR #2111, opened 2024-10-10
fix(openai): add structured output instrumentation

9dogs:

Fixes https://github.com/traceloop/openllmetry/issues/1815. I'm not sure if the [response_format](https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/#openai-attributes) span attribute should be added in this PR as well.

Also, the metrics tests are a bit fragile since each executed test adds its metrics to the `reader.get_metrics_data()` result, so figuring out what was added is a bit tricky.

### DIFF
--- a/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/__init__.py
+++ b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/__init__.py
@@ -198,6 +198,32 @@ class OpenAIV1Instrumentor(BaseInstrumentor):
                 assistants_create_wrapper(tracer),
             )
             wrap_function_wrapper(
+                "openai.resources.beta.chat.completions",
+                "Completions.parse",
+                chat_wrapper(
+                    tracer,
+                    tokens_histogram,
+                    chat_choice_counter,
+                    duration_histogram,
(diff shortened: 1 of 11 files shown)


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/__init__.py
```diff
@@ -197,6 +197,32 @@ def _instrument(self, **kwargs):
                 "Assistants.create",
                 assistants_create_wrapper(tracer),
             )
+            wrap_function_wrapper(
```
nirga, line 200: wondering why you added these?

nirga (commented):
Thanks @9dogs! We add the response format here or in a separate PR - up to you (if you do that - we might as well add everything from the recently published semantic conventions)

9dogs:
> Thanks @9dogs! We add the response format here or in a separate PR - up to you (if you do that - we might as well add everything from the recently published semantic conventions)

I cannot say I fully understand what's going on with semconvs, so let's leave the scope of this PR as it is.

nirga (commented):
Thanks @9dogs! Left a small comment

nirga (approved):
Thanks @9dogs!

PR #2150, opened 2024-10-16
fix(anthropic): add instrumentation for Anthropic tool calling (alternative to #1372)

dinmukhamedm:
This commit addresses [#879] by implementing instrumentation of Anthropic `tools` request args and `tool_use` response arguments.

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -152,7 +152,7 @@ async def _aset_input_attributes(span, kwargs):
                 set_span_attribute(
                     span, f"{SpanAttributes.LLM_PROMPTS}.{i}.role", message.get("role")
                 )
-        
+
         if kwargs.get("tools") is not None:
             for i, tool in enumerate(kwargs.get("tools")):
                 prefix = f"{SpanAttributes.LLM_REQUEST_FUNCTIONS}.{i}"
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -152,21 +152,56 @@ async def _aset_input_attributes(span, kwargs):
...
-                content.text,
-            )
+        tool_call_index = 0
+        text = ""
+        for content in response.get("content"):
+            content_block_type = content.type
+            # usually, Antrhopic responds with just one text block,
+            # but the API allows for multiple text blocks, so concatenate them
+            if content_block_type == "text":
+                text += content.text
```
dinmukhamedm, line 183: @nirga looking for your input on this. I tried to stay as close to OpenAI as possible. For OpenAI the `index`, is the ordinal number of the output in the `choices` array. I believe, there is no such thing in Anthropic, so `index` is always 0.

In theory, Anthropic could respond with something like `[text, tool_use, text, text, tool_use]`. I've never seen it do this in practice, but just in case, instead of hard-coding, I decided to concatenate everything in the text block.

An alternative to my approach would be to set these `text` blocks as `gen_ai.completions.i.content`, but I don't think this is consistent with the `choices` in OpenAI.

nirga: Makes sense @dinmukhamedm! Long term we plan to move these to otel events so there will be a "cleaner" solution

nirga (approved):

dinmukhamedm:
Sorry about the back and forth, had some trouble running `nx lint` on my machine. Perhaps a good sign to add an instruction note to the contribution guideline? :)

Issue #1983, opened 2024-09-16
🐛 Bug Report: Trace-ID Injection Missing Between LangChain and vLLM Spans


ronensc:
### Which component is this bug for?

Langchain Instrumentation

### 📜 Description

I am using the OpenLLMetry `LangchainInstrumentor` to instrument `VLLMOpenAI` in LangChain. Both vLLM and the `LangchainInstrumentor` are configured to export traces to the same trace collector.
However, the spans from vLLM and LangChain appear in separate traces. It seems that `LangchainInstrumentor` is not injecting its `trace-id` into the requests sent to vLLM. Is there a way to enable this?

### 👟 Reproduction steps

1. Run Jaeger as the trace collector
```
docker run --rm --name jaeger \
    -e COLLECTOR_ZIPKIN_HOST_PORT=:9411 \
    -p 6831:6831/udp \
    -p 6832:6832/udp \
    -p 5778:5778 \
    -p 16686:16686 \
    -p 4317:4317 \
    -p 4318:4318 \
    -p 14250:14250 \
    -p 14268:14268 \
    -p 14269:14269 \
    -p 9411:9411 \
    jaegertracing/all-in-one:1.57
```

2. Run vLLM
```
export OTEL_SERVICE_NAME="vllm-server"
export OTEL_EXPORTER_OTLP_TRACES_INSECURE=true
vllm serve facebook/opt-125m --otlp-traces-endpoint "grpc://localhost:4317"
```

3. Run LangChain with `LangchainInstrumentor`:
```
from opentelemetry.instrumentation.langchain import LangchainInstrumentor

from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.resources import Resource
from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
from opentelemetry.trace import set_tracer_provider

from langchain_community.llms import VLLMOpenAI

trace_provider = TracerProvider(resource=Resource(attributes={"service.name": "LangChain"}))
trace_provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter(endpoint="grpc://localhost:4317", insecure=True)))
trace_provider.add_span_processor(BatchSpanProcessor(ConsoleSpanExporter()))
set_tracer_provider(trace_provider)

LangchainInstrumentor().instrument()


llm = VLLMOpenAI(
[...]

PR #2152, opened 2024-10-16
fix(langchain): Add trace context to client requests

ronensc:
This PR addresses part of issue #1983 by wrapping the `_generate()`, `_agenerate()`, `_stream()`, and `_astream()` methods in the OpenAI classes to propagate the trace context through the request's HTTP headers.

However, the current implementation does not fully support streaming for `ChatModels`. The logic there differs slightly, and I was unable to access the `run_id` within `BaseChatOpenAI._stream()`. The `run_id` is essential for matching the correct span, and in other methods, I retrieve it via the `run_manager`. While `BaseChatOpenAI._stream()` accepts a `run_manager` parameter, it is not being passed in this case.

For reference:
- [BaseChatOpenAI._stream() signature](https://github.com/langchain-ai/langchain/blob/fc14f675f1af1e8fee23710f208259da9fcfd778/libs/partners/openai/langchain_openai/chat_models/base.py#L612)
- [Related section in chat_models.py](https://github.com/langchain-ai/langchain/blob/fc14f675f1af1e8fee23710f208259da9fcfd778/libs/core/langchain_core/language_models/chat_models.py#L385-L400)

The code that wraps the streaming methods `_stream()` and `_astream()` of the `ChatModels` is currently commented out, and the relevant tests are marked with `xfail()` to indicate their expected failure.


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -63,3 +163,40 @@ def __call__(
...
+        if run_manager:
+            run_id = run_manager.run_id
+            span_holder = self._callback_manager.spans[run_id]
+
+            extra_headers = kwargs.get("extra_headers", {})
+
+            # Inject tracing context into the extra headers
+            ctx = set_span_in_context(span_holder.span)
+            TraceContextTextMapPropagator().inject(extra_headers, context=ctx)
+            # TraceContextTextMapPropagator().inject(extra_headers, context=span_holder.context)
```
nirga, line 197: nit: remove?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -36,14 +41,109 @@ def _instrument(self, **kwargs):
...
+
+        if not self.disable_trace_context_propagation:
+            self._wrap_openai_functions_for_tracing(traceloopCallbackHandler)
+
+    def _wrap_openai_functions_for_tracing(self, traceloopCallbackHandler):
+        openai_tracing_wrapper = _OpenAITracingWrapper(traceloopCallbackHandler)
+
+        # Wrap langchain_community.llms.openai.BaseOpenAI
+        wrap_function_wrapper(
+            module="langchain_community.llms.openai",
```
nirga, line 59: Why not wrap the base class `BaseLLM`? https://github.com/langchain-ai/langchain/blob/316541536940a0e063ed09f48d2bec05ded862c3/libs/core/langchain_core/language_models/llms.py#L292
Then theoretically, it should work for all models

nirga (commented):
Thanks so much @ronensc! Really liked the through testing!

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
ronensc: I agree that wrapping the base class would be preferable. However, I couldn't find a way to access the `run_id`. For example, when I set a breakpoint in the `invoke()` method while debugging `test_trace_propagation[OpenAI]`, I found that `config["run_id"]` is [empty](https://github.com/langchain-ai/langchain/blob/316541536940a0e063ed09f48d2bec05ded862c3/libs/core/langchain_core/language_models/llms.py#L397). The `run_id` is determined further down the stack trace.

nirga: Oh well
It's a good start though

nirga (approved):

PR #2164, opened 2024-10-17
fix(sdk): add header for logging exporter

jinsongo:

The headers could be needed for LogExporter to connect to some logs receivers.
If no TRACELOOP_LOGGING_HEADERS defined, the TRACELOOP_HEADERS will be used.


nirga (approved):

jinsongo:
Just share a little more information, without the fix the user still can use  `OTEL_EXPORTER_OTLP_LOGS_HEADERS` to set the header because that's supported by `OTLPLogExporter` itself.
For example, `export OTEL_EXPORTER_OTLP_LOGS_HEADERS="key=abc,host=demos-vm"`

PR #2165, opened 2024-10-17
[Bugfix]: Replace the hardcoded Watsonx API URL value

### LINKED ISSUE #2163: 🐛 Bug Report: Watsonx hardcoded values

Watsonx Instrumentation

### 📜 Description

Some variables, showing in traceloop traces, are hardcoded.
For example, even if hitting watsonx in frankfurt, it shows that it's using https://us-south.ml.cloud.ibm.com, which is incorrect.

### 👟 Reproduction steps

Create a python script using watsonx in frankfurt for example, and see that in the traces it shows https://us-south.ml.cloud.ibm.com, as hardcoded in packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py line 103

### 👍 Expected behavior

It should show the real parameter istead of harcoded ones.

image

### 🤖 Python Version

3.12

_No response_

None

rajmanna-dev:
### What does this PR do?

Previously the watsonx api url was hardcoded in `_set_span_attribute`. To support the region based API URLs or on-prem, update the hardcoded API URL with a `env` variable to get the base URL by the customer.

Fixed #2163 

### Type of change
Update the hardcoded value with `env` var.

### DIFF
--- a/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
+++ b/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
@@ -104,7 +104,7 @@ def _set_api_attributes(span):
     _set_span_attribute(
         span,
         WatsonxSpanAttributes.WATSONX_API_BASE,
-        "https://us-south.ml.cloud.ibm.com",
+        os.getenv("API_BASE_URL"),
     )
     _set_span_attribute(span, WatsonxSpanAttributes.WATSONX_API_TYPE, "watsonx.ai")
     _set_span_attribute(span, WatsonxSpanAttributes.WATSONX_API_VERSION, "1.0")
--- a/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
+++ b/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
@@ -14,6 +14,7 @@ from opentelemetry.trace import get_tracer, SpanKind
 from opentelemetry.trace.status import Status, StatusCode
 from opentelemetry.metrics import get_meter
 from opentelemetry.metrics import Counter, Histogram
+from watsonx-langchain import watsonx_llm_init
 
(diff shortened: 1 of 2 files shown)


nirga (commented):
Thanks @rajmanna-dev - but this is not a good solution as it assumes the URL will always be in this environment variable and this may not be the case. A better solution is to try and extract it [from the WatsonxLLM object](https://github.com/traceloop/openllmetry/blob/0ba34ba9df7a015e3c3eb1f3c18fb30020e71104/packages/sample-app/sample_app/watsonx-langchain.py#L31). We've done something similar in the [vertexAI instrumentation](https://github.com/traceloop/openllmetry/blob/0ba34ba9df7a015e3c3eb1f3c18fb30020e71104/packages/opentelemetry-instrumentation-vertexai/opentelemetry/instrumentation/vertexai/__init__.py#L260).

rajmanna-dev:
I am trying to extract the url from WatsonxLLM object by getattr(instance, "url"). Should i set None if the url not found?

nirga:
@rajmanna-dev did you check that it's there? If not I'd default to the current value we have in the code

rajmanna-dev:
Is this modification works:
I set the `url` with `os.getenv("API_BASE_URL")` and then checked from `watsonx/__init__.py`.
```python
def _set_api_attributes(span, instance):
    if hasattr(instance, "url"):
        api_base_url = instance.url
    else:
        api_base_url = "https://us-south.ml.cloud.ibm.com"
    _set_span_attribute(
        span,
        WatsonxSpanAttributes.WATSONX_API_BASE,
        api_base_url
    )
    _set_span_attribute(span, WatsonxSpanAttributes.WATSONX_API_TYPE, "watsonx.ai")
    _set_span_attribute(span, WatsonxSpanAttributes.WATSONX_API_VERSION, "1.0")

    return
```

nirga:
@rajmanna-dev I mean - does this attribute exist? (for example when you run the sample app)

rajmanna-dev:
@nirga Yes "url" is exists inside the WatsonxLLM object.

nirga:
SG then @rajmanna-dev! Waiting for the updated PR :)

PR #2172, opened 2024-10-18
fix(watsonx): use with statement to automatically handle exception

jinsongo:

Use `with` statement to catch the exception and marks the span as being in an error state. then if an exception occurs, the span is properly ended and contains information about the exception.

### DIFF
--- a/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
+++ b/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
@@ -445,67 +445,66 @@ def _wrap(
 
     name = to_wrap.get("span_name")
 
-    span = tracer.start_span(
+    with tracer.start_span(
         name,
         kind=SpanKind.CLIENT,
         attributes={
             SpanAttributes.LLM_SYSTEM: "Watsonx",
             SpanAttributes.LLM_REQUEST_TYPE: LLMRequestTypeValues.COMPLETION.value,
         },
-    )
+    ) as span:
 
-    _set_api_attributes(span)
-    if "generate" in name:
-        _set_input_attributes(span, instance, kwargs)
-        if to_wrap.get("method") == "generate_text_stream":
-            if (raw_flag := kwargs.get("raw_response", None)) is None:
-                kwargs = {**kwargs, "raw_response": True}
-            elif raw_flag is False:
-                kwargs["raw_response"] = True
+        _set_api_attributes(span)
+        if "generate" in name:
+            _set_input_attributes(span, instance, kwargs)
+            if to_wrap.get("method") == "generate_text_stream":
+                if (raw_flag := kwargs.get("raw_response", None)) is None:
+                    kwargs = {**kwargs, "raw_response": True}
+                elif raw_flag is False:
+                    kwargs["raw_response"] = True
 
-    try:
-        start_time = time.time()
-        response = wrapped(*args, **kwargs)
-        end_time = time.time()
-    except Exception as e:
-        end_time = time.time()
-        duration = end_time - start_time if "start_time" in locals() else 0
+        try:
+            start_time = time.time()
+            response = wrapped(*args, **kwargs)
+            end_time = time.time()
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


jinsongo:
The fix is to resolve the problem of no trace for the erroneous call that's caused by incorrect watsonx API key. cc @gyliu513

packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
```diff
@@ -445,67 +445,66 @@ def _wrap(
...
-            exception_counter.add(1, attributes=attributes)
+            if duration > 0 and duration_histogram:
+                duration_histogram.record(duration, attributes=attributes)
+            if exception_counter:
+                exception_counter.add(1, attributes=attributes)
 
-        raise e
+            raise e
 
-    if "generate" in name:
```
nirga, line 485: I think this won't work well with the generator response. You need to manually close the span on all cases

jinsongo: Thanks, I will investigate and update with more details.

jinsongo: Explicitly call span.__exit__ for exception, just like using `with` statement to do it automatically.

jinsongo: @nirga please review again. `span.__exit__` is the default operation to record exception information and call span.end(), and no impact to generator response.

jinsongo: image

jinsongo: But looks like it's not ideal because of the parent span for workflow is missing. Do you have any suggestion?

nirga: @jinsongo why not just call `span.end()`?

jinsongo: @nirga __eixt__ will call span.end() and also handle the exception.

jinsongo: BTW, bedrock instrumentation use `with` statement and no 'Generator' problem, so the exception can be handled by `span.__exit__` automatically when exception raised from `line 166 response = fn(*args, **kwargs)`
https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py#L163-L171

jinsongo: `span.__exit__` implementation:
https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L986-L1008
`span.__exit__` of parent class:
https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L186-L194

nirga: Yes @jinsongo but you're setting the exception yourself anyway so you don't really need this, right?

jinsongo: I prefer to set the exception information into the span by simply calling `__exit__`.
As you know, without the fix, I cannot see any span sent after the exception raised.

nirga: @jinsongo why? You didn't call `__enter__` so I don't think it's preferred on explicitly logging the exception on the span

PR #1858, opened 2024-08-16
Feature: Support Prompt Caching

kowshik24:
This PR  supports tracking and logging prompt caching(#1838) metrics in the OpenTelemetry Anthropic instrumentation. 

Specifically, it captures and logs the `cache_creation_input_tokens` and `cache_read_input_tokens` fields from the API response. This enhancement allows users to monitor the effectiveness of their prompt caching strategy, providing insights into cache utilization and efficiency.

Changes include:

* Updated `_set_token_usage` and `_aset_token_usage` functions to log caching metrics.
* Added new span attributes for `cache_creation_input_tokens` and `cache_read_input_tokens`.
* Updated `streaming.py` to support prompt caching metrics in streaming responses.

Issue #1838, opened 2024-08-14
🚀 Feature: Support Prompt Caching


nirga:
### Which component is this feature for?

Anthropic Instrumentation

### 🔖 Feature description

Support logging of prompt caching, including cached token usage:

https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#how-can-i-track-the-effectiveness-of-my-caching-strategy

### 🎤 Why is this feature needed ?

-

### ✌️ How do you aim to achieve this?

-

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

samsmithspace:
Hi @nirga, can I work on this task?

nirga:
Yes! Thank you :) @samsmithspace

kowshik24:
Hi @samsmithspace  are you still working on it?

kowshik24:
Hi @nirga I have updated the prompt caching. Could you please check?

Karush2807:
is it still open for contribution?

nirga:
@dinmukhamedm this is fixed with #2175 right?

PR #2175, opened 2024-10-20
fix(anthropic): add instrumentation for Anthropic prompt caching

dinmukhamedm:
This PR addresses #1838 and is an alternative to the stale (and in the current form not working) #1858.

This is not a draft, it is a final working implementation. However, there are a couple open questions, that I will leave as comments to the relevant pieces of code.

Below is a screenshot of the new resulting attributes flattened to a YAML format:

anthropic_token_count_attributes


packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -220,27 +251,23 @@ async def _aset_token_usage(
...
+        cache_creation_tokens = dict(usage).get("cache_creation_input_tokens", 0)
+    else:
+        cache_creation_tokens = 0
+
+    input_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens
+
+    if token_histogram and type(input_tokens) is int and input_tokens >= 0:
         token_histogram.record(
-            prompt_tokens,
+            input_tokens,
```
dinmukhamedm, line 270: I am open to discussing whether this must be input tokens (i.e. all tokens sent by the user) or prompt tokens (i.e. the new tokens that anthropic has neither written to nor read from cache)

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -308,27 +345,23 @@ def _set_token_usage(
...
```
dinmukhamedm, line 364: ditto

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -687,7 +729,7 @@ def __init__(
         get_common_metrics_attributes: Callable[[], dict] = lambda: {},
         upload_base64_image: Optional[
             Callable[[str, str, str, str], Coroutine[None, None, str]]
-        ] = lambda *args: "",
+        ] = None,
```
dinmukhamedm, line 732: This is not exactly in the scope of this PR, but if `opentelemetry-instrumentation-anthropic` is used with another tracer/exporter than `openllmetry/traceloop_sdk`, and that initialization does not pass the `upload_base_64_image` then the entire instrumentation fails. The reason is that this function is `async`, whereas the default value, `lambda: *args: ""` is not, so awaiting an `str` fails.

Hope this change is small enough to piggyback :)

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -135,30 +157,26 @@ def build_from_streaming_response(
...
+            else:
+                completion_content = ""
+                if complete_response.get("events"):
+                    model_name = complete_response.get("model") or None
+                    for event in complete_response.get("events"):  # type: dict
+                        if event.get("text"):
+                            completion_content += event.get("text")
+
+                    if model_name:
+                        completion_tokens = instance.count_tokens(completion_content)
```
dinmukhamedm, line 179: In both these places (same in async below), I default to getting the input and output tokens from the response, because I think it is:
- a better source of truth
- more efficient

This also caused me to change some values and fixtures in test, where the estimate from `count_tokens` was way off – 17 instead of 8.

Let me know, if this was a design decision to default to counting tokens, and if I have missed something

packages/opentelemetry-instrumentation-anthropic/tests/data/1024+tokens.txt
```diff
@@ -0,0 +1,90 @@
+Open-Source Library OpenLLMetry Brings LLM Functionality to OpenTelemetry
```
dinmukhamedm, line 1: This is totally AI generated, so sorry if it is not exactly true haha

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -74,6 +74,11 @@ class SpanAttributes:
     LLM_OPENAI_API_VERSION = "gen_ai.openai.api_version"
     LLM_OPENAI_API_TYPE = "gen_ai.openai.api_type"
 
+    # Anthropic
```
dinmukhamedm, line 77: I am afraid that this PR won't pass tests, because `pyproject.toml` in the `anthropic` package looks at `0.4.1`, basically the PyPI version, but my changes rely on the new attributes here. Let me know if I need to merge this first somehow

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -74,6 +74,11 @@ class SpanAttributes:
...
+    LLM_ANTHROPIC_CACHE_CREATION_INPUT_TOKENS = "llm.anthropic.usage.cache_creation_input_tokens"
+    LLM_ANTHROPIC_CACHE_READ_INPUT_TOKENS = "llm.anthropic.usage.cache_read_input_tokens"
+    LLM_ANTHROPIC_TOTAL_INPUT_TOKENS = "llm.anthropic.usage.total_input_tokens"
```
dinmukhamedm, line 80: The logic I followed when naming this attributes was the following:
1. https://github.com/traceloop/openllmetry/pull/1858#discussion_r1720759231 this comment
2. `llm` is apparently less standardized than `gen_ai`, so we are free to experiment here
3. I haven't found a separate documentation for `gen_ai.anthropic` attributes like they have for `gen_ai.openai` [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/openai.md)

So I thought, a good in between would be `llm.anthropic.usage.*`, but I am open to additional thoughts.

Also, I am not sure if restricting this to anthropic is a good idea, because both google-generativeai and openai have at least some overlap with this.

dinmukhamedm:
Also, I was surprised to learn that system messages are not instrumented in Anthropic, so I opened https://github.com/traceloop/openllmetry/issues/2187. Also willing to contribute, unless anybody else picks that up faster

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
nirga: Nope, that's actually better! Anthropic used to not return the token count for streaming requests

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
nirga: Perfect :)

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -308,27 +345,23 @@ def _set_token_usage(
...
+        cache_read_tokens = dict(usage).get("cache_read_input_tokens", 0)
+    else:
+        cache_read_tokens = 0
+
+    if usage := response.get("usage"):
+        cache_creation_tokens = dict(usage).get("cache_creation_input_tokens", 0)
+    else:
+        cache_creation_tokens = 0
+
+    input_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens
```
nirga, line 360: Aren't you double-calculating the number of input tokens like this? From my understanding `cache_read_tokens+cache_creation_tokens` should be exactly the number of tokens in the input. Or is it the case that either `prompt_tokens` is set (for non-cached requests) OR `cache_read_tokens+cache_creation_tokens`?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
nirga: Yeah I need to manually publish the semconv package cause poetry doesn't support deeply nested local dependencies :/ 
https://github.com/python-poetry/poetry/issues/2270

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
nirga: I'd argue these should be `gen_ai` and without `anthropic`. Most of the "official" semantic conventions came after OpenLLMetry and were inspired but the work we've done here - so we should set the preferred path moving forward!

nirga (commented):
Excellent work @dinmukhamedm! Left one comment, and I'll manually publish the semconv package so the CI will pass here.

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
dinmukhamedm: Nope, Anthropic fills all three in. If it is a cache write, `cache_read_tokens == 0`. If it as a cache read, `cache_creation_tokens == 0`. `prompt_tokens` are tokens from uncached parts of the messages. For cache writes and reads is always about 3-4. I am assuming this is some control tokens or stop sequences or something. You can see some numbers I hard-coded in the tests [here](https://github.com/traceloop/openllmetry/blob/0cabaae2c86ff4d1a5957b04adbcb9108c03ddb4/packages/opentelemetry-instrumentation-anthropic/tests/test_completion.py#L624-L644).

For example, if I send two text blocks of sizes 1200 and 100 tokens in one message and only direct Anthropic to cache the first one, the usage will be:
- `{"cache_read_input_tokens": 0, "cache_creation_input_tokens": 1200, "input_tokens": 104, "output_tokens": ...}` for the first call, and
- `{"cache_read_input_tokens": 1200, "cache_creation_input_tokens": 0, "input_tokens": 104, "output_tokens": ...}` for the second

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
dinmukhamedm: I agree generally, but given the context I provide in https://github.com/traceloop/openllmetry/pull/2175#discussion_r1809080693 and the fact that what used to be called `gen_ai.usage.prompt_tokens` is now apparently called `gen_ai.usage.input_tokens` ["officially"](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md), we need to decide what to call these, especially the `TOTAL_INPUT_TOKENS` one.

So, for context, OpenAI does not return `cache_creation_input_tokens` as they don't charge for the operation. Their usage block looks like this:
```
usage: {
  total_tokens: 2306,
  prompt_tokens: 2006,
  completion_tokens: 300,
  
  prompt_tokens_details: {
    cached_tokens: 1920,
    audio_tokens: 0,
  },
  completion_tokens_details: {
    reasoning_tokens: 0,
    audio_tokens: 0,
  }
}
```

Gemini returns the number of `cache_creation_input_tokens` from the `cache.create` call, and they charge for storage of cache tokens per hour.

So I think, overall the formula is this:
```
input_tokens = cached_tokens + uncached_tokens
cached_tokens = cache_read_tokens # generally
cached_tokens = cache_read_tokens OR cache_creation_tokens # for Anthropic
```
and `cache_read_tokens` and `cache_creation_tokens` are charged for at vastly different prices.

With this in mind, the main open question is what are `gen_ai.usage.prompt_tokens` and `gen_ai.usage.input_tokens` bound to mean? Is it total input tokens or is it the number of uncached tokens charged for regular price? While the former makes more sense to me, I think this will sort of break any cost calculation implementations depending on this library?

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
nirga: My intuition is that we should keep the `input_tokens` constant across providers - so it should always be the number of tokens in the input - regardless if some of them were cached and some weren't

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
dinmukhamedm: Done! @nirga please check my latest commit

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
dinmukhamedm: Yes, except for one place where I accidentally added just uncached tokens. Fixed in the last commit now.

nirga:
@dinmukhamedm can you give me permissons to push to your PR?

dinmukhamedm:
@nirga invited you with write permissions on our fork

PR #2190, opened 2024-10-21
Update __init__.py

smit23patel:
1. Type Hints: Added type hints to function signatures for better clarity and type checking.

2. Optional Parameters: Used Optional for parameters that can be None, improving the understanding of the function's expected input.


nirga (commented):
Thanks @smit23patel, but if you're adding type hints - let's add the right ones! Can you replace the `Any` with the right classes from the appropriate libraries?

PR #2191, opened 2024-10-21
incomplete example to fix  #591

### LINKED ISSUE #591: 🚀 Feature: Make instrumentations compatible with "opentelemetry-instrument" automatic instrumentation

kguelzau:
- use optional dependencies, instead of test dependencies
- https://python-poetry.org/docs/pyproject/#extras


nirga:
@kguelzau attaching here the comment from the issue
https://github.com/traceloop/openllmetry/issues/591#issuecomment-2427374153

I'm not sure this is the right way to do that - I think this will force the instrumentation to pull an OpenAI dependency, no?

kguelzau:
I still think it's the right way to do, but we switched away from otel-operator and are using the dependencies directly in the app.
See [here](https://github.com/open-telemetry/opentelemetry-operator/issues/1885#issuecomment-2482593720) why.
So I am closing this draft PR.

PR #2192, opened 2024-10-21
fix(bedrock): Correctly set index on anthropic content attribute

RileyEv:


RileyEv:
There is some better logic for reading the message and converting to span attributes in the `opentelemetry-instrumentation-anthropic` library. Is it okay to add a dependency, or should I just pull this logic out and add to this library too?

nirga (commented):
Nice catch @RileyEv! Mind signing the CLA?

We can pull that dependency, or better yet - extract it to the ai semconv common library

PR #2214, opened 2024-10-27
fix(langchain): structured output response parsing

galkleinman:


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/utils.py
```diff
@@ -13,11 +13,16 @@ def default(self, o):
...
                 del o["callbacks"]
                 return o
+            
         if dataclasses.is_dataclass(o):
             return dataclasses.asdict(o)
 
         if hasattr(o, "to_json"):
             return o.to_json()
+        
+        # check if o is a pydantic model
```
nirga, line 23: pydantic should be handled with the `to_json`, no?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -585,7 +593,7 @@ def on_llm_end(
                 )
 
         _set_chat_response(span, response)
-        span.end()
+        self._end_span(span, run_id)
```
nirga, line 596: you're right - but it won't change anything since an LLM span is a leaf span anyway

nirga (approved):
Cool, though I'm not sure how it fixes structured outputs - it seems like you're fixing tool calling and not structured outputs - unless I'm missing something?

Also, I'd add tests for this

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/utils.py
galkleinman: depends on the version. in pydantic v1 (which is apparently installed in my env) it won't.. they method was throwing an exception which was caught by the `@dont_throw` decorator, took me a while to find it.

Issue #2195, opened 2024-10-22
🐛 Bug Report: Adding asyncio.run() in 0.31.3 makes the package fail when run from within async


jp-agenta:
### Which component is this bug for?

All Packages

### 📜 Description

Calling `openai` endpoints from within an `async` function works in `0.31.2` but fails starting in `0.31.3` since `asyncio.run()` was added.

### 👟 Reproduction steps

1. add `openai` instrumentation
2. add a call to an `openai` endpoint inside an `async` function

```
async def foo(input):
    
    ...
    result = openai.embeddings.create(input=input, model="text-embedding-ada-002")
    ...

    return result
```

### 👍 Expected behavior

3.a. test in `0.31.2` -> success
3.b. test in `0.31.3` -> success

### 👎 Actual Behavior with Screenshots

3.a. test in `0.31.2` -> success
3.b. test in `0.31.3` -> failure

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

nirga:
Thanks for reporting @jp-agenta! Can you verify this bug exists with the latest version? (v0.33.2)

jp-agenta:
Hi @nirga, thanks for your quick response.
I will try again with `0.33.2` as soon as practicable and I will get back to you soon !

mmabrouk:
> Thanks for reporting @jp-agenta! Can you verify this bug exists with the latest version? (v0.33.2)

Hi @nirga , I confirm that the bug still exists in v0.33.2

```
  File "...opentelemetry/instrumentation/openai/utils.py", line 78, in wrapper
    return func(
  File "...opentelemetry/instrumentation/openai/shared/chat_wrappers.py", line 81, in chat_wrapper
    asyncio.run(_handle_request(span, kwargs, instance))
  File ".../asyncio/runners.py", line 33, in run
    raise RuntimeError(
RuntimeError: asyncio.run() cannot be called from a running event loop
sys:1: RuntimeWarning: coroutine '_handle_request' was never awaited
```

nirga:
@mmabrouk thanks! do you happen to have a small demo that reproduces this?

mmabrouk:
Yes of course:

```
from opentelemetry.instrumentation.openai import OpenAIInstrumentor
from openai import OpenAI
import asyncio
openai = OpenAI()


OpenAIInstrumentor().instrument()


[...]

PR #2226, opened 2024-10-29
fix(openai+anthropic): async call crashing the app when already in a running asyncio loop

galkleinman:
fixes: https://github.com/traceloop/openllmetry/issues/2195


nirga (approved):

jp-agenta:
Thank you !

PR #2304, opened 2024-11-16
feat(instrumentation): Add instrument support for HTTPX

### LINKED ISSUE #2283: 🚀 Feature: Add instruments support for httpx

All Packages

### 🔖 Feature description

Today the solution currently supports Instruments.REQUESTS and Instruments.URLLIB3.  Is there any plans to extend the instruments to support HTTPX?

Many libraries that require async support have migrated from requests to httpx.  Providing httpx support natively would extend the use of this library to a larger user base.

Ideally, this feature would provide similar ease of configuration as is currently available for Requests.

### 🔄️ Additional Information

_No response_

None

whoiskrtk2:
Fixes #2283 

- [X] (If applicable) I have updated the documentation accordingly.

TODOs/Questions:
- `If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change.` Is this needed for this PR?
- HTTPX instrumentor from OpenTelemetry does not support excluded URLs at the moment. Do you think we should open an issue there and push a fix?


nirga:
Hey @whoiskrtk2, thanks! I was waiting for you to mark it as ready for review :) 
Given the fact there isn't a way to exclude URLs - I'd at least test and see what spans we get for some of our test apps and put those screenshots here. And yes, we should open an issue there I think.

whoiskrtk2:
Hi @nirga 
Sure, I'll open an issue with OTel, and also add some screenshots from this.
Apologies for the delay.

PR #2323, opened 2024-11-21
fix(openai): Add token count and system to assistants

### LINKED ISSUE #2318: 🐛 Bug Report: OpenAI Assistant API instrumentation doesn't use token stats from response

OpenAI Instrumentation

### 📜 Description

OpenAI Assistant API instrumentation does not use the token usage that one could get from the `run.usage` object. In addition, `gen_ai.system=openai` is not sent

### 👟 Reproduction steps

1. Use [sample app](https://github.com/traceloop/openllmetry/blob/main/packages/sample-app/sample_app/openai_assistant.py) to call an Assistant.
2. [Optional] add `print(run.usage)` after line 34, when `run.status == "completed"`.

### 👍 Expected behavior

- The resulting span must contain `gen_ai.system = openai` in its attributes.
- Tokens must be added as `gen_ai.usage` attributes even when `enrich_tokens` flag is not set, because they are returned in the response.

- The resulting span does not have the system attribute
- There are no `gen_ai.usage.*_tokens` attributes on the resulting span

### 🤖 Python Version

3.9

(I haven't tried with the token enricher/in-place counter, but I think we should not rely on it, now that the API returns token usage)

dinmukhamedm:
Fixes #2318 by adding prompt and completion tokens to the assistant spans.

In addition:
- Add `gen_ai.system = openai` attribute.
- Fix the dict key lookup that caused the assistant.run spans to have end_time later than the actual value (sometimes caused the span to outlive the parent)
- Remove `time.sleep` from test because it has a VCR fixture, reducing openai tests time from 38s to 12s


nirga (approved):
Thanks @dinmukhamedm!

PR #2383, opened 2024-12-10
fix(sdk): Update JSONEncoder to allow class instance methods to be serializable

gladwindos:
I noticed that when using the task decorator on class instance methods, the `traceloop.entity.input` attribute was not showing up in spans. I did some digging and saw that a TypeError was being triggered when trying to [serialize the methods args](https://github.com/traceloop/openllmetry/blob/b7d917ffbbe856e0ccb56f271860bac8e462c161/packages/traceloop-sdk/traceloop/sdk/decorators/base.py#L68) because the first arg is the instance of the class. 

To fix this issue, I've made an update to the custom `JSONEncoder` to properly handle class instances. Specifically, if the encoder encounters an object that cannot be directly serialized, it now falls back to returning the class’s name as a string. 

### DIFF
--- a/packages/traceloop-sdk/tests/test_class_tasks.py
+++ b/packages/traceloop-sdk/tests/test_class_tasks.py
@@ -1,0 +1,58 @@
+import json
+
+from opentelemetry.semconv_ai import SpanAttributes
+from traceloop.sdk.decorators import task
+
+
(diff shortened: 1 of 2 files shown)


nirga (approved):
Nice catch @gladwindos! Thanks 🙏🏼

Issue #2419, opened 2024-12-19
🐛 Bug Report: Gemini (google generativeai) instrumentation sets the input message attributes incorrectly


dinmukhamedm:
### Which component is this bug for?

All Packages

### 📜 Description

The attribute for `gen_ai.prompt.N.*` is not set correctly. Yes, I know we will soon/eventually move to sending these as events, but for now, this is an error.

### 👟 Reproduction steps

```
import google.generativeai as genai
import os
from traceloop.sdk import Traceloop

Traceloop.init()

genai.configure(api_key=os.environ['GEMINI_API_KEY'])
model = genai.GenerativeModel('gemini-1.5-flash')

def gemini_response(user_message: str):
    response = model.generate_content(user_message)
    print(response.text)

gemini_response('The opposite of hot is')
```

### 👍 Expected behavior

The following prompt and completion attributes must be set (by analogy with openai and anthropic):
```
gen_ai.prompt.0.content: "The opposite of hot is\n"
gen_ai.prompt.0.role: "user"
gen_ai.completion.0.content: "cold\n"
gen_ai.completion.0.role: "assistant"
```

### 👎 Actual Behavior with Screenshots

Instead, these attributes are set:
```
gen_ai.completion.0.content: "cold\n"
gen_ai.prompt.0.user: "The opposite of hot is\n"
```

### 🤖 Python Version

3.12

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
You're right! Appreciate the willingness to fix this :)

PR #2511, opened 2025-01-13
fix(gemini): set prompt attributes for role and content

dinmukhamedm:
Addressing #2419 

- [ ] I have added tests that cover my changes. <-- Yes, but they are not reproducible without my gemini key


packages/opentelemetry-instrumentation-google-generativeai/tests/test_generate_content.py
```diff
@@ -0,0 +1,41 @@
+# import pytest
+# import os
+
+# import google.generativeai as genai
+# from opentelemetry.semconv_ai import SpanAttributes
+
+def test_gemini_generate_content(exporter):
+    # This test is working, but since Gemini uses gRPC,
+    # vcr does not record it, therefore we cannot test this without
+    # setting the API key in a shared secret store like GitHub secrets
```
dinmukhamedm, line 10: @nirga any suggestions on how to go about this?

nirga: unfortunately no :/ only to mock responses

nirga (approved):

Issue #2513, opened 2025-01-14
🐛 Bug Report: Whole message history is stringified into a single 'user prompt'


alexmojaki:
### Which component is this bug for?

VertexAI Instrumentation

### 📜 Description

Passing a list of messages to `generate_content` combines them all into a single string and stores/presents them as a single user prompt instead of keeping them separate.

### 👟 Reproduction steps

```python
import vertexai
from traceloop.sdk import Traceloop
from vertexai.generative_models import Content, GenerativeModel, Part

Traceloop.init()
vertexai.init()
model = GenerativeModel('gemini-1.5-flash-002')
response = model.generate_content(
    [
        Content(
            role='user',
            parts=[
                Part.from_text("What's 2+2?"),
            ],
        ),
        Content(
            role='assistant',
            parts=[
                Part.from_text('5'),
            ],
        ),
        Content(
            role='user',
            parts=[
                Part.from_text('really?'),
            ],
        ),
    ]
)
```

### 👍 Expected behavior

Attributes look something like:

```json
{
  "gen_ai.prompt.0.role": "user",
  "gen_ai.prompt.0.content": "What's 2+2?\n",
  "gen_ai.prompt.1.role": "assistant",
  "gen_ai.prompt.1.content": "5\n",
  "gen_ai.prompt.2.role": "user",
  "gen_ai.prompt.2.content": "really?\n",
  "gen_ai.completion.3.role": "assistant",
  "gen_ai.completion.3.content": "Oops!  My apologies. 2 + 2 = 4.  I'm still under development and learning to perform these calculations correctly.\n"
}
```

### 👎 Actual Behavior with Screenshots

Attributes look like:

```json
{
  "gen_ai.prompt.0.user": "role: \"user\"\nparts {\n  text: \"What\\'s 2+2?\"\n}\n\nrole: \"assistant\"\nparts {\n  text: \"5\"\n}\n\nrole: \"user\"\nparts {\n  text: \"really?\"\n}\n\n",
  "gen_ai.completion.0.role": "assistant",
  "gen_ai.completion.0.content": "Oops!  My apologies. 2 + 2 = 4.  I'm still under development and learning to perform these calculations correctly.\n"
}
```

<img width="703" alt="Screenshot 2025-01-14 at 16 50 21" src="https://github.com/user-attachments/assets/b89abd71-bc88-4d7d-b562-b2221c715ae1" />

### 📃 Provide any additional context for the Bug.

[...]

PR #2517, opened 2025-01-18
fix bug #2513(vertexai): improve input attribute handling for prompts and conte…

kowshik24:
Fix  Bug Report: Whole message history is stringified into a single 'user prompt' #2513

1. Bug Analysis
Current Behavior
Messages from chat history are combined into single string
All roles and content merged into one user prompt
Loses message separation and role information

2. Root Cause
_set_input_attributes() treating Content list as plain text
Not properly parsing individual Content objects
Missing logic to handle chat message format

3. Impact
Loss of conversation structure in telemetry
Incorrect attribution of message roles
Poor observability of chat interactions

4. Fix Required
Update _set_input_attributes() to handle Content objects
Parse role and parts from each message
Store each message separately with correct role
Maintain message order with index


nirga (commented):
Thanks @kowshik24! Can you provide a screenshot of how it looks like?

Issue #2535, opened 2025-01-20
🐛 Bug Report: No module named 'weaviate.schema'


danieldaeschle:
### Which component is this bug for?

Anthropic Instrumentation

### 📜 Description

```
File "...\model\api\api.py", line 54, in create_app
    WeaviateInstrumentor().instrument()
  File "...\venv\Lib\site-packages\opentelemetry\instrumentation\instrumentor.py", line 116, in instrument
    result = self._instrument(  # pylint: disable=assignment-from-no-return
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\opentelemetry\instrumentation\weaviate\__init__.py", line 231, in _instrument
    module = importlib.import_module(wrap_module)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dd\AppData\Local\Programs\Python\Python311\Lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'weaviate.schema'
```

### 👟 Reproduction steps

weaviate-client==4.10.4

### 👍 Expected behavior

Work

### 👎 Actual Behavior with Screenshots

Crashes

### 🤖 Python Version

3.10

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

danieldaeschle:
everything is up to date.

nirga:
Thanks @danieldaeschle - which weaviate version are you using?

danieldaeschle:
1.25.6 and 1.28.3
weaviate client 4.10.4

Tim-ty-tang:
Hi Running into the same issue and currently using a custom WRAPPED_METHOD as a patch. Any updates on this bug?

nirga:
[...]

PR #2538, opened 2025-01-21
 Fix ModuleNotFoundError for weaviate.schema

Charlesnorris509:
**### This pull request fixes the ModuleNotFoundError for weaviate.schema by updating the import statements and adding error handling to ensure the module is correctly imported.**

_Changes:
Updated packages/opentelemetry-instrumentation-weaviate/opentelemetry/instrumentation/weaviate/__init__.py to handle import errors and log them appropriately.

Issue: Closes #2535_

### DIFF
--- a/packages/opentelemetry-instrumentation-llamaindex/pyproject.toml
+++ b/packages/opentelemetry-instrumentation-llamaindex/pyproject.toml
@@ -8,7 +8,7 @@ show_missing = true
 
 [tool.poetry]
 name = "opentelemetry-instrumentation-llamaindex"
-version = "0.36.1"
+version = "0.38.0"
 description = "OpenTelemetry LlamaIndex instrumentation"
 authors = [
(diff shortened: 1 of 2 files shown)


nirga (commented):
Thanks @Charlesnorris509, can you sign the CLA?
The problem with what you're doing is that it's hiding the problem - not solving it. Ideally we need to figure out what's the module name in the latest weaviate versions and fix that

Charlesnorris509:
@nirga I have signed the CLA

dvirski:
Thanks for the PR! Closing this — the original ModuleNotFoundError crash reported in #2535 is no longer reachable on main (the import is now wrapped in try/except ModuleNotFoundError: continue, so v3-only modules are silently skipped on Weaviate v4). The reporter also confirmed they can't repro anymore. Appreciate the contribution!

PR #2577, opened 2025-01-27
Check package version at import for better performance. Fix #2547

### LINKED ISSUE #2547: 🐛 Bug Report: Performance issues with opentelemetry-instrumentation-openai

Anthropic Instrumentation

### 📜 Description

Hi! First, thank you for opentelemetry-instrumentation-openai. It makes my life easier :-)

But, it's computationally extensive, significantly slowing the service of my team. I'm raising a bug report as I couldn't find a better issue type from the ones suggested. Feel free to change this as needed.

Overall, the impact of this issue is reduced throughput. In our production environment removing the library increased our throughput under load 6 folds! The minimal reproducible project [here](https://github.com/Nagasaki45/OTEL-OpenAI-performance-investigation) shows more than 4x throughput increase when dropping the library, without any other change.

### 👟 Reproduction steps

This is a bit elaborated, so I created a [minimal reproducible example](https://github.com/Nagasaki45/OTEL-OpenAI-performance-investigation) for demonstrating the problem and for going through the investigation steps. The README of that repo explains the steps in details. If more info is needed I'm happy to provide it.

### 👍 Expected behavior

Adding opentelemetry-instrumentation-openai to a project should have a minimal impact on a system's performance.

Adding opentelemetry-instrumentation-openai to a project reduces the throughput under load by ~4-6 times.

### 🤖 Python Version

3.12.0

_No response_

Nagasaki45:

Sorry I didn't check all of these checkboxes. I'm not sure what would be an accurate way to do this. It's a performance improvement, so unit-tests are not going to capture the fix, isn't it? Also, it's not a fix to a single instrumentation library, so I'm not sure about the PR name as this change touches multiple instrumentation packages.

What I can say is that before this change I get these stats [my the test repo](https://github.com/Nagasaki45/OTEL-OpenAI-performance-investigation) (more info on #2547).
- Requests per second: 1.60
- Median time per request: 35562ms

After the change.
- Requests per second: 4.62
- Median time per request: 15417ms

(the stats are not the same as in the ticket because these are from a different machine)

I'm happy to change the code and this PR as you see fit. Let me know ;-)

### DIFF
--- a/packages/opentelemetry-instrumentation-groq/opentelemetry/instrumentation/groq/utils.py
+++ b/packages/opentelemetry-instrumentation-groq/opentelemetry/instrumentation/groq/utils.py
@@ -9,6 +9,8 @@ from opentelemetry.semconv_ai import SpanAttributes
 GEN_AI_SYSTEM = "gen_ai.system"
 GEN_AI_SYSTEM_GROQ = "groq"
 
+_PYDANTIC_VERSION = version("pydantic")
+
 
 def set_span_attribute(span, name, value):
     if value is not None and value != "":
@@ -68,7 +70,7 @@ def error_metrics_attributes(exception):
 
 
 def model_as_dict(model):
-    if version("pydantic") < "2.0.0":
+    if _PYDANTIC_VERSION < "2.0.0":
         return model.dict()
     if hasattr(model, "model_dump"):
         return model.model_dump()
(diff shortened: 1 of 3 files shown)


nirga (approved):
Thanks @Nagasaki45! Mind signing the CLA so we can merge this?

Nagasaki45:
> Thanks @Nagasaki45! Mind signing the CLA so we can merge this?

Just did. Thanks!

Nagasaki45:
Hi @nirga. Any update on this? If there's anything else you would like me to change here I'm happy to do so.

PR #2618, opened 2025-02-06
fix(sdk): improve package name detection with type hints and null safety

doronkopit5:

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/utils/package_check.py
+++ b/packages/traceloop-sdk/traceloop/sdk/utils/package_check.py
@@ -2,13 +2,10 @@ from importlib.metadata import Distribution, distributions
 
 
 def _get_package_name(dist: Distribution) -> str | None:
-    # Try both 'Name' and 'name' keys to handle different metadata formats
-    if hasattr(dist, 'metadata') and dist.metadata is not None:
-        for key in ('Name', 'name'):
-            try:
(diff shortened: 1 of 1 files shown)


nirga (approved):

packages/traceloop-sdk/traceloop/sdk/utils/package_check.py
```diff
@@ -1,20 +1,14 @@
...
-    # Try both 'Name' and 'name' keys to handle different metadata formats
-    if hasattr(dist, 'metadata') and dist.metadata is not None:
-        for key in ('Name', 'name'):
-            try:
-                return dist.metadata[key].lower()
-            except (KeyError, AttributeError):
-                continue
+def _get_package_name(dist: Distribution) -> str | None:
+    try:
+        return dist.name.lower()
```
galkleinman, line 6: shouldn't be:

```suggestion
        return dist.name().lower()
```

?

galkleinman (approved):

packages/traceloop-sdk/traceloop/sdk/utils/package_check.py
doronkopit5: nope, it's a property
```
@property
def name(self):
    """Return the 'Name' metadata for the distribution package."""
    return self.metadata['Name']
```

PR #2635, opened 2025-02-16
fix(sdk): async generator wrapping

doronkopit5:


packages/traceloop-sdk/tests/test_workflows.py
```diff
@@ -273,7 +273,7 @@ async def stream_numbers():
 
     results = []
 
-    async for num in await stream_numbers():
```
doronkopit5, line 276: this code is not python valid. previously our decorator made it which means it's not just decorating

nirga (approved):

PR #2743, opened 2025-03-06
fix(sdk): When tracing task with no `name` provided , use qualified name instaed of name

RoiTabach:
- [V] PR name follows conventional commits format: `feat(instrumentation): ...` or `fix(instrumentation): ...`.
- [V] (If applicable) I have updated the documentation accordingly.

----

> [!IMPORTANT]
> Use `fn.__qualname__` instead of `fn.__name__` in `entity_method` for unnamed tasks in `base.py`.
> 
>   - **Behavior**:
>     - In `entity_method` in `base.py`, use `fn.__qualname__` instead of `fn.__name__` when `name` is not provided.
>     - Affects tracing by using the qualified name for unnamed tasks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 368a40f1193d18ad44ec3fc3c3007e19dce9ec43. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/decorators/base.py
+++ b/packages/traceloop-sdk/traceloop/sdk/decorators/base.py
@@ -258,7 +258,7 @@ def entity_class(
     tlp_span_kind: Optional[TraceloopSpanKindValues] = TraceloopSpanKindValues.TASK,
 ):
     def decorator(cls):
-        task_name = name if name else camel_to_snake(cls.__name__)
+        task_name = name if name else camel_to_snake(cls.__qualname__)
         method = getattr(cls, method_name)
         setattr(
             cls,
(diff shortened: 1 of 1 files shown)


packages/traceloop-sdk/traceloop/sdk/decorators/base.py
doronkopit5, line 1: what about `decorate` under `entity_class`
shouldn't be the same

RoiTabach: agree, added. Also grep'd the repo for more `__name__` but found no more places that requires changing

doronkopit5 (approved):
thanks for the fix

Issue #2774, opened 2025-03-10
🚀 Feature: Add thinking tokens to the instrumentation


dinmukhamedm:
### Which component is this feature for?

Anthropic Instrumentation

### 🔖 Feature description

We should track the content of thinking tokens for thinking models that return those, such as `claude-sonnet-3-7-thinking`

### 🎤 Why is this feature needed ?

It is extremely helpful in debugging prompts, especially for larger and more complex tasks

### ✌️ How do you aim to achieve this?

Anthropic returns its "thoughts" in a separate field. The only unclear thing is how to name the attributes.

I propose storing the thought as a separate role, i.e. 

- `gen_ai.completions.0.role = "thinking"`
- `gen_ai.completions.0.content = "I am asked to ... I should ..."`

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
Thanks! @dinmukhamedm! We definitely need this :) 
(Wondering if OpenAI has something similar)

dinmukhamedm:
Awesome! I'll open a PR later this week. I think OpenAI doesn't return it's thinking contents by design (as of now)

nirga:
I think they do now with the new response api (for which we also need to add support)

dinmukhamedm:
@nirga I created #2782 for this

PR #2780, opened 2025-03-15
fix(anthropic): add thinking as a separate completion message

dinmukhamedm:
Addresses #2774 

This PR introduces many small changes:
1. #2774 itself, i.e. add a thinking block as `gen_ai.completion.0.role = "thinking"` and `gen_ai.completion.0.content` for both streaming and non-streaming.
2. Updates `anthropic` SDK dependency to support streaming in test. As a result, the following few changes:
    - Handle the case when `cache_creation_input_tokens` and `cache_read_input_tokens` is explicitly set to `None`. I think this was not the case in older SDK, and this was breaking token usage on both spans and metrics
    - `count_tokens` is not available in the SDK itself anymore. There is a `anthropic.beta.messages.count_tokens()` now that accepts a `messages` array and a model name (starting from `claude-2`), but it is actually a call to the API – I don't think we need to impose additional latency and network
3. Tiny refactor around how cache tokens are counted
4. Split the test file and move cassettes accordingly – thus, the huge change

----

> [!IMPORTANT]
> This PR adds 'thinking' as a separate completion message in the Anthropic integration, updates the SDK, and refactors tests to support these changes.
> 
>   - **Behavior**:
>     - Adds 'thinking' as a separate completion message with `gen_ai.completion.0.role = "thinking"` in `__init__.py`.
>     - Updates `anthropic` SDK dependency to support streaming.
>     - Handles `None` values for `cache_creation_input_tokens` and `cache_read_input_tokens` in `__init__.py`.
>     - Removes `count_tokens` usage due to SDK changes.
>   - **Tests**:
>     - Splits and refactors tests into `test_messages.py`, `test_prompt_caching.py`, and `test_thinking.py`.
>     - Adds new test cases for 'thinking' in `test_thinking.py`.
>     - Moves cassettes to new test files.
>   - **Misc**:
>     - Refactors token counting logic in `__init__.py`.
>     - Updates `verify_metrics` in `utils.py` to handle new metrics.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 7b39ca059ddffafca5ca3699f4cb3bb85ce47b98. It will automatically update as commits are pushed.</sup>


dinmukhamedm:
@nirga any thoughts/feedback?

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -52,8 +55,8 @@ def _set_token_usage(
     token_histogram: Histogram = None,
     choice_counter: Counter = None,
 ):
-    cache_read_tokens = complete_response.get("usage", {}).get("cache_read_input_tokens", 0)
-    cache_creation_tokens = complete_response.get("usage", {}).get("cache_creation_input_tokens", 0)
+    cache_read_tokens = complete_response.get("usage", {}).get("cache_read_input_tokens", 0) or 0
```
nirga, line 58: nit: you don't need this since get returns 0 by default

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -52,8 +55,8 @@ def _set_token_usage(
...
+    cache_creation_tokens = complete_response.get("usage", {}).get("cache_creation_input_tokens", 0) or 0
```
nirga, line 59: same

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -159,13 +164,13 @@ def build_from_streaming_response(
             completion_tokens = -1
             # prompt_usage
             if usage := complete_response.get("usage"):
-                prompt_tokens = usage.get("input_tokens", 0)
+                prompt_tokens = usage.get("input_tokens", 0) or 0
```
nirga, line 167: same?

nirga (commented):
sorry for the delay @dinmukhamedm! I looked at it on my phone and I was like "ok I need my laptop for this it's a huge PR" and then forgot about it 😨 
looks great, just had a super small nit comment - can you fix?

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
dinmukhamedm: No, the thing is that the new anthropic SDK explicitly sets these fields to `None`, and so get, returns a `None`, which fails below, where we try to add things up.

```
>>> d = {"key": "val", "none_key": None}
>>> v = d.get("none_key", "fallback")
>>> print (v)
None
>>> 
```

same for all the changes below

dinmukhamedm:
No worries @nirga ! Responded to the comment

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
nirga: Oh 
I don't know if I hate python more or Anthropic more. Thanks!

Issue #2553, opened 2025-01-25
🐛 Bug Report: All OpenAI Assistant messages are registered as completions, even if they are user messages


dinmukhamedm:
### Which component is this bug for?

OpenAI Instrumentation

### 📜 Description

Open AI assistant is initialized with one or more system messages, and then the rest of the chat history may be added later in the run. These followup messages are always recorded as `gen_ai.completions.N.content` even if the message was created by the user.

Yes, I know that
- There is room for discussions if `completions.role=user` makes sense, and if the role must be the source of truth
- Work on migrating to the events instead of attributes is going on, and conventions will be slightly different.

But, given that this should be a fairly small change, I'm suggesting we update this

### 👟 Reproduction steps

1. Run the code below with some observability configured
2. Have a look at the attributes of the resulting span with name `openai.assistant.run`

```python
import time
from openai import OpenAI

NOT_GIVEN = None

class MyAssistant:
    def __init__(self):
        self.client = OpenAI(
        )
        self.assistant = self.client.beta.assistants.create(
            name="Math Tutor",
            instructions="You are a personal math tutor. Write and run code to answer math questions.",
            tools=[{"type": "code_interpreter"}],
            model="gpt-4o-mini",
        )

    def execute(self):
        self.thread = self.client.beta.threads.create()
        self._create_thread_message(
            "I need to solve the equation `3x + 11 = 14`. Can you help me?"
        )
        run = self._create_run(
            instructions="Please address the user as Jane Doe. The user has a premium account."
        )
        self._process_run(run)

    def _create_thread_message(self, prompt: str):
        return self.client.beta.threads.messages.create(
            thread_id=self.thread.id,
            role="user",
            content=prompt,
        )

    def _create_run(
        self,
        tool_override: None = None,
        instructions: str | None = None,
    ):
        return self.client.beta.threads.runs.create(
            thread_id=self.thread.id,
            assistant_id=str(self.assistant.id),
            model=self.assistant.model if self.assistant.model else NOT_GIVEN,
            instructions=instructions,
            temperature=(
                self.assistant.temperature if self.assistant.temperature else NOT_GIVEN
            ),
            top_p=self.assistant.top_p if self.assistant.top_p else NOT_GIVEN,
            tools=self.assistant.tools if self.assistant.tools else NOT_GIVEN,
            tool_choice=None,
        )

    def _process_run(self, run):
        while run.status in ["queued", "in_progress", "cancelling"]:
[...]

PR #2781, opened 2025-03-15
fix(openai): set user messages as prompts, not completions

dinmukhamedm:
Addresses #2553. This saves user messages as prompts, and only the actual assistant responses as completions.

This is a quick fix for non-streaming, but when I looked at streaming, I realised we'd need to change our approach completely to add user messages at all. We would need to wrap `messages_create`. I decided not to make major changes, given the apparently low interest in the assistant API. @nirga what do you think? I can open a new issue though to at least track this

----

> [!IMPORTANT]
> Fix user messages being incorrectly saved as completions by categorizing them as prompts in `messages_list_wrapper`.
> 
>   - **Behavior**:
>     - User messages are now saved as prompts, not completions, in `messages_list_wrapper`.
>     - Only assistant responses are saved as completions.
>     - Streaming scenarios are not addressed; future changes may be needed.
>   - **Code Changes**:
>     - Introduced `prompt_index` and `completion_index` to differentiate between prompts and completions in `assistant_wrappers.py`.
>     - Updated logic to handle user and system roles as prompts.
>   - **Tests**:
>     - Updated `test_assistant.py` to verify user messages are categorized as prompts.
>     - Adjusted test assertions to reflect new prompt and completion indexing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 9b4d25eee51639b4ea8ea4fe6e8a437d0bca0b34. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
+++ b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
@@ -106,7 +106,7 @@ def messages_list_wrapper(tracer, wrapped, instance, args, kwargs):
         start_time=run.get("start_time"),
     )
 
-    i = 0
+    prompt_index = 0
     if assistants.get(run["assistant_id"]) is not None or Config.enrich_assistant:
         if Config.enrich_assistant:
             assistant = model_as_dict(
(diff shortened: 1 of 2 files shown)


nirga (approved):
I think this api is going to be deprecated anyway - and we should work on supporting the new responses api

PR #2788, opened 2025-03-18
feat(instrumentation): add metric for Bedrock prompt caching

thisthat:

image


nirga (approved):
Thanks @thisthat! I think it might also be useful to mark the span as a span with cached prompt

PR #2805, opened 2025-04-03
fix(langchain): ensure llm span is created

### LINKED ISSUE #2271: 🐛 Bug Report: OpenAI Requests Not Traced When Sent from a LangGraph Node

Langchain Instrumentation

### 📜 Description

In my Langchain-based application using LangGraph, I noticed that OpenAI requests with OpenAI's own client within a LangGraph node are not traced. Specifically, when I call the OpenAI GPT-4o model from within a LangGraph node, I do not see a span related to the OpenAI call in my exported trace log, nor do I see any associated LLM call metrics.

### 👟 Reproduction steps

Here is an example demonstrating the issue:
```python
from openai import OpenAI
from dotenv import load_dotenv
import os
from typing import TypedDict
from langgraph.graph import StateGraph

from opentelemetry.sdk.trace.export import ConsoleSpanExporter

# Load environment variables
load_dotenv()

# Setup directories for logs
logs_dir = "path/to/logs"
langtrace_logs_dir = os.path.join(logs_dir, "traceloop")
traceloop_log_file_path = os.path.join(langtrace_logs_dir, "traceloop_issue_example.log")
traceloop_log_file = open(traceloop_log_file_path, "w")

# Initialize Traceloop with ConsoleSpanExporter
exporter = ConsoleSpanExporter(out=traceloop_log_file)
Traceloop.init(disable_batch=True, exporter=exporter)

client = OpenAI()

# Define state for LangGraph
class State(TypedDict):
    request: str
    result: str

# Define a calculation node
def calculate(state: State):
    request = state["request"]
    completion = client.chat.completions.create(
        model="gpt-4o",
        messages=[
            {"role": "system", "content": "You are a mathematician."},
            {"role": "user", "content": request}
        ]
    )
    return {"result": completion.choices[0].message.content}

# Create the workflow graph
workflow = StateGraph(State)
workflow.add_node("calculate", calculate)
workflow.set_entry_point("calculate")

langgraph = workflow.compile()

# Invoke the graph
user_request = "What's 5 + 5?"
result = langgraph.invoke(input={"request": user_request})

print(f"Request: {user_request}")
print(f"Result: {result['result']}")
```
In the trace logs, there are no gen_ai attributes or metrics for the OpenAI call. However, if I replace the OpenAI client with Langchain’s own OpenAI client, a span with LLM metrics is generated as expected.

### 👍 Expected behavior

When an OpenAI call is made within a LangGraph node, I expect to see tracing data that includes gen_ai attributes and other metrics associated with the LLM call, such as:
 ```json
{
    "name": "openai.chat",
    "context": {
        "trace_id": "0x993046216162c1a4dddf7c3484062d26",
        "span_id": "0xcfeb41cffa9b1021",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": null,
    "start_time": "2024-11-06T14:30:28.540006Z",
    "end_time": "2024-11-06T14:30:31.680890Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "llm.request.type": "chat",
        "gen_ai.system": "OpenAI",
        "gen_ai.request.model": "gpt-4o",
        "llm.headers": "None",
        "llm.is_streaming": false,
        "gen_ai.openai.api_base": "https://api.openai.com/v1/",
        "gen_ai.prompt.0.role": "system",
        "gen_ai.prompt.0.content": "You are a mathematician.",
        "gen_ai.prompt.1.role": "user",
        "gen_ai.prompt.1.content": "claculate 5 + 5",
        "gen_ai.response.model": "gpt-4o-2024-08-06",
        "gen_ai.openai.system_fingerprint": "fp_45cf54deae",
        "llm.usage.total_tokens": 39,
        "gen_ai.usage.completion_tokens": 15,
        "gen_ai.usage.prompt_tokens": 24,
        "gen_ai.completion.0.finish_reason": "stop",
        "gen_ai.completion.0.role": "assistant",
        "gen_ai.completion.0.content": "The result of \\(5 + 5\\) is \\(10\\)."
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "/Users/muhammadkanaan/Desktop/work_repos/agent-analytics/insighter/src/test/openai/openai_llm_call.py"
        },
        "schema_url": ""
    }
}


When running the code from within a LangGraph node, no tracing data is recorded for the OpenAI calls. Below is the actual log output captured during the execution: 
 ```json
{
    "name": "__start__.task",
    "context": {
        "trace_id": "0xc01627149044876058dc3ed2d419a5aa",
        "span_id": "0x1b2becef5d6179ab",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0xc77e31c8e27e4710",
    "start_time": "2024-11-06T14:22:00.764176Z",
    "end_time": "2024-11-06T14:22:00.764384Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "traceloop.association.properties.langgraph_step": 0,
        "traceloop.association.properties.langgraph_node": "__start__",
        "traceloop.association.properties.langgraph_triggers": [
            "__start__"
        ],
        "traceloop.association.properties.langgraph_path": [
            "__pregel_pull",
            "__start__"
        ],
        "traceloop.association.properties.langgraph_checkpoint_ns": "__start__:91fbf40a-29dd-606a-8186-ca138e2e7803",
        "traceloop.workflow.name": "LangGraph",
        "traceloop.entity.path": "",
        "traceloop.span.kind": "task",
        "traceloop.entity.name": "__start__",
        "traceloop.entity.input": "{\"inputs\": {\"request\": \"whats 5 + 5\"}, \"tags\": [\"graph:step:0\", \"langsmith:hidden\", \"langsmith:hidden\"], \"metadata\": {\"langgraph_step\": 0, \"langgraph_node\": \"__start__\", \"langgraph_triggers\": [\"__start__\"], \"langgraph_path\": [\"__pregel_pull\", \"__start__\"], \"langgraph_checkpoint_ns\": \"__start__:91fbf40a-29dd-606a-8186-ca138e2e7803\"}, \"kwargs\": {\"name\": \"__start__\"}}",
        "traceloop.entity.output": "{\"outputs\": {\"request\": \"whats 5 + 5\"}, \"kwargs\": {\"tags\": [\"graph:step:0\", \"langsmith:hidden\", \"langsmith:hidden\"]}}"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "/Users/muhammadkanaan/Desktop/work_repos/agent-analytics/insighter/src/test/openai/tmp.py"
        },
        "schema_url": ""
    }
}
{
    "name": "ChannelWrite<calculate,request,result>.task",
    "context": {
        "trace_id": "0xc01627149044876058dc3ed2d419a5aa",
        "span_id": "0x2e1ca943f99c35fd",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0x822c13f276dd7dc7",
    "start_time": "2024-11-06T14:22:01.836989Z",
    "end_time": "2024-11-06T14:22:01.837304Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "traceloop.association.properties.langgraph_step": 1,
        "traceloop.association.properties.langgraph_node": "calculate",
        "traceloop.association.properties.langgraph_triggers": [
            "start:calculate"
        ],
        "traceloop.association.properties.langgraph_path": [
            "__pregel_pull",
            "calculate"
        ],
        "traceloop.association.properties.langgraph_checkpoint_ns": "calculate:ba4fa053-c8d2-c4dd-5456-f56c9095719e",
        "traceloop.workflow.name": "LangGraph",
        "traceloop.entity.path": "calculate",
        "traceloop.span.kind": "task",
        "traceloop.entity.name": "ChannelWrite<calculate,request,result>",
        "traceloop.entity.input": "{\"inputs\": {\"result\": \"\\\\(5 + 5 = 10\\\\).\"}, \"tags\": [\"seq:step:2\", \"langsmith:hidden\"], \"metadata\": {\"langgraph_step\": 1, \"langgraph_node\": \"calculate\", \"langgraph_triggers\": [\"start:calculate\"], \"langgraph_path\": [\"__pregel_pull\", \"calculate\"], \"langgraph_checkpoint_ns\": \"calculate:ba4fa053-c8d2-c4dd-5456-f56c9095719e\"}, \"kwargs\": {\"name\": \"ChannelWrite<calculate,request,result>\"}}",
        "traceloop.entity.output": "{\"outputs\": {\"result\": \"\\\\(5 + 5 = 10\\\\).\"}, \"kwargs\": {\"tags\": [\"seq:step:2\", \"langsmith:hidden\"]}}"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "/Users/muhammadkanaan/Desktop/work_repos/agent-analytics/insighter/src/test/openai/tmp.py"
        },
        "schema_url": ""
    }
}
{
    "name": "calculate.task",
    "context": {
        "trace_id": "0xc01627149044876058dc3ed2d419a5aa",
        "span_id": "0x822c13f276dd7dc7",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0xc77e31c8e27e4710",
    "start_time": "2024-11-06T14:22:00.766167Z",
    "end_time": "2024-11-06T14:22:01.837811Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "traceloop.association.properties.langgraph_step": 1,
        "traceloop.association.properties.langgraph_node": "calculate",
        "traceloop.association.properties.langgraph_triggers": [
            "start:calculate"
        ],
        "traceloop.association.properties.langgraph_path": [
            "__pregel_pull",
            "calculate"
        ],
        "traceloop.association.properties.langgraph_checkpoint_ns": "calculate:ba4fa053-c8d2-c4dd-5456-f56c9095719e",
        "traceloop.workflow.name": "LangGraph",
        "traceloop.entity.path": "",
        "traceloop.span.kind": "task",
        "traceloop.entity.name": "calculate",
        "traceloop.entity.input": "{\"inputs\": {\"request\": \"whats 5 + 5\"}, \"tags\": [\"graph:step:1\"], \"metadata\": {\"langgraph_step\": 1, \"langgraph_node\": \"calculate\", \"langgraph_triggers\": [\"start:calculate\"], \"langgraph_path\": [\"__pregel_pull\", \"calculate\"], \"langgraph_checkpoint_ns\": \"calculate:ba4fa053-c8d2-c4dd-5456-f56c9095719e\"}, \"kwargs\": {\"name\": \"calculate\"}}",
        "traceloop.entity.output": "{\"outputs\": {\"result\": \"\\\\(5 + 5 = 10\\\\).\"}, \"kwargs\": {\"tags\": [\"graph:step:1\"]}}"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "/Users/muhammadkanaan/Desktop/work_repos/agent-analytics/insighter/src/test/openai/tmp.py"
        },
        "schema_url": ""
    }
}
{
    "name": "LangGraph.workflow",
    "context": {
        "trace_id": "0xc01627149044876058dc3ed2d419a5aa",
        "span_id": "0xc77e31c8e27e4710",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": null,
    "start_time": "2024-11-06T14:22:00.763510Z",
    "end_time": "2024-11-06T14:22:01.838540Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "traceloop.workflow.name": "LangGraph",
        "traceloop.entity.path": "",
        "traceloop.span.kind": "workflow",
        "traceloop.entity.name": "LangGraph",
        "traceloop.entity.input": "{\"inputs\": {\"request\": \"whats 5 + 5\"}, \"tags\": [], \"metadata\": {}, \"kwargs\": {\"name\": \"LangGraph\"}}",
        "traceloop.entity.output": "{\"outputs\": {\"request\": \"whats 5 + 5\", \"result\": \"\\\\(5 + 5 = 10\\\\).\"}, \"kwargs\": {\"tags\": []}}"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "/Users/muhammadkanaan/Desktop/work_repos/agent-analytics/insighter/src/test/openai/tmp.py"
        },
        "schema_url": ""
    }
}

As shown, the logs lack the expected gen_ai and llm attributes or metrics related to the OpenAI call, which would normally be included when using Langchain's OpenAI client directly.

### 🤖 Python Version

Python 3.12.4

langchain==0.2.16
langchain-cohere==0.1.9
langchain-community==0.2.17
langchain-core==0.2.41
langchain-experimental==0.0.65
langchain-openai==0.1.25
langchain-text-splitters==0.2.4
langgraph==0.2.23
langgraph-checkpoint==1.0.10

openai==1.47.0

traceloop-sdk==0.33.3

obs-gh-abhishekrao:

Fixes #2271 

## Reproducible code from bug report
Courtesy #2271. Thanks @jemo21k!
Expected behaviour: LLM span should be created.
### Before
No `openai.chat` span.
image

### After
`openai.chat` span is present.
image

## sample_app/langgraph_example
Expected behaviour: No change. Additional LLM span should not be created since `ChatOpenAI` callback already creates one.
### Before
image

### After
image

----

> [!IMPORTANT]
> Fixes LLM span creation and context retention in Langchain and OpenAI instrumentation by setting default values for missing attributes and ensuring spans are correctly attached to context.
> 
>   - **Behavior**:
>     - Fixes missing `openai.chat` span creation in `callback_handler.py` by ensuring `LLM_RESPONSE_MODEL` is set to `model` or `"unknown"`.
>     - Ensures parent context is retained and appropriate LLM span is created in `_create_llm_span()` in `callback_handler.py`.
>   - **Attributes**:
>     - Sets `LLM_RESPONSE_MODEL` to `model` or `"unknown"` in `_build_from_streaming_response()` and `_abuild_from_streaming_response()` in `chat_wrappers.py`.
>     - Updates `messages_list_wrapper()` and `runs_create_and_stream_wrapper()` in `assistant_wrappers.py` to set `LLM_RESPONSE_MODEL` to `model` or `"unknown"`.
>   - **Misc**:
>     - Minor updates to ensure spans are correctly attached to context in `callback_handler.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 08116d985397a18e342bd5764ae3e12fe83ae503. It will automatically update as commits are pushed.</sup>


nirga (commented):
Thanks for submitting @obs-gh-abhishekrao - can you take a look into the failing tests, lint and sign the CLA?

obs-gh-abhishekrao:
Hi @nirga, sorry it's taking time, I'm still working on getting the CLA signed.

Took a look a the test failures. The langchain test failure happening on SequentialChain raises some suspicions for me. It looks like context isn't updated / propagated during `on_llm_start` for `SequentialChain.ainvoke`. What seems to be happening is the OpenAI instrumentation is triggered (asynchronously?) as a child of the task span as opposed to as the child of the `_create_llm_span` call, and hence ends up not seeing `SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY`

So far I'm wondering if this is because the langchain instrumentation uses BaseCallbackHandler and (probably) missing an [AsyncCallbackHandler](https://github.com/langchain-ai/langchain/blob/4f9f97bd124012d65a9c7de531d5697a7077a32d/libs/core/langchain_core/callbacks/base.py#L477).
Some references I was reading into:
- https://python.langchain.com/docs/concepts/callbacks/#callback-handlers
- https://python.langchain.com/docs/how_to/callbacks_async/
The second link mentions this:
```
If you use a sync CallbackHandler while using an async method to run your LLM / Chain / Tool / Agent, it will still work. However, under the hood, it will be called with [run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) which can cause issues if your CallbackHandler is not thread-safe.
```
 I tried adding [opentelemetry-instrumentation-threading](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/threading/threading.html) to the test to see if it solves the context propagation but it didn't help.

Suggestions would be really helpful here, so I'm curious to hear your thoughts.

obs-gh-abhishekrao:
@nirga Please review the PR at your convenience. CLA is signed.

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
```diff
@@ -213,7 +213,7 @@ def runs_create_and_stream_wrapper(tracer, wrapped, instance, args, kwargs):
         )
         _set_span_attribute(
             span,
-            SpanAttributes.LLM_RESPONSE_MODEL,
+            SpanAttributes.LLM_RESPONSE_MODEL or "unknown",
```
nirga, line 216: why do you need this?

nirga (commented):
Thanks @obs-gh-abhishekrao - I think it's almost there. Can you add a test for this specific use case?

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
obs-gh-abhishekrao: Hmm that was incorrectly coalesced. The reason I added this was that I'd encountered silent failures in the metric generation which I resolved in https://github.com/traceloop/openllmetry/pull/2805/commits/4263495f901f232adb954c1e8b8225146887cefa. I was hoping similar logic was needed here. Although, I don't see any metrics generation as part of the assistant wrapper
My latest commit https://github.com/traceloop/openllmetry/pull/2805/commits/a2daa8646010695991c3c4ecc77c576e7f426548 fixes the coalescing mistake, but I'm happy to remove it altogether too.

obs-gh-abhishekrao:
Added test based on the sample code, but noticing some unexpected behaviour that I'm still debugging.
- If I run the sample code in isolation, I'm able to successfully validate this PR
- If I run the same code through the test, it does not produce spans `openai.chat` at all, as well as forsome other operations.
For what it's worth, I tried to link our own LLM app to this PR, and saw the `openai.chat` spans produced correctly.

My analysis so far: 
Somehow during the test, the `SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY` is set to True, so the openai instrumentor ends up not creating a span at all. However, going by this PR, this context key is only set as part of `_create_llm_span` in langchain instrumentor, which isn't even called through the test code, so I'm not sure where else or how it's getting set.

obs-gh-abhishekrao:
Found the reason 

> Added test based on the sample code, but noticing some unexpected behaviour that I'm still debugging.
> 
>     * If I run the sample code in isolation, I'm able to successfully validate this PR
> 
>     * If I run the same code through the test, it does not produce spans `openai.chat` at all, as well as forsome other operations.
>       For what it's worth, I tried to link our own LLM app to this PR, and saw the `openai.chat` spans produced correctly.
> 
> 
> My analysis so far: Somehow during the test, the `SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY` is set to True, so the openai instrumentor ends up not creating a span at all. However, going by this PR, this context key is only set as part of `_create_llm_span` in langchain instrumentor, which isn't even called through the test code, so I'm not sure where else or how it's getting set.

This happens because all the (langchain) tests are happening under a shared trace context (and hence all spans nest under the same trace). So, `test A` may set a context key and `test B` ends up accidentally reading it in its OpenAI instrumentor. Adding a pytest fixture that starts a root span before each test solves this. However some cases tend to fail now, which need fixing.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -413,9 +413,7 @@ def _create_span(
         _set_span_attribute(span, SpanAttributes.TRACELOOP_WORKFLOW_NAME, workflow_name)
         _set_span_attribute(span, SpanAttributes.TRACELOOP_ENTITY_PATH, entity_path)
 
-        token = context_api.attach(
-            context_api.set_value(SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY, True)
-        )
+        token = context_api.attach(set_span_in_context(span))
```
ronensc, line 416: I believe this should have a corresponding `detach()` in `_end_span()`. Without it, the context of the caller could be altered, potentially breaking the parent-child relationship between spans and causing unrelated spans to be linked together.

Consider the following unit test, which currently fails due to this issue:
```python
@pytest.mark.vcr
def test_langgraph_double_invoke(exporter):
    class DummyGraphState(TypedDict):
        result: str

    def mynode_func(state: DummyGraphState) -> DummyGraphState:
        return state

    def build_graph():
        workflow = StateGraph(DummyGraphState)
        workflow.add_node("mynode", mynode_func)
        workflow.set_entry_point("mynode")
        langgraph = workflow.compile()
        return langgraph
    
    graph = build_graph()
    assert "test_langgraph_double_invoke" == trace.get_current_span().name
    graph.invoke({"result": "init"})
    assert "test_langgraph_double_invoke" == trace.get_current_span().name # incorrectly set to 'mynode.task' 
    graph.invoke({"result": "init"})
    assert "test_langgraph_double_invoke" == trace.get_current_span().name # still incorrectly set to 'mynode.task' 
```
As you can see, the span name unexpectedly switches to 'mynode.task', indicating that the span context is not properly detached after each invocation.

Please take a look at my proposed fix here:
[Fix Suggestion](https://github.com/traceloop/openllmetry/commit/23d1059b2a7d971881684b49d09b30fb99e0eecb)

Feel free to review and adjust as needed.

ronensc (changes requested):
@obs-gh-abhishekrao Thanks for working on fixing this issue! I’ve come across a subtle problem with the proposed fix. Please see the details below.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
obs-gh-abhishekrao: Thanks @ronensc! Great catch, will look into adding this test and fix.

packages/opentelemetry-instrumentation-langchain/tests/test_langgraph.py
```diff
@@ -0,0 +1,150 @@
...
+            "__start__.task",
+            "calculate.task",
+            "_write.task",
+            "openai.chat"
+        ]
+    ) == set([span.name for span in spans])
+
+
+@pytest.mark.vcr
+def test_langgraph_double_invoke(exporter):
```
ronensc, line 85: The `test_langgraph_double_invoke()` and `test_langgraph_double_ainvoke()` don't catch the regression I mentioned.
If I comment out the `detach()` invocation, the tests pass even though they should have fail.
https://github.com/traceloop/openllmetry/pull/2805/files#diff-caff857eafc9121a585e40cdac1c0815d97f474cd3d0bf1a23755572e3b38105R702-R703

obs-gh-abhishekrao: Apologies, I agree the test case is incorrect. However the fix you suggested earlier does not work for me (at least not locally). Not only does the (fixed) test fail but I'm also noticing there's a problem with detaching context in `_end_span`. There appears to be a flurry of these errors when it's detached there, possibly due to async behaviour:
```
ValueError: <Token var=<ContextVar name='current_context' default={} at 0xXXXXX> at 0xYYYYY> was created in a different Context
```

ronensc (commented):
I've reviewed the rest of the PR and shared a few thoughts and questions.

packages/opentelemetry-instrumentation-langchain/pyproject.toml
```diff
@@ -28,6 +28,7 @@ opentelemetry-api = "^1.28.0"
 opentelemetry-instrumentation = ">=0.50b0"
 opentelemetry-semantic-conventions = ">=0.50b0"
 opentelemetry-semantic-conventions-ai = "0.4.9"
+langgraph = "^0.4.3"
```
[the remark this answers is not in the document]
obs-gh-abhishekrao: Sorry, this was added erroneously. Will remove it.

packages/opentelemetry-instrumentation-langchain/tests/test_langgraph.py
ronensc: Thanks, I'm able to reproduce the issue.

It looks like the root cause is that the `on_XXX_end()` callbacks run in a different task (and thus a different context) from the `on_XXX_start()` callbacks. I added the following lines to each callback to verify this:

```python
task = asyncio.current_task()
print(f"Task ID: {id(task)} Task name: {task.get_name()}")
```

I also came across this [LangChain discussion](https://github.com/langchain-ai/langchain/discussions/27954), though it hasn't received any input from the maintainers yet.

From what I can tell, the `on_XXX_end()` callbacks are dispatched on a different task due to the use of the `@shielded` decorator:
https://github.com/langchain-ai/langchain/blob/1ebcbf1d11578cb55db26013c745dc1e5722966e/libs/core/langchain_core/callbacks/manager.py#L962-L963

Removing the decorator might resolve the issue, but I couldn't find any explanation in the [PR](https://github.com/langchain-ai/langchain/pull/17010/files#diff-829fac3edfc0df71d5134d603bb03183d2d7ca3fb7fd13cfcdd4c944a7d8b335R880-R881) that introduced it for why it was added in the first place.

obs-gh-abhishekrao: I'm of the opinion that we should skip fixing span relationships for now and retain behavior from `main`. The main issue this PR addresses is missing downstream spans, which should be fixed.  Aside from reporting upstream in langchain, I've also asked the OTel folks if there's any gotchas with shielded functions. I have however retained the double invoke/ainvoke tests for posterity for when we get to addressing the span relationships.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -327,8 +327,8 @@ def _sanitize_metadata_value(value: Any) -> Any:
         return None
     if isinstance(value, (bool, str, bytes, int, float)):
         return value
-    if isinstance(value, (list, tuple)):
```
nirga: revert this?

nirga (commented):
Thanks @obs-gh-abhishekrao but can you help me understand what does this changes in terms of functionality? I'm not sure it resolves the context propagation issue, right?

packages/opentelemetry-instrumentation-langchain/tests/test_langgraph.py
ronensc: AFAIU, `context_api.detach()` should be invoked only from the context that created the token and the `contextvars` package doesn't have an API to get the current context (only a copy of it).

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
obs-gh-abhishekrao: There's some edge case that's causing span attributes and metrics to not get added / computed. I (locally) added tests but could not find a way to replicate this. 
However, the sample code when run through `sample-app/` consistently reproduces the missing spans and metrics.

I took a closer look, and have narrowed it down to this line of code..
https://github.com/traceloop/openllmetry/blob/6d13035c3f5723974782b16dfe8ab6624180970b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/__init__.py#L255
What's really strange is between the above change and `main`, there is no difference in the response being passed to `model_as_dict` and subsequently `model.model_dump()`

All that said I'm happy to revert this and file a separate bug. I can also add the sample code to sample-app for posterity.

obs-gh-abhishekrao:
> Thanks @obs-gh-abhishekrao but can you help me understand what does this changes in terms of functionality? I'm not sure it resolves the context propagation issue, right?

Thanks for reviewing again @nirga. Apologies for the flurry of commits. 
Initially, I thought I could address the missing LLM span as well as fix the parent span scope from the langchain task / workflow that triggers it. However, based on @ronensc's feedback and further discovery, `context.attach` and `context.detach` don't work predictably within the scope of langchain callbacks.

To sum it up - the PR will address 
- The main concern i.e., missing LLM spans
- The newly created LLM spans did not have many span attributes. I addressed that as well.

Context will not be propagated from langchain to downstreams unfortunately, and this means the new LLM spans get disconnected from the trace.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
obs-gh-abhishekrao: Found the root cause and fixed it (and reverted the change from langchain). This was actually happening within the SDK here https://github.com/traceloop/openllmetry/blob/6d13035c3f5723974782b16dfe8ab6624180970b/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L1067

ronensc:
@obs-gh-abhishekrao Building on one of your earlier suggestions, would it be possible to implement `context.attach()` and `context.detach()` around the non-async callbacks to handle context propagation at least in those cases, and leave the async cases for future work?

obs-gh-abhishekrao:
> @obs-gh-abhishekrao Building on one of your earlier suggestions, would it be possible to implement `context.attach()` and `context.detach()` around the non-async callbacks to handle context propagation at least in those cases, and leave the async cases for future work?

@ronensc Ack. I think this is doable. I'll update the PR once I have it.

packages/opentelemetry-instrumentation-langchain/tests/test_langgraph.py
ronensc, line 1: Would it make sense to add an xfail test for the async case? For example:
```python
@pytest.mark.vcr
@pytest.mark.asyncio
@pytest.mark.xfail(reason="Context propagation is not yet supported for async LangChain callbacks", strict=True)
async def test_langgraph_ainvoke_xfail(exporter):
    client = OpenAI()

    class State(TypedDict):
        request: str
        result: str

    def calculate(state: State):
        request = state["request"]
        completion = client.chat.completions.create(
            model="gpt-4o",
            messages=[
                {"role": "system", "content": "You are a mathematician."},
                {"role": "user", "content": request}
            ]
        )
        return {"result": completion.choices[0].message.content}
    workflow = StateGraph(State)
    workflow.add_node("calculate", calculate)
    workflow.set_entry_point("calculate")

    langgraph = workflow.compile()

    user_request = "What's 5 + 5?"
    await langgraph.ainvoke(input={"request": user_request})
    assert set(
        [
            "LangGraph.workflow",
            "calculate.task",
            "openai.chat"
        ]
    ) == set([span.name for span in spans])
    openai_span = next(span for span in spans if span.name == "openai.chat")
    calculate_task_span = next(span for span in spans if span.name == "calculate.task")

    assert openai_span.parent.span_id == calculate_task_span.context.span_id
```

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -180,12 +193,28 @@ def __call__(
...
             instance.add_handler(self._callback_manager, True)
+            if isinstance(instance, (
+                    CallbackManager,
+                    CallbackManagerForChainGroup,
+                    CallbackManagerForChainRun,
+                    CallbackManagerForLLMRun,
+                    CallbackManagerForRetrieverRun,
+                    CallbackManagerForToolRun,
+                )
+            ):
```
ronensc: Wouldn't it be more straightforward to utilize the [`is_async`](https://python.langchain.com/api_reference/core/callbacks/langchain_core.callbacks.base.BaseCallbackManager.html) attribute here?

```suggestion
            if not instance.is_async:
```

obs-gh-abhishekrao: Great callout!

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -180,12 +185,20 @@ def __call__(
...
         wrapped(*args, **kwargs)
         for handler in instance.inheritable_handlers:
             if isinstance(handler, type(self._callback_manager)):
                 break
         else:
             instance.add_handler(self._callback_manager, True)
+            if not instance.is_async:
+                if span and not self._callback_manager.spans[parent_run_id].token:
+                    token = context_api.attach(set_span_in_context(span))
+                    self._callback_manager.spans[parent_run_id].token = token
```
ronensc: @obs-gh-abhishekrao  What do you think about moving the `if not async` logic back into `_create_span()` in the callback handler?
That way, the symmetry between `attach`/`detach` and `_create_span`/`_end_span` would be clearer.

obs-gh-abhishekrao: It won't be the same logic because that would be moving into the scope of the callback handler itself, where we lose sight of the `is_async` property. I might have to do something like `asyncio.get_running_loop()` with a try-except which doesn't feel like a clean approach. Let me know your thoughts on that.

ronensc: Agreed, `asyncio.get_running_loop()` doesn’t feel like a clean solution.

My main concern is that `attach` lives in the **CallbackManager**, while `detach` is in the **CallbackHandler**. I believe both should reside in the handler, with `is_async` passed in as needed.

Here’s a draft illustrating that approach (note: still needs debugging to get all tests passing): https://github.com/traceloop/openllmetry/commit/84ddc73c4cff1b80b0ef1d26eb0d1e4912e115a4

obs-gh-abhishekrao: I like the reasoning but the solution gives me doubt. As much as it might work, you're overriding a sync method to be async, which causes type errors and introduces a ton of new code complexity. Here's what I tried instead, which involves a few lines of code and doesn't require spinning up a new handler class:
1. Set a property of TraceloopCallbackHandler instance
```python
if instance.is_async:
    self._callback_manager.is_async = True
```
2. Use the property to attach context
```python
token = None
if not self.is_async:
    token = context_api.attach(set_span_in_context(span))
```
Now what's really strange is that the test passes in isolation, but when all the tests are run at once (i.e., `nx run opentelemetry-instrumentation-langchain:test`) the sync test fails because `if instance.is_async:` somehow turns out to be true :-(
Here are some debug logs showing this conflict:
```
request.node.name='test_langgraph_invoke'
<langchain_core.callbacks.manager.CallbackManager object at 0x13fcbe410> instance.is_async=False # wrapper
self.is_async=True # during attach
<langchain_core.callbacks.manager.CallbackManager object at 0x13fcbfc10> instance.is_async=False # wrapper
<langchain_core.callbacks.manager.CallbackManager object at 0x13fcbcb10> instance.is_async=False # wrapper
self.is_async=True # during attach
<langchain_core.callbacks.manager.CallbackManager object at 0x13fcd9d50> instance.is_async=False # wrapper
<langchain_core.callbacks.manager.CallbackManager object at 0x13f09c6d0> instance.is_async=False # wrapper
```

obs-gh-abhishekrao: Committed something that may be viable. Please have a look!

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -191,6 +196,7 @@ def __call__(
             if isinstance(handler, type(self._callback_manager)):
                 break
         else:
+            self._callback_manager._instance = instance
```
ronensc: Maybe we could rename `_callback_manager` to `_callback_handler`, and `_instance` to `_callback_manager`? That might make the naming more intuitive.

packages/opentelemetry-instrumentation-langchain/tests/conftest.py
```diff
@@ -82,3 +84,12 @@ def clear_metrics_test_context(metrics_test_context):
 
     reader.shutdown()
     provider.shutdown()
+
+
+@pytest.fixture(scope="function", autouse=True)
+def reset_span_context(request):
+    tracer = get_tracer(__name__)
+    # starts a new trace per test function
+    # this avoids mixing of trace contexts leading to flaky results
```
ronensc: Could you clarify what flaky results `reset_span_context()` is intended to address?  
Now that `context.detach()` is in place, I believe it should resolve the flakiness.

ronensc (approved):
LGTM
@obs-gh-abhishekrao Thank you for the hard work!

Should we add a TODO comment for the async case, or document that it currently only supports the non-async path?

@nirg Could you please review the PR as well?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
ronensc: > Now what's really strange is that the test passes in isolation, but when all the tests are run at once (i.e., `nx run opentelemetry-instrumentation-langchain:test`) the sync test fails because `if instance.is_async:` somehow turns out to be true :-( 

Interesting... I think my draft also has the same issue...

packages/opentelemetry-instrumentation-langchain/tests/conftest.py
obs-gh-abhishekrao: It's not just about `detach`. Without isolating contexts, the entire unit test ends up being one giant trace and callbacks get mixed and form parent-child relationships where they shouldn't.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
obs-gh-abhishekrao: Makes sense, will add this.

obs-gh-abhishekrao:
> LGTM @obs-gh-abhishekrao Thank you for the hard work!
> 
> Should we add a TODO comment for the async case, or document that it currently only supports the non-async path?
> 
> @nirga Could you please review the PR as well?

Thanks for reviewing and approving this PR @ronensc! Just made the changes you last proposed.

packages/opentelemetry-instrumentation-langchain/tests/conftest.py
ronensc: That sounds to me like the kind of issue you'd expect from a missing `detach()`.
AFAIU, each test should start with a clean context and result in separate traces, even without calling `reset_span_context()`. If that's not the case, it might indicate a bug in the instrumentor.

obs-gh-abhishekrao: Makes sense. Double checked and I couldn't see it happening anymore so removed it 👍.

packages/opentelemetry-instrumentation-langchain/tests/test_langgraph.py
```diff
@@ -0,0 +1,194 @@
...
+        workflow.add_node("mynode", mynode_func)
+        workflow.set_entry_point("mynode")
+        langgraph = workflow.compile()
+        return langgraph
+
+    graph = build_graph()
+
+    from opentelemetry import trace
+
+    assert isinstance(trace.get_current_span(), NonRecordingSpan)
```
ronensc:
```suggestion
    assert trace.get_current_span() == INVALID_SPAN
```

**nit:** According to the [`get_current_span()`](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.get_current_span) documentation, I believe it would be more accurate to assert against `INVALID_SPAN`.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -437,13 +443,14 @@ def _create_span(
         else:
             span = self.tracer.start_span(span_name, kind=kind)
 
+        token = None
+        # TODO: update this once attach/detach works properly with async callbacks
```
nirga: what does this todo mean? can you elaborate?

nirga: you should write why it doesn't work, link to an open issue on langchain (if there is), and explain what needs to be once it's fixed

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
nirga, line 1: revert

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/assistant_wrappers.py
nirga, line 1: revert

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py
nirga, line 1: revert

packages/opentelemetry-instrumentation-langchain/tests/conftest.py
nirga, line 1: revert

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -188,10 +193,11 @@ def __call__(
     ) -> None:
         wrapped(*args, **kwargs)
         for handler in instance.inheritable_handlers:
-            if isinstance(handler, type(self._callback_manager)):
+            if isinstance(handler, type(self._callback_handler)):
                 break
         else:
-            instance.add_handler(self._callback_manager, True)
+            self._callback_handler._callback_manager = instance
```
nirga, line 202: why do you need this?

ronensc: We want to call `attach()` only in the non-async case:  
https://github.com/traceloop/openllmetry/pull/2805/files#diff-caff857eafc9121a585e40cdac1c0815d97f474cd3d0bf1a23755572e3b38105R448-R449

The challenge is that the Traceloop callback handler can't itself determine whether it's in an async context.
However, the `CallbackManager` exposes an [`is_async` property](https://python.langchain.com/api_reference/core/callbacks/langchain_core.callbacks.manager.CallbackManager.html), which is why the handler keeps a reference to the manager.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
obs-gh-abhishekrao: Updated comment, please have a look.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
obs-gh-abhishekrao: Thanks @ronensc.
@nirga For better clarity I've added a comment to this logic. Hope it's clear.

obs-gh-abhishekrao:
> @obs-gh-abhishekrao can you rebase from main? I don't have access to modify the PR :/

Done.

nirga:
@obs-gh-abhishekrao can you give me write permissions to your PR? It'll be impossible to merge if I can't rebase myself

ronensc:
@obs-gh-abhishekrao To allow edits by maintainers on this PR, you simply mark the `Allow edits from maintainers` checkbox as described here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

obs-gh-abhishekrao:
> @obs-gh-abhishekrao To allow edits by maintainers on this PR, you simply mark the `Allow edits from maintainers` checkbox as described here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

Sorry for the hassle. I don't see any such option. Apparently this is a limitation when a fork was made from an org :(. https://github.com/orgs/community/discussions/5634. I'm checking to see what my options are. Worst case, I'll have to re-create this PR off a personal fork.

obs-gh-abhishekrao:
@nirga I've submitted a different PR here - https://github.com/traceloop/openllmetry/pull/3094. I see the option to allow maintainer edits now, so it should work.

ronensc:
Since #3201 has been merged, I believe we can close this PR.

PR #2807, opened 2025-04-08
fix(llama-index): use the correct instrumentation point

### LINKED ISSUE #2799: 🐛 Bug Report: LlamaIndex Instrumentation doesn't come through

LlamaIndex Instrumentation

### 📜 Description

When sending to Dynatrace (a validated provider) the llamatrace does not show up at all. I tried a mixture of out of box and decorators to see if I could get it to trigger. Nothing worked. It did give the time lapse but just a mixture of HEAD and GET with no information. 

However, when I switched to using OpenInference it came up right away. This leads me to believe its the llamaindex instrumentor itself causing this issue. 

### 👟 Reproduction steps

1. Init TraceLoop
2. Have llamaindex code
3. Get traces

Optional
4. Specify the llamaindex instrumentor from Traceloop Packages (it's in by default but this was also tried) and start that after traceloop init

### 👍 Expected behavior

It should show the Methods, inputs, and outputs of all the llamaindex spans.

This is what it looks like with OpenInference
Image

This is what is happening with Traceloop's instrumentation

Image

### 🤖 Python Version

3.10

traceloop-sdk==0.36.1
opentelemetry-instrumentation-llamaindex==0.36.1

^ this was what was used. 

This is the one that actually worked
openinference-instrumentation-llama_index>=3.1.4

None

thisthat:
Fixes #2799

The instrumentation only attaches to `*.core` packages. Therefore, it should be the dependency used to decide if the instrumentation is triggered.

image


nirga (commented):
Thanks @thisthat! The only issue is that this package didn't exist in the past - so I wonder if we should make this change (which means we won't be backward compatible anymore)

thisthat:
Another possibility would be to create two Instruments: `llamaindex-legacy` and `llamaindex` that only differ on the pkg used to decide if the instrumentation should be kicking in. 
The current `LlamaIndexInstrumentor` class would then override the `instrument/uninstrument` methods delegating them to its two sub-instruments. WDYT @nirga? 
Based on the agreement on this, I would apply the same strategy for PineCone since I noticed a similar problem with `pinecone-client` being deprecated in favor of `pinecone` ([see docs here](https://pypi.org/project/pinecone-client/)).

nirga:
Sounds like a great plan actually!

nirga (approved):

Issue #2662, opened 2025-02-18
Feature Request: Support for Model Context Protocol (MCP)


nirga:
**Summary**
Add support for the [Model Context Protocol (MCP)](https://modelcontextprotocol.io/introduction) in openllmetry to enhance compatibility with standardized model context tracking and improve observability for generative AI workflows.

**Background**
MCP provides a standardized way to define and propagate contextual metadata across AI/ML inference workflows. Integrating MCP with openllmetry can enable better context-aware logging, tracing, and performance analysis.

**Proposed Solution**
Implement MCP-compatible context propagation in openllmetry, ensuring seamless integration with OpenTelemetry-based tracing.
Map existing openllmetry context attributes to MCP's schema where applicable.
Provide utilities or middleware to facilitate MCP adoption for users of openllmetry.
Ensure compatibility with frameworks like LangChain, LlamaIndex, and other AI model orchestration tools.

**Open Questions**
What is the best way to structure MCP integration within openllmetry?
Should we provide automatic instrumentation for MCP, or require explicit user adoption?
Are there any existing MCP-compatible libraries we can leverage?

**Next Steps**
Investigate MCP’s data model and how it aligns with openllmetry.
Design an integration approach (manual vs. automatic context propagation).
Implement a prototype and test with supported AI model frameworks.
Gather feedback from users and iterate.

gyliu513:
Some reference may help a bit:

- https://github.com/tinybirdco/mcp-tinybird/tree/main/mcp-server-analytics 
- https://www.tinybird.co/blog-posts/analyze-mcp-server-usage

fali007:
Thankyou @gyliu513 , 
We would be more than happy to add mcp observability into traceloop. Will share a draft PR at most by end of the week. We are following the semantic conventions discussed [here](https://github.com/open-telemetry/semantic-conventions/pull/2045)

gyliu513:
@fali007 Yes, traceloop first, then we can check how to contribute to otel/mcp to be a standard.

PR #2812, opened 2025-04-12
Draft : MCP open telemetry instrumentation

fali007:
Screenshot 2025-04-12 at 6 39 29 PM

This is a draft PR for adding opentelemetry instrumentation of MCP servers in traceloop.

----

> [!IMPORTANT]
> Add OpenTelemetry instrumentation for MCP servers with new `McpInstrumentor` class and integrate it into the tracing framework.
> 
>   - **Instrumentation**:
>     - Adds `McpInstrumentor` class in `instrumentation.py` to instrument MCP server and client methods `Server._handle_request` and `BaseSession.send_request`.
>     - Integrates `McpInstrumentor` into `tracing.py` with `init_mcp_instrumentor()` function.
>   - **Configuration**:
>     - Adds `.flake8` and `.python-version` files for MCP package.
>     - Updates `pyproject.toml` and `poetry.toml` for MCP package dependencies and settings.
>   - **Documentation**:
>     - Adds `README.md` for MCP instrumentation with installation and usage instructions.
>   - **Misc**:
>     - Updates `instruments.py` to include `MCP` in the `Instruments` enum.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 2caa60b1b4716e137292d8d733f424809363c728. It will automatically update as commits are pushed.</sup>


gyliu513:
Fixed https://github.com/traceloop/openllmetry/issues/2662

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,133 @@
...
+        args[0].root.params = mcp.types.RequestParams()
+        meta = {}
+    else:
+        meta = args[0].root.params.meta
+        if meta is None:
+            meta = {}
+    
+    carrier = None
+    ctx = None
+    if hasattr(args[0].root.params, 'arguments'):
```
nirga, line 110: I'm not a big fan of this as it creates a dependency between the langchain instrumentation and the MCP one (it took me a while to understand that the `__meta__` argument is being injected there and then read here. This is exactly why we have the Python context that OpenTelemetry is already using. What we need is to instrument the MCP tool call method within langchain and set the trace context there. Then, we can just read it it here directly from the context

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,133 @@
...
+        arguments = args[0].root.params.arguments
+        if '__meta__' in arguments:
+            carrier = arguments['__meta__']
+            args[0].root.params.arguments.pop('__meta__')
+    if carrier:    
+        ctx = TraceContextTextMapPropagator().extract(carrier=carrier)
+
+    with tracer.start_as_current_span(f"{method}", context=ctx) as span:
+        span.set_attribute("method", f"{method}")
```
nirga, line 119: We should use semantic convention attributes here

nirga (commented):
Thanks @fali007! I think if we want to get this in we need:
- [ ] Add Tests
- [ ] Fix CI issues (you need to run `poetry lock`)
- [ ] Fix pending comments (mainly the one about the propagation of langchain args to the MCP instrumentation)

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
fali007: I understand that this is unconventional. But I couldn't find any other way (I am also not very happy with this implementation). `StructuredTool` has a coroutine to call below function in `MCP client`. As you can see it only accepts arguments. The original message where I can add trace contexts are created in the MCP Client code. I am sending trace contexts using `__meta__` field inside arguments and removes it in MCP client. If you know any other way to remove this please guide me.
```
async def call_tool(
        self, name: str, arguments: dict[str, Any] | None = None
    ) -> types.CallToolResult:
        """Send a tools/call request."""
        return await self.send_request(
            types.ClientRequest(
                types.CallToolRequest(
                    method="tools/call",
                    params=types.CallToolRequestParams(name=name, arguments=arguments),
                )
            ),
            types.CallToolResult,
        )
```

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,134 @@
...
+    def instrumentation_dependencies(self) -> Collection[str]:
+        return _instruments
+    
+    def _instrument(self, **kwargs):
+        tracer_provider = kwargs.get("tracer_provider")
+        tracer = get_tracer(__name__, __version__, tracer_provider)
+
+        _W(
+            "mcp.server.lowlevel.server",
+            "Server._handle_request",
```
galkleinman, line 30: a bit smelly and prone to be break the fact we're wrapping "private" method. is there any "public" wrapper for it?

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,134 @@
...
+    """Helper for providing tracer for wrapper functions."""
+
+    def _with_tracer(operation_name, tracer):
+        def wrapper(wrapped, instance, args, kwargs):
+            return func(operation_name, tracer, wrapped, instance, args, kwargs)
+        return wrapper
+    return _with_tracer
+
+def serialize(request, depth=0):
+    if depth > 2:
```
galkleinman, line 53:
```suggestion
def serialize(request, depth=0, max_depth=2):
    if depth > max_depth:
```

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,134 @@
...
+        return {}
+    depth += 1
+    def is_serializable(request):
```
galkleinman, line 56: [STYLE] newline above imo for the sake of readability

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -0,0 +1,134 @@
...
+    def _with_tracer(operation_name, tracer):
+        def wrapper(wrapped, instance, args, kwargs):
+            return func(operation_name, tracer, wrapped, instance, args, kwargs)
+        return wrapper
+    return _with_tracer
+
+def serialize(request, depth=0):
+    if depth > 2:
+        return {}
+    depth += 1
```
galkleinman, line 55: it's better to avoid manipulating argument values

galkleinman (commented):
General comments:
1. Please make all the `args[i]` calls failsafe (there are plenty of them)
2. We gotta discuss the context propagation in Langchain, it is quite hacky

Issue #2866, opened 2025-04-28
🐛 Bug Report: (ollama)failed to instrument pre-imported functions


minimAluminiumalism:
### Which component is this bug for?

Ollama Instrumentation

### 📜 Description

When import ollama functions before initializing the traceloop sdk:

```
from ollama import chat

Traceloop.init()
```
The instrumentation doesn't work and no traces or metrics will be sent to endpoint.

But if importing ollama funcs after initialize the traceloop sdk, the instrumentation works.

### 👟 Reproduction steps

Code to reproduce the issue:

```
from traceloop.sdk import Traceloop
from ollama import chat

Traceloop.init(app_name="ollama-instr", api_endpoint="http://localhost:4318")

base_model = "gemma3:1b"

def ollama_chat():
    stream_response = chat(
        model=base_model,
        messages=[
            {
                'role': 'user',
                'content': 'Tell a joke about opentelemetry'
            },
        ],
        stream=True
    )

    full_response = ""
    print("Streaming response:")
    for chunk in stream_response:
        if chunk.message and chunk.message.content:
            print(chunk.message.content, end="", flush=True)
            full_response += chunk.message.content


def main():
    ollama_chat()


if __name__ == "__main__":
    main()
```

### 👍 Expected behavior

The instrumentation works when importing ollama funcs before initialize traceloop configuration.

### 👎 Actual Behavior with Screenshots

The instrumentation not works.

### 🤖 Python Version

Python 3.12.1 (main, Feb 19 2025, 04:15:25) [Clang 19.1.7 ] on darwin

### 📃 Provide any additional context for the Bug.

**Root cause**

Let's see `packages/sample-app/.venv/lib/python3.12/site-packages/ollama/__init__.py`, that func `chat` is `chat = _client.chat`. This binding happens as soon as the module is imported, any downstream code that does

```
from ollama import chat
```
will receive a direct reference to the original `Client.chat` method. Later instrumentation that wrap `Client.chat` 
 have no effect on those already-exported references. 

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
Interesting - how do you suggest solving this?

minimAluminiumalism:
[...]

PR #2871, opened 2025-04-28
fix(ollama): pre-imported funcs instrumentation failure

minimAluminiumalism:

Fixes https://github.com/traceloop/openllmetry/issues/2866

----
**Background**
In many scenarios related to ollama, we write code like this which is invalid for instrumentation.

```
from ollama import chat

Traceloop.init()
```

Looking at `packages/sample-app/.venv/lib/python3.12/site-packages/ollama/__init__.py`, the module exports a `chat` name which is simply an alias for `_client.chat`, established via `chat = _client.chat`. Since this assignment happens during the module's initialization, code using from ollama import chat effectively captures the current reference of `_client.chat` and binds it to the chat name in its own scope. 

```
_client = Client()

generate = _client.generate
chat = _client.chat
embed = _client.embed
```
Later modifications(applying wrappers for instrumentation) to the source method (Client.chat) do not propagate to these pre-existing imported references. The imported chat continues to point to the object it referenced at the time of import.

**Verification**
If we import `chat`(`generate`) only when use it rather than import it immediately at the entry of the file, the instrumentation works fine.

```
import ollama # do not import chat()

Traceloop.init()
...
# instrumentation works
ollama.chat() 
...
```

*While our unit tests(`packages/opentelemetry-instrumentation-ollama/tests/test_chat.py`) follow this pattern, I don't think they're sufficient to catch the pre-imported function instrumentation failure.*

**Solution**

```
for module_name, module in sys.modules.items():
    for attr_name, attr_value in module.__dict__.items():
        if attr_name == "generate" and attr_value.__module__.startswith("ollama"):
            module.__dict__[attr_name] = InstrumentedFunction("generate")
```
Using the above code we scan every loaded module and find any attribute named `chat` (or `generate`, `embeddings`, etc.) whose implementation lives in the ollama package, and replace it with our dynamic proxy. That proxy then delegates to the latest `ollama.generate` under the hood—so no matter how the user imported the function, their calls still get instrumented.

----

> [!IMPORTANT]
> Fixes instrumentation failure for pre-imported Ollama functions by introducing a proxy class and replacing references in other modules.
> 
>   - **Behavior**:
>     - Fixes instrumentation failure for pre-imported functions in Ollama by replacing references in other modules.
>     - Introduces `InstrumentedFunction` proxy class to dynamically call latest Ollama function implementations.
>   - **Instrumentation**:
>     - Modifies `_instrument()` in `__init__.py` to iterate through `sys.modules` and replace references to imported Ollama functions with `InstrumentedFunction`.
>     - Logs replacement actions and handles exceptions with warnings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 883dd00502932961bf32ccd0b5bf5a3501899546. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/opentelemetry-instrumentation-ollama/opentelemetry/instrumentation/ollama/__init__.py
```diff
@@ -477,6 +496,27 @@ def _instrument(self, **kwargs):
                 _wrap(tracer, token_histogram, duration_histogram, wrapped_method),
             )
 
+        # replace imported ollama functions in other modules
+        try:
+            # Iterate through all loaded modules to find modules that reference imported ollama functions
+            for module_name, module in list(sys.modules.items()):
```
nirga, line 502: meh I'm not a huge fan of that - this can be extremely slow and will stall the app at startup - can you elaborate why this needed / how it solves the issue?

minimAluminiumalism: I've added modification details to the PR description.

nirga: Thanks. Can't we wrap the client class directly? (And then it should catch that specific instance as well)

nirga: @minimAluminiumalism not sure if you saw my comment here ⬆️

minimAluminiumalism:
> This logic is going to significantly affect the start time of apps - so I'm super hesitant. Ideally we should be able to instrument some method and then it should be wrapped - no matter when the singleton is created. Have you tried wrapping something else with the "otel-native" method (that we have now) and see if it resolves the issue?

Thanks for the instruction and I will try that :)

minimAluminiumalism:
@nirga I've removed the original logic of iterating and wrapping/instrumenting all loaded modules, and change it to instrument func `Client._request`.
I've also modified the unit tests to cover this scenario.

nirga (approved):

PR #2887, opened 2025-05-05
feat(google-generativeai): implement emitting events in addition to current behaviour

LuizDMM:
### ✅ PR Requirements

- [x] PR name follows conventional commits format: feat(instrumentation): ... or fix(instrumentation): ....

### 📌 Issue Requirements

- [x] Make sure to keep the current way of emitting events and the new (event-based) way with the official semantic
- [x] Add a config to each instrumentation called use_legacy_attributes which defaults to true, and if set to false emit events instead of the prompt / completion attributes.
- [x] Propagate a new initialization parameter of the SDK called use_legacy_attributes that sets this config in all instrumentations.

### DIFF
--- a/packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/__init__.py
+++ b/packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/__init__.py
@@ -3,12 +3,17 @@
 import logging
 import os
 import types
-from typing import Collection, Union
+from typing import Collection
 
 from google.generativeai.types.generation_types import GenerateContentResponse
 from opentelemetry import context as context_api
-from opentelemetry._events import Event, EventLogger, get_event_logger
+from opentelemetry._events import Event, get_event_logger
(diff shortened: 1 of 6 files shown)


packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/span_utils.py
```diff
@@ -0,0 +1,153 @@
...
+    _set_span_attribute(
+        span, SpanAttributes.LLM_PRESENCE_PENALTY, kwargs.get("presence_penalty")
+    )
+    _set_span_attribute(
+        span, SpanAttributes.LLM_FREQUENCY_PENALTY, kwargs.get("frequency_penalty")
+    )
+
+
+@dont_throw
+def set_response_attributes(span, response, llm_model):
```
nirga, line 101: you should check for should_send_prompts here as well

LuizDMM: Done :smile:

nirga (approved):

PR #2889, opened 2025-05-05
feat(langchain): implement emitting events in addition to current behavior

LuizDMM:
### ✅ PR Requirements

- [x] PR name follows conventional commits format: feat(instrumentation): ... or fix(instrumentation):

### 📌 Issue Requirements

- [x] Make sure to keep the current way of emitting events and the new (event-based) way with the official semantic
- [x] Add a config to each instrumentation called use_legacy_attributes which defaults to true, and if set to false emit events instead of the prompt / completion attributes.
- [x] Propagate a new initialization parameter of the SDK called use_legacy_attributes that sets this config in all instrumentations.


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/event_handler.py
ronensc, line 1: Noticed this file is empty. Is it intentional, or should it be removed?

LuizDMM:
> I think you're reverting several changes we've incorporated in `callback_handler`, can you take a look?

Hey @nirga, I'll take a look here in the merges I did and fix if that's the case, sorry for the delay, I hadn't seen your comment.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/event_handler.py
LuizDMM: Hey @ronensc, I'm taking a look in my branch here, if any code was lost in the merges, I'll remove it if it isn't necessary anymore, thank you for point out.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -639,9 +411,13 @@ def on_llm_start(
...
         )
-        _set_llm_request(span, serialized, prompts, kwargs, self.spans[run_id])
+        set_request_params(span, kwargs, self.spans[run_id])
+        if should_emit_events():
+            for prompt in prompts:
+                emit_event(MessageEvent(content=prompt, role="user"))
+        else:
+            set_llm_request(span, serialized, prompts, kwargs, self.spans[run_id])
 
-    @dont_throw
```
ronensc, line 644: @LuizDMM  Could you please clarify the reason for removing `@dont_throw` from `on_llm_end()`?

PR #3015, opened 2025-06-16
feat(milvus) Add semantic convention for Milvus DB metrics

DrishyaDas:


packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -21,6 +21,12 @@ class Meters:
     PINECONE_DB_USAGE_READ_UNITS = "db.pinecone.usage.read_units"
     PINECONE_DB_USAGE_WRITE_UNITS = "db.pinecone.usage_write_units"
 
+    MILVUS_DB_QUERY_DURATION = "db.milvus.query.duration"
```
nirga: why do we need this? this is just the span duration which is already available

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -21,6 +21,12 @@ class Meters:
...
+    MILVUS_DB_SEARCH_DISTANCE = "db.milvus.search.distance"
```
nirga: don't we have a generic metric for that?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -21,6 +21,12 @@ class Meters:
...
+    MILVUS_DB_USAGE_INSERT_UNITS = "db.milvus.usage.insert_units"
```
nirga: can't we make these "generic"?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
DrishyaDas: I understand this is available via spans, but having it as a separate metric helps with aggregation, alerting, and visualizing query performance on a dedicated dashboard.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
DrishyaDas: Agreed, I’m okay with generalizing. I used Milvus-specific names initially since Pinecone only exposes write_units without op-level counts. But since insert/delete are common across DBs, I’ll update the names.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
DrishyaDas: since DB_SEARCH_RESULT_DISTANCE exists in the EventAttributes enum, would you suggest reusing it for the metric as well, or should we define a new value in the Meters enum specifically for metrics?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
nirga: why not use [this](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclientoperationduration)?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
nirga: A new one for metrics - but generic one for all vector DBs

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
DrishyaDas: IMO using `gen_ai.client.operation.duration` would misrepresent the source system and make it harder to differentiate between AI SDK metrics and DB-level metrics. I would suggest to have one generic for vector db as we discussed for other metrics. WDYT?

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
DrishyaDas: Updated to a generic one

DrishyaDas: @nirga  Just a general question. If you check pinecone, it has specific metric names. Are we planning to update it to generic metrics later?

PR #3028, opened 2025-06-18
Fix: Replace Pydantic 'Config' class with model_config for v2 compatibility

samikshaaagarwal:
This PR replaces the deprecated use of `class Config` in `RegistryObjectBaseModel` with `model_config = ConfigDict(...)`, making the code compatible with Pydantic v2.

Confirmed that the deprecation warning no longer appears after running tests.


nirga (commented):
Thanks @samikshaaagarwal! Can you rebase and sign the CLA?

samikshaaagarwal:
Hi @nirga , rebased the branch with the latest `main`, resolved all merge conflicts and signed the CLA. Lemme know if there is anything else you'd like me to change. Thanks! :)

CHANGELOG.md
nirga, line 1: revert?

packages/traceloop-sdk/traceloop/sdk/prompts/model.py
```diff
@@ -2,16 +2,25 @@
 from typing import List, Literal, Optional, Union
 from typing_extensions import Annotated
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ConfigDict
```
duynguyenhoang, line 5: It is not safe to import `ConfigDict` here. It won't work with v1 (Not all minor versions but some)
We already cover it in below import with try/except.

PR #3099, opened 2025-07-09
fix vertexai should_send_prompts

igtm:
I found that `TRACELOOP_TRACE_CONTENT` doesn't work on VertexAI instrument.

```json
{
  "gen_ai": {
    "completion": [
      {
        "content": "...my prompt..",
        "role": "assistant"
      }
    ],
...
```

so i added if statement before setting generation_text.

----

> [!IMPORTANT]
> Adds a conditional check in `_set_response_attributes()` to set content attribute only if `should_send_prompts()` returns true.
> 
>   - **Behavior**:
>     - Adds a conditional check in `_set_response_attributes()` in `__init__.py` to set `LLM_COMPLETIONS.0.content` only if `should_send_prompts()` returns true.
>   - **Functions**:
>     - `should_send_prompts()` checks `TRACELOOP_TRACE_CONTENT` environment variable or context value to determine if prompts should be sent.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for acc61b0407118d0d0b0c43a50f7e10eaa49146a8. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/opentelemetry-instrumentation-vertexai/opentelemetry/instrumentation/vertexai/__init__.py
+++ b/packages/opentelemetry-instrumentation-vertexai/opentelemetry/instrumentation/vertexai/__init__.py
@@ -195,11 +195,12 @@ def _set_response_attributes(span, llm_model, generation_text, token_usage):
         )
 
     _set_span_attribute(span, f"{SpanAttributes.LLM_COMPLETIONS}.0.role", "assistant")
-    _set_span_attribute(
-        span,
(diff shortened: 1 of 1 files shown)


nirga:
Thanks @igtm! It was actually just fixed in #2942 as part of a bigger change we're introducing to the instrumentation :)

PR #3129, opened 2025-07-14
fix(traceloop-sdk): redefine histogram bucket boundaries

minimAluminiumalism:

Removed LLM_OPERATION_DURATION to make it use OTel SDK's boundary defaultly. For token we're utilizing the definitions at [python-contrib](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/109173fec7bfe39fd1c05a0e726e98643e37b55a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/instruments.py#L35).


packages/traceloop-sdk/traceloop/sdk/metrics/metrics.py
```diff
@@ -88,27 +88,6 @@ def metric_views() -> Sequence[View]:
     return [
         View(
             instrument_name=Meters.LLM_TOKEN_USAGE,
-            aggregation=ExplicitBucketHistogramAggregation(
```
nirga, line 91: Should we define a different aggregation?

minimAluminiumalism: No. Since we have explicitly specified the boundary values, we should just use `ExplicitBucketHistogramAggregation` rather than `ExponentialBucketHistogramAggregation` or `DefaultAggregation` here.

https://github.com/open-telemetry/opentelemetry-python/blob/14401e1dd82852fdea76ecf4d720fbd74834b705/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/aggregation.py#L4

PR #3150, opened 2025-07-18
fix(openai_agents):Added the attributes to the duration histogram

adharshctr:


packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/__init__.py
```diff
@@ -192,6 +192,8 @@ async def _wrap_agent_run_streamed(
                     duration,
                     attributes={
                         "gen_ai.agent.name": agent_name,
+                        SpanAttributes.LLM_SYSTEM: "openai",
```
nirga, line 195: @adharshctr shouldn't this be the vendor (which may not always be OpenAI)?

PR #3199, opened 2025-07-28
fix(instrumentation): MCP Instrumentation: streamablehttp_client Parameter Corruption

feng-95:

image

When using the MCP Instrumentation to instrument the mcp streamablehttp_client, the parameters returned by the instrumentation are incorrect. The streamablehttp_client returns 
```python
     read_stream,
     write_stream,
     transport.get_session_id,
```
https://github.com/modelcontextprotocol/python-sdk/blob/v1.9.0/src/mcp/client/streamable_http.py#L486
But the instrumentated function only returns `read_stream, write_stream`.

### DIFF
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
@@ -130,11 +130,21 @@ class McpInstrumentor(BaseInstrumentor):
             async with wrapped(*args, **kwargs) as result:
                 try:
                     read_stream, write_stream = result
+                    yield InstrumentedStreamReader(
+                        read_stream, tracer
+                    ), InstrumentedStreamWriter(write_stream, tracer)
                 except ValueError:
-                    read_stream, write_stream, _ = result
-                yield InstrumentedStreamReader(
-                    read_stream, tracer
-                ), InstrumentedStreamWriter(write_stream, tracer)
+                    try:
+                        read_stream, write_stream, get_session_id_callback = result
+                        yield InstrumentedStreamReader(
+                            read_stream, tracer
+                        ), InstrumentedStreamWriter(write_stream, tracer), get_session_id_callback
+                    except Exception as e:
+                        logging.warning(f"mcp instrumentation _transport_wrapper exception: {e}")
+                        yield result
+                except Exception as e:
+                    logging.warning(f"mcp instrumentation transport_wrapper exception: {e}")
+                    yield result
 
         return traced_method
 


feng-95:
@fali007 @nirga Could you please take a look to this pr? The instrumented streamable-http mcp-client fails to initialize, , throwing 'ValueError: not enough values to unpack (expected 3, got 2)' errors during startup.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -130,11 +130,21 @@ async def traced_method(
             async with wrapped(*args, **kwargs) as result:
                 try:
                     read_stream, write_stream = result
+                    yield InstrumentedStreamReader(
```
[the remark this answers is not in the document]
nirga: I think it's a valid point, no? @feng-95

nirga:
@feng-95 can you look into the comment from Elipsis?

feng-95:
> @feng-95 can you look into the comment from Elipsis?

@nirga Updated the type annotation.

nirga (commented):
Why can't you yield the same thing? @feng-95

feng-95:
> Why can't you yield the same thing? @feng-95

Since `stdio_client` and `sse_client` yield (read_stream, write_stream), while `streamablehttp_client` yield (read_stream, write_stream, get_session_id), we need to conditionally yield different responses to support a unified wrapper for all these functions.

I'm unsure whether creating a new wrapper function specifically for `streamablehttp_client` is the better approach? 
@nirga

feng-95:
Hi @nirga , could you please take a look to this again? Would you think it will be better to creating a new wrapper function specifically for `streamablehttp_client`?

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -125,16 +125,30 @@ def _transport_wrapper(self, tracer):
...
+            Union[
+                Tuple[InstrumentedStreamReader, InstrumentedStreamWriter],
+                Tuple[InstrumentedStreamReader, InstrumentedStreamWriter, Any]
+            ],
+            None
         ]:
             async with wrapped(*args, **kwargs) as result:
                 try:
                     read_stream, write_stream = result
+                    yield InstrumentedStreamReader(
```
nirga, line 136: @feng-95 - shouldn't you yield the same thing that you got from `wrapped`?

feng-95: No, we need to yield InstrumentedClient to make the instrument work. This is the same logic with the first version of mcp instrumentor

PR #3290, opened 2025-08-17
fix(openai): improve LiteLLM LegacyAPIResponse handling for streaming

nirga:
## Summary

- Fixes streaming response handling for LiteLLM's LegacyAPIResponse objects used by libraries like strands
- Improves architecture by handling LegacyAPIResponse parsing at wrapper creation time instead of during iteration
- Prevents mutation of original stream objects to maintain integrity across multiple consumers
- Enhances error handling and type detection robustness

## Key Changes

- **Better LegacyAPIResponse handling**: Parse at wrapper creation time, not during iteration
- **Stream integrity**: No longer mutate original response objects
- **Robust type detection**: Use duck typing instead of fragile string matching  
- **Enhanced error handling**: Proper logging with graceful fallback behavior
- **Comprehensive test coverage**: Added tests for stream integrity, multiple consumers, and edge cases

## Test Plan

- [x] All existing OpenAI instrumentation tests pass (167/167)
- [x] New LegacyAPIResponse tests validate stream integrity
- [x] Real-world validation with strands library example
- [x] Error handling scenarios covered
- [x] Multiple consumer safety verified


nirga:
## Review Comment Responses

Thanks for the thorough review! I've addressed the key issues in commit 487ea034:

### ✅ Fixed Issues:

**1. Async Context Manager (__aexit__)** - Fixed to ensure cleanup runs and properly handle exceptions like __exit__ does

**2. Missing Parse Success Flag** - Added `_response_was_parsed = True` to both sync and async paths when LegacyAPIResponse parsing succeeds. This was incomplete logic - the flag was checked in `_record_partial_metrics()` but never set anywhere.

### 📝 Design Decisions (No Changes):

**3. Deep Copy vs Shallow Copy** - Keeping shallow copy because `_accumulate_stream_items()` creates fresh nested objects rather than modifying existing ones, so deep copy would add unnecessary overhead.

**4. Duck Typing for LegacyAPIResponse** - Keeping the duck typing approach (`hasattr(obj, 'parse') and callable(obj.parse)`) because it's more robust across LiteLLM versions and handles edge cases gracefully. The function is private and used in controlled contexts.

The core functionality now works correctly with proper span status and cleanup behavior for LiteLLM streaming responses.

PR #3367, opened 2025-09-07
feat(langchain): allow configuration of metadata key prefix

ianchi:
Allowing to configure the key prefix for metadata enables use on other backends.
As an example for Langfuse v3 otel backend, you can use prefix `langfuse.metadata` to make user.id and session.id available.

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
@@ -37,10 +37,24 @@ class LangchainInstrumentor(BaseInstrumentor):
         exception_logger=None,
         disable_trace_context_propagation=False,
         use_legacy_attributes: bool = True,
+        metadata_key_prefix: Optional[str] = None
     ):
+        """Create a Langchain instrumentor instance.
+        
+        Args:
+            exception_logger: A callable that takes an Exception as input. This will be
+                used to log exceptions that occur during instrumentation. If None, exceptions will not be logged.
+            disable_trace_context_propagation: If True, disables trace context propagation to LLM providers.
+            use_legacy_attributes: If True, uses span attributes for Inputs/Outputs instead of events.
+            metadata_key_prefix: Prefix for metadata keys added to spans. If None, defaults to
+                `SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES`.
+                Useful for using with other backends.
+        """
         super().__init__()
         Config.exception_logger = exception_logger
         Config.use_legacy_attributes = use_legacy_attributes
+        if metadata_key_prefix:
+            Config.metadata_key_prefix = metadata_key_prefix
         self.disable_trace_context_propagation = disable_trace_context_propagation
 
     def instrumentation_dependencies(self) -> Collection[str]:
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
@@ -290,7 +290,7 @@ class TraceloopCallbackHandler(BaseCallbackHandler):
             for key, value in sanitized_metadata.items():
                 _set_span_attribute(
                     span,
-                    f"{SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES}.{key}",
+                    f"{Config.metadata_key_prefix}.{key}",
                     value,
                 )
 
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/config.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/config.py
@@ -1,9 +1,10 @@
 from typing import Optional
 
 from opentelemetry._events import EventLogger
-
+from opentelemetry.semconv_ai import SpanAttributes
 
 class Config:
     exception_logger = None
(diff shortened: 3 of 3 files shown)


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -37,10 +37,24 @@ def __init__(
         exception_logger=None,
         disable_trace_context_propagation=False,
         use_legacy_attributes: bool = True,
+        metadata_key_prefix: Optional[str] = None
```
nirga: this should default to the way it works tdoay

ianchi: It defaulted from Config, but I also added here for clarity

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -37,10 +37,24 @@ def __init__(
...
     ):
```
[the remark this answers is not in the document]
ianchi: fixed

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -290,7 +290,7 @@ def _create_span(
             for key, value in sanitized_metadata.items():
                 _set_span_attribute(
                     span,
-                    f"{SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES}.{key}",
+                    f"{Config.metadata_key_prefix}.{key}",
```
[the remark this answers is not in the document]
ianchi: fixed

ianchi:
Hi @nirga, could you please review the updated PR.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -37,10 +36,24 @@ def __init__(
...
+        metadata_key_prefix: Optional[str] = SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES
     ):
+        """Create a Langchain instrumentor instance.
+        
+        Args:
+            exception_logger: A callable that takes an Exception as input. This will be
+                used to log exceptions that occur during instrumentation. If None, exceptions will not be logged.
+            disable_trace_context_propagation: If True, disables trace context propagation to LLM providers.
+            use_legacy_attributes: If True, uses span attributes for Inputs/Outputs instead of events.
+            metadata_key_prefix: Prefix for metadata keys added to spans. If None, defaults to
```
nirga: Nit: this is inaccurate - only if not set the default kicks in. If it's None then it's None

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -37,10 +36,24 @@ def __init__(
         exception_logger=None,
         disable_trace_context_propagation=False,
         use_legacy_attributes: bool = True,
+        metadata_key_prefix: Optional[str] = SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES
```
nirga: Type shouldn't be optional - it always has a value

nirga (commented):
Thanks @ianchi - left a small comment

ianchi:
Hi @nirga, updated

ianchi:
Hi @nirga fixed linting (tested locally) and rebased to fix merge conflicts

ianchi:
Hi @nirga, I added a fix on an invalid semconv attribute.
this bug was introduced a couple of weeks ago when replacing SpanAttributes strings to otel semconv.

This one does not exist upstream `GenAIAttributes.GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS`

It errors and completely  breaks logging of output

nirga (commented):
Thanks @ianchi - CI seems to fail. Can you take a look?

ianchi:
Hi @nirga, it seems that when the faulty attribute was introduced, some test must have failed (indicating the problem), but they where changed to ignore it.
Now that the behaviour is correct again the tampered tests fail.

I've tried to revert them all, but I'm having trouble running them locally.
Can you run the pipeline again?

ianchi:
@nirga  I see that commit d9a0d7d00561ae920190cf3cd894491bdac036c1 tried to address part of the problem, but it introduced yet another bug. It is using:

`SpanAttributes.GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS,`

but this does not exist in SpanAttributes.

In this PR I rebased and fixed it for langchain, but that PR also introduces the bug in Anthropic

nirga:
@ianchi it does exist - https://github.com/traceloop/openllmetry/blob/589e1047edf979f5a0b107f4916b72c014a282d7/packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py#L66

I've fixed the tests anyway, let me take another look on the PR to verify and we're good to go

ianchi:
Hi @nirga 
Regarding this:
>it does exist 

I see it in the codebase, but can it be that the change was never released as a new package? when you install the langchain instrumentation locally it pulls version 0.4.13 of semconv_ai an it does not have those attributes.

nirga:
Might be the issue indeed @ianchi - I'll investigate. This should work now in any case

PR #3138, opened 2025-07-15
feat(instrumentation): updated GenAI attributes to use OTel's

martimfasantos:
This pull request updates the token usage terminology across multiple packages and test files. The changes involve replacing `prompt_tokens` with `input_tokens` and `completion_tokens` with `output_tokens` to align with updated naming conventions present in the [Documentation](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#deprecated-genai-attributes). Additionally, all related assertions in the tests have been modified to reflect these updates.

### Screenshots
image

---

- [x] I have added tests that cover my changes.
- [ ] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change.
- [x] PR name follows conventional commits format: `feat(instrumentation): ...` or `fix(instrumentation): ...`.
- [ ] (If applicable) I have updated the documentation accordingly.

----

> [!IMPORTANT]
> Renamed `prompt_tokens` to `input_tokens` and `completion_tokens` to `output_tokens` across various test files and packages to align with updated naming conventions.
> 
>   - **Terminology Update**:
>     - Renamed `prompt_tokens` to `input_tokens` and `completion_tokens` to `output_tokens` in `SpanAttributes` in `__init__.py`.
>     - Updated assertions in test files like `test_completion.py`, `test_messages.py`, and `test_prompt_caching.py` to use `input_tokens` and `output_tokens`.
>   - **Test Files**:
>     - Modified token usage assertions in `test_completion.py` for AlephAlpha, Anthropic, OpenAI, and Together packages.
>     - Updated `test_prompt_caching.py` for Anthropic and OpenAI to reflect new token terminology.
>     - Adjusted `test_chat.py` and `test_completion.py` for Together package to use updated token names.
>   - **Misc**:
>     - Ensured all related test assertions and logs reflect the new token terminology.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 3dad6f9895709d81dfc4c2fb58d718fa639f648f. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified GenAI attributes surface for telemetry and streaming first-token metrics.

* **Refactor**
[...]

PR #3465, opened 2025-11-25
fix(langchain): incorrect role on generations

ianchi:
This fixes an error introduced by #3138 that records the wrong `role` it uses the generation type instead of the message type.
The generation type is only for serialization and is NOT "ai"/"human".

----

> [!IMPORTANT]
> Fixes incorrect role assignment in `set_chat_response` by using message type instead of generation type in `span_utils.py`.
> 
>   - **Bug Fix**:
>     - Corrects role assignment in `set_chat_response` in `span_utils.py` by using `generation.message.type` instead of `generation.type`.
>     - Ensures role is accurately set to "assistant" when `generation.message` is not present.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 2a6c479bfee5374fe1a73091e58ec47b4019e665. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat-response span role assignment now derives from the message type when present, falls back to "assistant" for non-message generations, and no longer overwrites role when tool calls exist — improving role accuracy and attribution.

* **Tests**
  * Added tests verifying role assignment, default assistant fallback, content storage, and recording of tool calls (including function_call and parsed tool-call arguments) with trace-content enabled.

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
@@ -269,7 +269,10 @@ def set_chat_response(span: Span, response: LLMResult) -> None:
                 )
                 if tool_calls and isinstance(tool_calls, list):
                     _set_chat_tool_calls(span, prefix, tool_calls)
-
+            else:
+                _set_span_attribute(
+                    span, f"{prefix}.role", "assistant"
+                )
             i += 1
 
 
--- a/packages/opentelemetry-instrumentation-langchain/tests/test_span_utils.py
+++ b/packages/opentelemetry-instrumentation-langchain/tests/test_span_utils.py
@@ -82,5 +82,5 @@ def test_plain_generation_defaults_to_assistant_role():
     set_chat_response(span, result)
 
     prefix = f"{GenAIAttributes.GEN_AI_COMPLETION}.0"
-    assert prefix + ".role" not in span.attributes
+    assert span.attributes[prefix + ".role"] == "assistant"
     assert span.attributes[prefix + ".content"] == "plain completion"


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
```diff
@@ -268,12 +263,19 @@ def set_chat_response(span: Span, response: LLMResult) -> None:
                     else generation.message.additional_kwargs.get("tool_calls")
                 )
                 if tool_calls and isinstance(tool_calls, list):
-                    _set_span_attribute(
-                        span,
-                        f"{prefix}.role",
-                        "assistant",
```
nirga, line 274: this is incorrect - it should be tool

ianchi: That "else" is the default role for an llm, when there is no "message" and so no direct way to extract `message.type`. A legacy non-chat call.

But even in that case, this represents the response of an LLM. I believe the default for that is still "assistant".

In fact for v1 Langchain "all" model responses return AIMessage. So in practice there will not be any other role but "assistant" even if using `message.type`

nirga: yes but look at the OpenAI instrumentation - we differentiate between tool call responses and textual ai responses

ianchi: In that case it directly does not include the "role" at all, as far as I can see, only `content` and `finish_reason`.
Do you prefer that I directly remove all the "else" part, so it does not add the role?

nirga: I still don't fully understand why we need this change. Do you have an example of a faulty span?

ianchi: yes, any llm generation without a tool calls currently gets a role of `unknown`

image

```python

            _set_span_attribute(
                span,
                f"{prefix}.role",
                _message_type_to_role(generation.type),
            )
            
 ```
 
 The problem is that the current code tries to get the role from `generation.type`
 
But the type of generations are:

```python
class Generation(Serializable):
    # .....
    type: Literal["Generation"] = "Generation"
    """Type is used exclusively for serialization purposes.
    Set to "Generation" for this class."""
    
 class ChatGeneration(Generation):
     # .....
     type: Literal["ChatGeneration"] = "ChatGeneration"  # type: ignore[assignment]
    """Type is used exclusively for serialization purposes."""

 ```
 
 It is deffinetly not possible to derive the `role` from there, and you end up with `unknown`. It should be done from generation.**message**.type
 
Even though the use of `generation.type` comes from before that PR, that part of the code was never hit, and so you always got a fixed "assistant", and the problem was not visible. But the change introduced by #3138 made it surface.

**Comment by ianchi** [comment]
Hi @nirga I added a test and aligned the "completions" caso to OpenAI, (no role at all).

**Comment by nirga** [comment]
@ianchi it should always be with a role. Can you walk me through what are you looking to fix?

**Comment by ianchi** [comment]
I added a detailed explanation of the error, with printscreen and code snippets, in this [comment](https://github.com/traceloop/openllmetry/pull/3465#discussion_r2560800638).
Basically, right now any llm response "without" a tool_calls is logged in a span with `.role = unknown` instead of "assistant".
This is so because current code tries to extract the "role" from the wrong object. (it uses `generation.type` instead of `generation.message.type`).

Please let me know if any additional info is required.

Regarding this:
> it should always be with a role.

As far as I undertand the openai implementation, in the case of a plain "completion" (not a chat completion) does **NOT** add a "role" to the span.
Look at `opentelemetry.instrumentation.openai.shared.completion_wrappers` the contend is added in `_handle_response -> _set_completions` which only adds attributes `gen_ai.completion.{i}.finish_reason` and `.content` it nevers adds a `.role` attribute.

In the case of "chat completions" the `_set_completions` does add a `.role`, extracting it from the message.

From your [previous comment](https://github.com/traceloop/openllmetry/pull/3465#discussion_r2560494656):
> yes but look at the OpenAI instrumentation - we differentiate between tool call responses and textual ai responses

 I understood that you wanted to mirror this behaviour in the Langchain instrumentation also. And thus I also modified the "textual response" case.
 But if you prefer it to return role="assistant", I can add it.

**Comment by ianchi** [comment]
Hi @nirga, any comments on this?
Let me know if anything else is needed

**Comment by ianchi** [comment]
@nirga  Friendly ping.

We are still getting all responses as `role=unkown`.
We need a quick fix for this.

Thanks

File: packages/opentelemetry-instrumentation-langchain/tests/test_span_utils.py [hunk 2]
```diff
@@ -0,0 +1,86 @@
...
+
+def test_plain_generation_defaults_to_assistant_role():
+    span = _DummySpan()
+    generation = Generation(text="plain completion")
+    result = LLMResult(generations=[[generation]])
+
+    set_chat_response(span, result)
+
+    prefix = f"{GenAIAttributes.GEN_AI_COMPLETION}.0"
+    assert prefix + ".role" not in span.attributes
```
nirga: it should always have a role - it should be "assistant" in this case

ianchi: Sorry @nirga, I'm a bit confuse.

For the case of *plain* completions (not chat generations) you also want to add a "role" even if it does not actually exist in any of the underlying entities?

In some prior message you mentioned that I needed to follow the openai instrumentation example.
In that one, if I look into `packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/completion_wrappers.py` and `packages/opentelemetry-instrumentation-openai/tests/traces/test_completions.py` there is no role at all.

That is why in the *plain complations* case I dind't add the role and tested accordingly.

If I understand correctly your last comment, I should modify the behaviour (and test) to also add a `.role = assistant` in the completions case.

Please confirm

ianchi:
Hi @nirga 
as requested I added back the "role=assistant" for plain complations.

ianchi:
Hi  @nirga, any chance to fix this issue?

PR #3487, opened 2025-12-02
fix(openai-agents): apply content tracing flag to content

duanyutong:
## Changes

* should_send_prompts is defined in `openai-agents` instrumentation but unused; use it to control content logging in a way similar to the `openai` package
* make `should_send_prompts` implementations in the two packages consistent (they had different behaviours)
* apply `_is_truthy` to both the env var and the trace context override variable for consistency and maximum compatibility. 

----

> [!IMPORTANT]
> Enable content tracing in `openai-agents` by using `should_send_prompts()` and ensure consistency with `openai` package.
> 
>   - **Behavior**:
>     - Use `should_send_prompts()` in `_hooks.py` to control content logging for OpenAI Agents.
>     - Apply `_is_truthy` to both environment variable and trace context override in `should_send_prompts()` for consistency.
>   - **Consistency**:
>     - Align `should_send_prompts()` behavior in `openai_agents/utils.py` and `openai/utils.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 77fa6f41de0ec8d21f177311cbb62de711ef2253. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Configurable content tracing added: users can now enable/disable inclusion of prompts, responses, tool arguments, and realtime outputs in traces.

* **Behavior Changes**
  * Tracing now respects the content-tracing setting to prevent accidental capture of prompt/response content while preserving public API compatibility.

* **Documentation**
  * CONTRIBUTING.md updated with detailed local testing, linting, and per-package development instructions.

### DIFF
--- a/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
+++ b/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
@@ -45,7 +45,7 @@ except ImportError:
     SpeechGroupSpanData = None
 
 
-def _extract_prompt_attributes(otel_span, input_data):
+def _extract_prompt_attributes(otel_span, input_data, trace_content: bool):
     """
     Extract prompt/input data from messages and set them as span attributes.
 
@@ -101,10 +101,15 @@ def _extract_prompt_attributes(otel_span, input_data):
                     {
                         "id": msg.get("id", ""),
                         "name": msg.get("name", ""),
-                        "arguments": msg.get("arguments", ""),
-                    }
+                    } | (
+                        {"arguments": msg.get("arguments", "")}
+                        if trace_content else {}
+                    )
                 ]
-            elif msg_type == "function_call_output":
+            elif (
+                msg_type == "function_call_output"
+                and trace_content
+            ):
                 # Tool outputs are tool messages
                 role = "tool"
                 content = msg.get("output")
@@ -115,7 +120,7 @@ def _extract_prompt_attributes(otel_span, input_data):
             otel_span.set_attribute(f"{prefix}.role", role)
 
         # Set content attribute
-        if content is not None:
+        if content is not None and trace_content:
             if not isinstance(content, str):
                 content = json.dumps(content)
             otel_span.set_attribute(f"{prefix}.content", content)
@@ -162,14 +167,14 @@ def _extract_prompt_attributes(otel_span, input_data):
                     otel_span.set_attribute(
                         f"{prefix}.tool_calls.{j}.name", tool_call["name"]
                     )
-                if tool_call.get("arguments"):
+                if tool_call.get("arguments") and trace_content:
                     args = tool_call["arguments"]
                     if not isinstance(args, str):
                         args = json.dumps(args)
                     otel_span.set_attribute(f"{prefix}.tool_calls.{j}.arguments", args)
 
 
-def _extract_response_attributes(otel_span, response):
+def _extract_response_attributes(otel_span, response, trace_content: bool):
     """
     Extract model settings, completions, and usage from a response object
     and set them as span attributes.
@@ -214,7 +219,7 @@ def _extract_response_attributes(otel_span, response):
     # Extract completions from response.output
     if hasattr(response, "output") and response.output:
         for i, output in enumerate(response.output):
-            if hasattr(output, "content") and output.content:
+            if hasattr(output, "content") and output.content and trace_content:
                 # Text message with content array (ResponseOutputMessage)
                 content_text = ""
                 for content_item in output.content:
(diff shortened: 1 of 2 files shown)


packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/utils.py
```diff
@@ -174,10 +174,18 @@ def run_async(method):
         asyncio.run(method)
 
 
-def should_send_prompts():
-    return (
-        os.getenv(TRACELOOP_TRACE_CONTENT) or "true"
-    ).lower() == "true" or context_api.get_value("override_enable_content_tracing")
+def _is_truthy(value):
+    return str(value).strip().lower() in ("true", "1", "yes", "on")
```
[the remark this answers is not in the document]
duanyutong: Each package has to remain independent I think.

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/utils.py
```diff
@@ -18,10 +20,14 @@ def _is_truthy(value):
...
-    env_setting = os.getenv("TRACELOOP_TRACE_CONTENT", "true")
+def should_send_prompts() -> bool:
+    """Determine if LLM content tracing should be enabled.
+
+    Content includes not only prompts, but also responses.
+    """
+    env_setting = os.getenv(_TRACELOOP_TRACE_CONTENT, "true")
     override = context_api.get_value("override_enable_content_tracing")
-    return _is_truthy(env_setting) or bool(override)
+    return _is_truthy(env_setting) or _is_truthy(override)
```
nirga, line 33: why did you change this?

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/utils.py
```diff
@@ -174,10 +174,18 @@ def run_async(method):
...
+
+
+def should_send_prompts() -> bool:
+    """Determine if LLM content tracing should be enabled.
+
+    Content includes not only prompts, but also responses.
+    """
+    env_setting = os.getenv(TRACELOOP_TRACE_CONTENT, "true")
+    override = context_api.get_value("override_enable_content_tracing")
+    return _is_truthy(env_setting) or _is_truthy(override)
```
nirga, line 188: same here - we set the override flag so it will always be a boolean

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/utils.py
duanyutong: `context_api` is not typed. This flag isn't documented and users could set it to any value `1`, `True`, `"true"`, `"yes"`, `"on"` etc. Having a single implementation of `_is_truthy` reduces surprises and provides better compatibility with a range of possible values. The rationale is the same as that for the environment variable.

packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/utils.py
duanyutong: > we set the override

I see that `packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py` sets it, but this flag is also needed by general users to apply fine control over content tracing at the per-request level, when they are using simple otel without `traceloop-sdk` necessarily.

The `_is_truthy` call here adds extra safety, without breaking any existing behaviour, and makes the parsing consistent between the context value and the environment variable as the two are equivalent and only differ in precedence.

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
```diff
@@ -613,7 +620,12 @@ def on_span_end(self, span):
 
                 # Add function/tool specifications to the request using OpenAI semantic conventions
                 response = getattr(span_data, "response", None)
-                if response and hasattr(response, "tools") and response.tools:
+                if (
+                    response
+                    and hasattr(response, "tools")
+                    and response.tools
+                    and trace_content
+                ):
```
galkleinman, line 633: I wonder if there's real data in here or just method signatures and stuff like that which might be ok'ish to report even when trace_content is disabled

duanyutong: Good question. Looking at this more closely it seems to be just the signature and not the actual tool call arguments. Removed the condition from this block. Thanks!

duanyutong: I also added this condition to a couple of more places per comments by the bot.

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
```diff
@@ -649,23 +666,24 @@ def on_span_end(self, span):
...
                 if response:
-                    model_settings = _extract_response_attributes(otel_span, response)
+                    model_settings = _extract_response_attributes(otel_span, response, trace_content)
                     self._last_model_settings = model_settings
 
             elif (
                 _has_realtime_spans
                 and SpeechSpanData
                 and isinstance(span_data, SpeechSpanData)
+                and trace_content
```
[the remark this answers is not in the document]
duanyutong: cc @galkleinman this comment seems unrelated to my change in this particular PR. I'll defer to you on that.

PR #3646, opened 2026-01-30
fix: Change attributes datasets and experiment on client to lazyloading and mark datasets as deprecated

tassadarius:
- [ ] ~~If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change~~.
- [ ] ~~(If applicable) I have updated the documentation accordingly.~~

Changes
-------
- Import and load attributes `datasets` and `experimental` only once when actually requested
- Mark attribute `datasets` on Client as deprecated

Reasoning
---------
Calling Traceloop.init() allocates ~35 MB of memory when pandas is
installed, either by the attributes `datasets` or `experiment`. Both of these attributes seem
not relevant for production

People interested in the Datasets class should IMO just create an
instance on their own.

Since it's already released I propose a deprecation process.

**In general I think it is a bad idea to expose and initialize all these internal attributes directly on the public main class.**
 
Here you can see the memory usages done with [memray](https://github.com/bloomberg/memray) of this simple execution:

```bash
TRACELOOP_API_KEY=asdf uv  run --active memray run -c "from traceloop.sdk import Traceloop;Traceloop.init(app_name='joke_generation_service')
```

|Code| Mem-usage | Link to report |
|------|------|--------|
| Current main | 70.1 MB | [memray-flamegraph-before-patch.html](https://github.com/user-attachments/files/24965425/memray-flamegraph-before-patch.html)  |
| Patch | 37.6 MB | [memray-flamegraph-with-patch.html](https://github.com/user-attachments/files/24965438/memray-flamegraph-with-patch.html) |

----

> [!IMPORTANT]
> Optimize `Client` memory usage by lazy-loading `datasets` and `experiment` attributes, marking `datasets` as deprecated.
> 
>   - **Behavior**:
>     - `datasets` and `experiment` attributes in `Client` are now lazy-loaded, reducing initial memory usage.
>     - `datasets` attribute is marked as deprecated with a warning, suggesting users create their own instances.
>   - **Tests**:
>     - Added `test_client_lazy_loads_datasets()` and `test_client_lazy_loads_experiment()` in `test_client.py` to verify lazy loading.
>     - Added `test_datasets_deprecation_warnings()` to check for deprecation warnings.
>   - **Misc**:
>     - Removed direct imports of `Datasets` and `Experiment` in `client.py` and moved them inside the properties.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 2f53649d563f152b5c032c71b6e48553f5bce00c. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **Deprecations**
  * The `datasets` property on the client is deprecated, including its setter and deleter; it will be removed in a future release.

* **Changes**
  * The `experiment` property now initializes on first access instead of during client creation.

* **Tests**
  * Added tests verifying lazy-loading behavior for `experiment` and `datasets`, caching/idempotence, and that deprecation warnings are emitted for `datasets`.

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/client/client.py
+++ b/packages/traceloop-sdk/traceloop/sdk/client/client.py
@@ -3,7 +3,6 @@ import os
 from deprecated import deprecated
 
 from traceloop.sdk.annotation.user_feedback import UserFeedback
-from traceloop.sdk.experiment.experiment import Experiment
 from traceloop.sdk.client.http import HTTPClient
 from traceloop.sdk.version import __version__
 from traceloop.sdk.associations.associations import Associations
@@ -25,7 +24,6 @@ class Client:
     api_endpoint: str
     api_key: str
     user_feedback: UserFeedback
-    experiment: Experiment
     associations: Associations
     guardrails: Guardrails
     _http: HTTPClient
@@ -65,15 +63,22 @@ class Client:
         )
         self.user_feedback = UserFeedback(self._http, self.app_name)
         self._datasets = None
-        experiment_slug = os.getenv("TRACELOOP_EXP_SLUG")
+        self._experiment = None
         # TODO: Fix type - Experiment constructor should accept Optional[str]
-        self.experiment = Experiment(self._http, self._async_http, experiment_slug)  # type: ignore[arg-type]
         self.associations = Associations()
         self.guardrails = Guardrails(self._async_http)
 
     @property
-    @deprecated(
+    def experiment(self):
+        # Lazy load only if accessed.
+        if self._experiment is None:
+            from traceloop.sdk.experiment.experiment import Experiment
+            experiment_slug = os.getenv("TRACELOOP_EXP_SLUG")
+            self._experiment = Experiment(self._http, self._async_http, experiment_slug)  # type: ignore[arg-type]
+        return self._experiment
 
+    @property
+    @deprecated(
         reason="datasets as client attribute is deprecated. Use a dedicated instance "
                "of Datasets from traceloop.sdk.datasets"
     )
--- a/packages/traceloop-sdk/tests/test_client.py
+++ b/packages/traceloop-sdk/tests/test_client.py
@@ -3,6 +3,7 @@ from traceloop.sdk.client import Client
 from traceloop.sdk.client.http import HTTPClient
 from traceloop.sdk.annotation.user_feedback import UserFeedback
 from traceloop.sdk.datasets.datasets import Datasets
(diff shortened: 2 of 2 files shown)


packages/traceloop-sdk/traceloop/sdk/client/client.py
```diff
@@ -65,9 +62,45 @@ def __init__(
...
+            experiment_slug = os.getenv("TRACELOOP_EXP_SLUG")
+            self._experiment = Experiment(self._http, self._async_http, experiment_slug)  # type: ignore[arg-type]
+        return self._experiment
+
+    @property
+    @deprecated(
+        reason="datasets as client attribute is deprecated. Use a dedicated instance "
+               "of Datasets from traceloop.sdk.datasets"
+    )
+    def datasets(self):
```
nina-kollman, line 85: Can you pls explain why you need this change ?

tassadarius: As laid out already in my initial PR [description](https://github.com/traceloop/openllmetry/pull/3646#issue-3875616042) (see section `Reasoning`)

You let everybody who has pandas installed pay a 35 MB tax, when loading this library, where it is not clear why the experimental and datasets should be loaded alongside the main client.

If it is experimental code, it should not be loaded automatically into the environment of consumers of the library.

nina-kollman: Thanks for the detailed explanation @tassadarius , this is a valid concern.
We're currently working on some architectural changes to the SDK, and as part of that effort, we're planning to introduce package extras to properly address this.
This is on our roadmap but will take some time to implement properly without breaking existing users.
In the meantime, if this is blocking you, you could apply a local patch to defer the import — for example, by setting an environment variable check before the pandas-dependent imports in your local installation.

tassadarius: @nina-kollman Patching dependencies of dozens of our production servers in a pre-deployment-hook is definitely too cumbersome for us.

I would kindly ask you to be a little bit more mindful about introducing experimental code, and especially 3rd party dependencies into official releases and loading them into memory.

How do you now plan to proceed with this `experimental` and `datasets` attributes? IMO they still have nothing to do with the Client class and should get taken out from there.

PR #3649, opened 2026-02-01
feat(traceloop-sdk): Add guardrails

nina-kollman:
Fixes: TLP-1692
Fixes: TLP-1691
Fixes: TLP-1438

## Summary by CodeRabbit

- **New Features**
  - Revamped Guardrails system: built-in guards, conditions, on-failure handlers, default input mapping, and a guardrail decorator for sync/async functions.
  - New client API to create guardrails and a standalone evaluator "run" interface.
  - New interactive/sample guardrail examples demonstrating evaluators, custom functions, multi-guard flows, and validation.

- **Refactor / Deprecation**
  - Legacy guardrails module removed and marked deprecated; older demo scripts deleted.

- **Tests**
  - Extensive unit and integration tests with recorded API interactions.

- **Chores**
  - Pytest asyncio mode configured.

### DIFF
--- a/packages/sample-app/sample_app/guardrails/custom_function_guard.py
+++ b/packages/sample-app/sample_app/guardrails/custom_function_guard.py
@@ -52,9 +52,8 @@ async def simple_lambda_guard_example():
 
     try:
         guardrail = Guardrails(
-            lambda z: z["word_count"] < 100,
-            on_failure="raise",
-        )
+            lambda z: len(z["text"].split()) < 100,
+        ).log_on_failure()
         result = await guardrail.run(generate_summary)
         print(f"Summary (passed guard): {result}")
     except GuardValidationError as e:
@@ -111,8 +110,7 @@ async def custom_function_guard_example():
     try:
         guardrail = Guardrails(
             content_safety_guard,  # Pass function reference
-            on_failure="raise",
-        )
+        ).raise_on_failure()
         # Default mapper handles str -> {"text": text, "prompt": text, "completion": text}
         result = await guardrail.run(generate_travel_advice)
         print(f"Travel advice: {result[:100]}...")
@@ -148,8 +146,7 @@ async def custom_handler_example():
     try:
         guardrail = Guardrails(
             lambda z: "danger" not in z["text"].lower(),
-            on_failure=custom_alert_handler,
-        )
+        ).on_failure(custom_alert_handler)
         result = await guardrail.run(generate_content)
         print(f"Content: {result[:100]}...")
     except GuardValidationError:


packages/sample-app/sample_app/guardrails/traceloop_evaluator_guard.py
```diff
@@ -0,0 +1,198 @@
...
+import json
+import os
+
+from openai import AsyncOpenAI
+
+from traceloop.sdk import Traceloop
+from traceloop.sdk.decorators import workflow, agent
+from traceloop.sdk.guardrail import (
+    OnFailure,
+    Guards,
```
galkleinman: **CRITICAL: `Guards` class does not exist — this will crash at import time**

`Guards` is never defined anywhere in the codebase. The `guards.py` module only exports bare functions like `toxicity_guard()`, `pii_guard()`, etc. This import will raise `ImportError`.

All 4 example files (`traceloop_evaluator_guard.py`, `decorator_example.py`, `multiple_guards_example.py`, `custom_evaluator_guard.py`) have this same issue. None of them will run.

Either create a `Guards` namespace class that maps to the factory functions, or rewrite the examples to use the bare function imports:
```python
from traceloop.sdk.guardrail.guards import toxicity_guard, pii_guard
```

packages/traceloop-sdk/traceloop/sdk/decorators/__init__.py
```diff
@@ -154,3 +160,95 @@ def atool(
         method_name=method_name,
         tlp_span_kind=TraceloopSpanKindValues.TOOL,
     )
+
+
+def guardrail(
```
galkleinman, line 218: **CRITICAL: Decorator signature is incompatible with how examples use it**

The examples pass guards as a keyword argument:
```python
@guardrail(name="...", guards=[Guards.toxicity_detector()], ...)
```

But this signature accepts guards as `*args` (positional). Passing `guards=[...]` as a keyword will raise `TypeError`.

Either:
- Change the signature to accept `guards` as a keyword: `def guardrail(*guards, guards=None, ...)`
- Or update all examples to pass guards positionally: `@guardrail(toxicity_guard(), on_failure=...)`

packages/traceloop-sdk/traceloop/sdk/guardrail/default_mapper.py
```diff
@@ -0,0 +1,54 @@
...
+        ValueError: If output type cannot be handled
+    """
+    if isinstance(output, str):
+        # Map string to common field names used by evaluators
+        input_dict = {
+            "text": output,
+            "prompt": output,
+            "completion": output
+        }
+        return [input_dict] * num_guards
```
galkleinman: **CRITICAL: Shared mutable reference — all guards receive the same dict object**

`[input_dict] * num_guards` creates a list where every element is the **same dictionary object**. If any guard mutates its input, all other guards see the mutation.

Fix:
```python
return [dict(input_dict) for _ in range(num_guards)]
```
Or use `copy.deepcopy` if nested structures are expected.

The same issue exists on the other `return` path below (`return [enriched] * num_guards`).

packages/traceloop-sdk/traceloop/sdk/guardrail/guardrail.py
```diff
@@ -0,0 +1,411 @@
...
+    def __init__(self, async_http_client: httpx.AsyncClient):
+        self._async_http = async_http_client
+        self._evaluator = Evaluator(async_http_client)
+        self._guards = []
+        self._on_failure = None
+        self._run_all = False
+        self._parallel = True
+        self._name = ""
+
+    def create(
```
galkleinman: **IMPORTANT — Non-Pythonic: `create()` method is a Java builder pattern**

Calling `create()` on an existing `Guardrails` instance to produce a *new* instance is the Java builder/factory pattern. A Python developer wouldn't expect `client.guardrails.create(...)` to return a new `Guardrails` object rather than mutating `self`.

Pythonic alternatives:
- Accept config in `__init__` directly: `Guardrail(client=client, guards=[...])`
- Use a `@classmethod` factory: `Guardrail.from_config(guards=[...])`
- Use `client.create_guardrail(...)`

This is a public API surface — worth getting right before the initial release.

packages/traceloop-sdk/traceloop/sdk/guardrail/condition.py
```diff
@@ -0,0 +1,155 @@
...
+Conditions are functions that take an evaluator result value and return a boolean
+indicating whether the guard should pass (True) or fail (False).
+
+The condition_field is extracted by the guard function before passing to these
+conditions, so conditions receive the field value directly.
+"""
+from typing import Any, Callable
+
+
+class Condition:
```
galkleinman, line 14: **IMPORTANT — Non-Pythonic: Class with only `@staticmethod` methods is a Java-ism**

The `Condition` class is purely a namespace of static methods. In Python, module-level functions are the idiomatic approach:
```python
from traceloop.sdk.guardrail.condition import is_true, greater_than
```

That said, the namespace does provide IDE autocomplete discoverability (`Condition.` shows all options). If this is intentional, document the rationale — otherwise Python developers may find it unexpected.

Same applies to the `OnFailure` class in `on_failure.py`.

packages/traceloop-sdk/traceloop/sdk/guardrail/condition.py
```diff
@@ -0,0 +1,155 @@
...
+    def is_true() -> Callable[[Any], bool]:
+        """
+        Pass if value is True.
+
+        Example:
+            toxicity_guard(condition=Condition.is_true())
+        """
+
+        def check(value: Any) -> bool:
+            return value is True
```
galkleinman: **IMPORTANT — `is True` uses identity check, not truthiness**

Using `value is True` means `1`, `"yes"`, non-empty lists, and any other truthy value will **not** pass. A Python developer would expect `is_true()` to mean "truthy" (i.e., `bool(value)`).

If strict identity is intentional, the docstring must clearly warn:
> "Only exact `True` passes; truthy values like `1` or non-empty strings do not pass."

Otherwise, use `return bool(value)` for Python-native truthiness semantics.

packages/traceloop-sdk/traceloop/sdk/guardrail/guardrail.py
```diff
@@ -0,0 +1,411 @@
...
+            duration_ms = (time.perf_counter() - start_time) * 1000
+            span.set_attribute(GEN_AI_GUARDRAIL_STATUS, "PASSED" if all_passed else "FAILED")
+            span.set_attribute(GEN_AI_GUARDRAIL_DURATION, duration_ms)
+            span.set_attribute(GEN_AI_GUARDRAIL_GUARD_COUNT, len(self._guards))
+            if failed_indices:
+                span.set_attribute(GEN_AI_GUARDRAIL_FAILED_GUARD_COUNT, len(failed_indices))
+
+        return all_passed, failed_indices
+
+    async def run(
```
galkleinman: **IMPORTANT — Non-Pythonic: Zero-argument callable forces awkward lambda wrapping**

The type annotation `Callable[[], Awaitable[Any]]` forces users to wrap their functions:
```python
result = await g.run(lambda: my_function(arg1, arg2))
```

A Pythonic API would accept `*args, **kwargs`:
```python
async def run(self, func_to_guard, *args, **kwargs):
    result = await func_to_guard(*args, **kwargs)
```

The `@guardrail` decorator already hides this by wrapping internally, but direct `run()` callers are stuck with the lambda pattern.

packages/traceloop-sdk/traceloop/sdk/decorators/__init__.py
```diff
@@ -154,3 +160,95 @@ def atool(
...
+                from traceloop.sdk import Traceloop
+
+                client = Traceloop.get()
+                g = client.guardrails.create(
+                    guards=guards_list,
+                    on_failure=failure_handler,
+                    name=name or func.__name__,
+                )
+                # Run async guardrail in event loop for sync functions
+                return asyncio.run(
```
galkleinman: **IMPORTANT — `asyncio.run()` will crash inside existing event loops**

`asyncio.run()` raises `RuntimeError: This event loop is already running` when called inside Jupyter notebooks, Django async views, or any context with an existing event loop.

Consider detecting an existing running loop and using an alternative:
```python
try:
    loop = asyncio.get_running_loop()
except RuntimeError:
    loop = None

if loop is not None:
    # Already in async context — use nest_asyncio or thread-with-new-loop
    ...
else:
    return asyncio.run(...)
```

packages/traceloop-sdk/traceloop/sdk/guardrail/guardrail.py
```diff
@@ -0,0 +1,411 @@
...
+    _async_http: httpx.AsyncClient
+    _guards: list[Guard]
+    _on_failure: Optional[OnFailureHandler]
+    _run_all: bool
+    _parallel: bool
+    _name: str
+
+    def __init__(self, async_http_client: httpx.AsyncClient):
+        self._async_http = async_http_client
+        self._evaluator = Evaluator(async_http_client)
```
galkleinman: **IMPORTANT — Dead code: `_evaluator` is created but never used**

This `Evaluator` instance is assigned here but never referenced anywhere in the `Guardrails` class. Guards create their own `Evaluator` instances inside `guard_fn` in `_create_guard()` (in `guards.py`). This field should be removed, or guards should receive and reuse it.

packages/traceloop-sdk/traceloop/sdk/decorators/__init__.py
```diff
@@ -154,3 +160,95 @@ def atool(
         method_name=method_name,
         tlp_span_kind=TraceloopSpanKindValues.TOOL,
     )
+
+
+def guardrail(
+    *guards: Guard,
+    input_mapper: InputMapper | None = None,
+    on_failure: Union[OnFailureHandler, str, None] = None,
```
galkleinman: **IMPORTANT — Inconsistent type annotation syntax**

Line above uses `InputMapper | None` (PEP 604 / Python 3.10+ syntax), while this line uses `Union[OnFailureHandler, str, None]` (old `typing.Union` syntax).

Pick one style. If the package supports Python 3.9, `|` syntax without `from __future__ import annotations` will fail at runtime.

packages/traceloop-sdk/traceloop/sdk/guardrail/model.py
```diff
@@ -0,0 +1,144 @@
...
+        self.guard_index = guard_index
+        self.expected_type = expected_type
+        self.actual_type = actual_type
+        self.validation_error = validation_error
+        super().__init__(message)
+
+    def __str__(self) -> str:
+        return (
+            f"{self.args[0]} [guard {self.guard_index}]: "
+            f"expected {self.expected_type}, got {self.actual_type.__name__}"
```
galkleinman, line 143: **IMPORTANT — `__str__` can raise `AttributeError`**

If `self.actual_type` is `None` or not a class (lacks `__name__`), this will crash. Safer:
```python
f"got {getattr(self.actual_type, '__name__', str(self.actual_type))}"
```

packages/traceloop-sdk/traceloop/sdk/guardrail/guards.py
```diff
@@ -0,0 +1,664 @@
...
+        # Convert Pydantic model to dict, or use dict directly
+        if isinstance(input_data, dict):
+            input_dict = input_data
+        elif hasattr(input_data, "model_dump"):
+            input_dict = input_data.model_dump()
+        else:
+            input_dict = dict(input_data)
+
+        client = Traceloop.get()
+        evaluator = Evaluator(client._async_http)
```
galkleinman: **Suggestion: New `Evaluator` created on every guard invocation**

Each guard call creates a fresh `Evaluator` instance and looks up the global `Traceloop` client. Consider caching the evaluator instance or passing it during guard creation. Note that `Guardrails.__init__` already creates an `Evaluator` (see other comment) — if that were passed down, guards could reuse it.

packages/traceloop-sdk/traceloop/sdk/guardrail/default_mapper.py
```diff
@@ -0,0 +1,54 @@
...
+        if "text" in output:
+            enriched.setdefault("prompt", output["text"])
+            enriched.setdefault("completion", output["text"])
+        if "question" in output:
+            enriched.setdefault("query", output["question"])
+        if "answer" in output:
+            enriched.setdefault("answer", output["answer"])
+            enriched.setdefault("completion", output["answer"])
+        if "context" in output:
+            enriched.setdefault("context", [output["context"]])
```
galkleinman, line 51: **Suggestion: Context wrapping may double-nest lists**

If `output["context"]` is already a list, `[output["context"]]` wraps it in another list. And if the key already exists in `enriched`, `setdefault` silently does nothing.

Consider normalizing unconditionally:
```python
if "context" in output:
    val = output["context"]
    enriched["context"] = val if isinstance(val, list) else [val]
```

packages/traceloop-sdk/traceloop/sdk/guardrail/guards.py
```diff
@@ -0,0 +1,664 @@
...
+        from traceloop.sdk import Traceloop
+        from traceloop.sdk.evaluator.evaluator import Evaluator
+
+        # Convert Pydantic model to dict, or use dict directly
+        if isinstance(input_data, dict):
+            input_dict = input_data
+        elif hasattr(input_data, "model_dump"):
+            input_dict = input_data.model_dump()
+        else:
+            input_dict = dict(input_data)
```
galkleinman, line 68: **Suggestion: `dict(input_data)` fallback will raise unhelpful `TypeError`**

For non-iterable, non-mapping types (a plain string, an int, etc.), `dict(input_data)` will raise a confusing `TypeError`. Consider raising a clear error instead:
```python
raise TypeError(
    f"Guard input must be a dict or Pydantic model, got {type(input_data).__name__}"
)
```

packages/traceloop-sdk/traceloop/sdk/guardrail/on_failure.py
```diff
@@ -0,0 +1,99 @@
...
+"""
+import logging
+from typing import Any, Callable, Type
+
+from .model import GuardedResult, GuardValidationError
+
+logger = logging.getLogger("traceloop.guardrail")
+
+
+class OnFailure:
```
galkleinman, line 15: **IMPORTANT — Same Java-ism as `Condition`: class with only `@staticmethod` methods**

See comment on `Condition` class. Module-level functions (`raise_exception()`, `log()`, `noop()`, `return_value()`) would be more Pythonic. If the namespace is intentional for discoverability, document the rationale.

packages/sample-app/sample_app/guardrails/validate_example.py
```diff
@@ -0,0 +1,140 @@
...
+    result = await output_guardrail.run(
+        lambda: generate_response(user_prompt),
+        input_mapper=lambda response_text: [
+            AnswerRelevancyInput(answer=response_text, question=user_prompt),
+            SexismDetectorInput(text=response_text),
+            ToxicityDetectorInput(text=response_text),
+        ],
+    )
+
+    print("Output validation passed.")
```
galkleinman, line 102: **Suggestion: This prints success unconditionally**

This line runs regardless of whether the guard actually passed or the `on_failure` handler returned a fallback value. Should check the result before claiming success.

packages/traceloop-sdk/traceloop/sdk/guardrail/condition.py
nina-kollman: @galkleinman I tried to convert to the module function design. I decided to revert it because I think the class usage in this case is more comfortable as it revels which options you have. 
Take a look on the [commit](https://github.com/traceloop/openllmetry/pull/3649/commits/1a8755bb9bccc974014d03ef865dd5c2b1720bfa) if you disagree.

galkleinman:
### Code review

Found 2 issues:

1. `Condition.is_true()` uses identity check (`value is True`) instead of truthiness. Python developers will expect `is_true()` to mean "truthy" (`bool(value)`), but it rejects `1`, `"yes"`, non-empty lists, and other truthy values. The docstring warns about this, but the fact that extensive documentation is needed suggests the naming is misleading. `is_exact_true()` or similar would be clearer, or `is_true()` should check truthiness and `is_exactly_true()` should check identity.

https://github.com/traceloop/openllmetry/blob/2d0aaa454d8822ee0eeb15adc515c2abfc66c233/packages/traceloop-sdk/traceloop/sdk/guardrail/condition.py#L8-L19

2. `guardrail` is not exported in `__all__` in `decorators/__init__.py`. The list still only contains the original 9 decorators, so `from traceloop.sdk.decorators import *` will silently omit `guardrail`.

https://github.com/traceloop/openllmetry/blob/2d0aaa454d8822ee0eeb15adc515c2abfc66c233/packages/traceloop-sdk/traceloop/sdk/decorators/__init__.py#L15-L25

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

galkleinman:
### DX feedback: API patterns are not Pythonic

The current guardrails API leans heavily on Java/C#-style static factory classes (`OnFailure.return_value(...)`, `Condition.is_true()`) and imperative factory methods (`client.create_guardrail(...)`). These work, but they fight how Python developers naturally express things. Below is a pattern-by-pattern breakdown with suggested alternatives.

---

#### 1. `OnFailure.*` — static namespace class should be plain strings / callables / chaining

**Current** (appears ~40+ times across production, examples, and tests):
```python
on_failure=OnFailure.raise_exception("PII detected in response")
on_failure=OnFailure.return_value(value="Sorry, I can't help you with that.")
on_failure=OnFailure.log(level=logging.WARNING, message="Guard failed")
on_failure=OnFailure.noop()
```

**Suggested — strings for common cases, callables for custom:**
```python
# Common cases as strings (like pytest markers, SQLAlchemy cascade, etc.)
on_failure="raise"
on_failure="log"
on_failure="ignore"  # noop

# With parameters — just pass the value directly
on_failure="Sorry, I can't help you with that."  # string = return this value

# Custom behavior — plain callable
on_failure=lambda result: log_and_alert(result)
```

**Or — fluent chaining on the result:**
```python
result = await guardrail.run(my_func).on_failure("raise")
result = await guardrail.run(my_func).on_failure(return_value="fallback")
```

The `OnFailure` class is just 4 `@staticmethod` methods returning closures. Module-level functions or string literals achieve the same with less ceremony.

---

#### 2. `Condition.*` — static namespace class should be plain Python expressions

**Current** (appears ~60+ times across `condition.py`, `guards.py`, `test_conditions.py`):
```python
condition=Condition.is_true()
condition=Condition.is_false()
condition=Condition.greater_than(0.8)
condition=Condition.less_than(1000)
condition=Condition.between(0.3, 0.7)
condition=Condition.equals("approved")
condition=Condition.greater_than_or_equal(0.8)
```

**Suggested — plain Python, since these are just predicates:**
```python
# Python already has a vocabulary for this
condition=bool                          # instead of Condition.is_truthy()
condition=lambda v: not v               # instead of Condition.is_falsy()
condition=lambda v: v > 0.8             # instead of Condition.greater_than(0.8)
condition=lambda v: 0.3 <= v <= 0.7     # instead of Condition.between(0.3, 0.7)
condition=lambda v: v == "approved"     # instead of Condition.equals("approved")
```

Or if you want named helpers for discoverability, use module-level functions with `operator`:
```python
from traceloop.sdk.guardrail.conditions import gt, lt, between, eq

condition=gt(0.8)       # reads naturally, no class namespace
condition=between(0.3, 0.7)
```

Every Python developer already knows `>`, `<`, `==`, `lambda`. Wrapping these in `Condition.greater_than_or_equal()` adds 30 characters for zero expressiveness.

---

#### 3. `client.create_guardrail(...)` — factory method should be direct construction or fluent

**Current** (appears in every example and all tests):
```python
client = Traceloop.init(app_name="my-app", ...)
guardrail = client.create_guardrail(
    guards=[pii_guard(), toxicity_guard()],
    on_failure=OnFailure.raise_exception("Content policy violation"),
    name="safety-check",
)
result = await guardrail.run(my_func)
```

**Suggested — direct construction, Python-style:**
```python
guardrail = Guardrail(
    pii_guard(),
    toxicity_guard(),
    on_failure="raise",
    name="safety-check",
)
result = await guardrail.run(my_func)
```

Or with chaining:
```python
result = await Guardrail(pii_guard(), toxicity_guard()).run(my_func).on_failure("raise")
```

In Python, if a class can be instantiated directly, it should be. The `client.create_guardrail()` indirection forces users to manage a client object just to create a guardrail. The HTTP client dependency can be resolved internally (from the initialized Traceloop singleton, which the guards already do).

---

#### 4. `guardrail.run(lambda: func(...))` — lambda wrapping should be transparent

**Current** (in decorator internals and many examples):
```python
# Decorator wraps in lambda internally:
return await g.run(
    lambda: func(*args, **kwargs),
    input_mapper=input_mapper,
)

# Examples force users to define zero-arg wrappers:
async def generate_summary() -> str:
    return await openai.chat(...)

result = await guardrail.run(generate_summary)
```

**Suggested — accept func + args directly:**
```python
result = await guardrail.run(generate_summary, prompt, model="gpt-4")

# or as a decorator that's transparent:
@guardrail(pii_guard(), on_failure="raise")
async def generate_summary(prompt: str) -> str:
    return await openai.chat(prompt)

# just call it normally:
result = await generate_summary("tell me about X")
```

This is how `asyncio.create_task(coro)`, `functools.partial(fn, *args)`, and `concurrent.futures.submit(fn, *args)` work in the stdlib. Python developers expect `run(fn, *args, **kwargs)`, not `run(lambda: fn(*args, **kwargs))`.

---

#### 5. `input_mapper=lambda text: [PIIDetectorInput(text=text)]` — positional list protocol is fragile

**Current** (appears in all examples using Traceloop evaluator guards):
```python
input_mapper=lambda response: [
    AnswerRelevancyInput(answer=response, question=prompt),
    SexismDetectorInput(text=response),
    ToxicityDetectorInput(text=response),
]
```

The user must return a list whose elements are position-matched to the guards list. If guards are reordered, the mapper silently sends wrong inputs to wrong guards.

**Suggested — dict keyed by guard name or guard reference:**
```python
input_mapper=lambda response: {
    "answer_relevancy": AnswerRelevancyInput(answer=response, question=prompt),
    "sexism": SexismDetectorInput(text=response),
    "toxicity": ToxicityDetectorInput(text=response),
}
```

Or let each guard declare its own input mapping:
```python
pii_guard(input=lambda r: PIIDetectorInput(text=r))
toxicity_guard(input=lambda r: ToxicityDetectorInput(text=r))
```

This is self-documenting and order-independent.

---

#### 6. Guard factory functions should compose more naturally

**Current:**
```python
guardrail = client.create_guardrail(
    guards=[
        toxicity_guard(condition=Condition.is_true()),
        pii_guard(condition=Condition.is_false()),
        custom_evaluator_guard(
            evaluator_slug="medicaladvice",
            condition=Condition.greater_than_or_equal(0.8),
        ),
    ],
    on_failure=OnFailure.raise_exception("Content failed safety checks"),
)
```

**Suggested — more composable, reads like English:**
```python
guardrail = Guardrail(
    toxicity_guard(),           # sensible defaults, no Condition needed
    no_pii(),                   # name implies the condition
    custom_evaluator("medicaladvice", threshold=0.8),
    on_failure="raise",
)
```

The `Condition` parameter is needed because the guard names don't encode their polarity (`pii_guard` + `Condition.is_false()` means "pass when no PII"). Renaming to `no_pii()` or `pii_free()` eliminates the need for an explicit condition in the common case.

---

#### TL;DR

The current API requires users to learn `OnFailure`, `Condition`, `create_guardrail`, `input_mapper` list protocol, and lambda-wrapping patterns. A Pythonic API would use strings, plain callables, direct construction, and `*args/**kwargs` — things every Python developer already knows. The goal should be that a developer can read a guardrail definition and understand it without reading any docs.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/sample-app/sample_app/guardrails/custom_evaluator_guard.py
```diff
@@ -0,0 +1,158 @@
...
+                    "role": "user",
+                    "content": user_question,
+                }
+            ],
+        )
+        return completion.choices[0].message.content
+
+    g = Guardrails(
+        custom_evaluator_guard(evaluator_slug="diagnosis-blocker", condition_field="pass"),
+        on_failure="raise",
```
galkleinman, line 116: shouldn't be something strong/structured than hardcoded string?

packages/sample-app/sample_app/guardrails/multiple_guards_example.py
```diff
@@ -0,0 +1,293 @@
...
+            {"caps_ratio": caps_ratio},   # For capitalization guard
+        ]
+
+    guardrail = Guardrails(
+        lambda z: z["word_count"] < 200,           # Guard 0: Length check
+        lambda z: "danger" not in z["text"].lower(),  # Guard 1: Forbidden words
+        lambda z: z["caps_ratio"] < 0.3,           # Guard 2: No excessive caps
+        on_failure="raise",
+        parallel=True,
+    )
```
galkleinman: also, could look better imo:

```suggestion
    guardrail = Guardrails(
        lambda z: z["word_count"] < 200,           # Guard 0: Length check
        lambda z: "danger" not in z["text"].lower(),  # Guard 1: Forbidden words
        lambda z: z["caps_ratio"] < 0.3,           # Guard 2: No excessive caps
    ).parallel().raise_on_failure()
```

packages/sample-app/sample_app/guardrails/traceloop_evaluator_guard.py
```diff
@@ -0,0 +1,268 @@
...
+    """Demonstrate PII detection guard using Traceloop evaluator."""
+
+    guardrail = Guardrails(
+        pii_guard(probability_threshold=0.7, timeout_in_sec=45),
+        on_failure="raise",
+    )
+    result = await guardrail.run(
+        generate_customer_response,
+        input_mapper=lambda text: [PIIDetectorInput(text=text)],
+    )
```
galkleinman, line 65: same here with:


```suggestion
    result = await guardrail.run(
        generate_customer_response
    ).map_input_with(lambda text: [PIIDetectorInput(text=text)])
```

galkleinman (approved):
in general it looks good, i think in terms of the DX, we can do better imo and introduce a DSL which is more functional. Maybe it is worth that we'll merge, start experimenting with it and than improve if we'll find it necessary. 

The only thing which is a must change here is those hardcoded string which is used to decide the action to be taken upon failure and so on (like the "raise") - it is really a bad practice.

Issue #3439, opened 2025-11-12
🚀 Feature: Add `graph_structure` attribute to LangGraph instrumentation


ronensc:
### Which component is this feature for?

Langchain Instrumentation

### 🔖 Feature description

Add an attribute named `graph_structure` to the main LangGraph workflow span that represents the structure of the invoked graph, i.e., the nodes and edges that compose it.

### 🎤 Why is this feature needed ?

Currently, there is no way to reconstruct the executed graph from the exported spans.  
Including a `graph_structure` attribute would allow visualizing and analyzing the shape of the LangGraph workflow (e.g., nodes, edges, and branching paths) within a trace visualization or observability platform.
This enables better understanding of execution flow and easier debugging of agentic workflows.

### ✌️ How do you aim to achieve this?

An initial implementation will be proposed via a pull request, adding the `graph_structure` attribute to the root LangGraph workflow span.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
Sounds good @ronensc!

Avijeet-Mandal:
that a smart idea to have this attribute. @ronensc  please assign this task to me

ronensc:
Thanks @Avijeet-Mandal!
I have an initial implementation including a basic test that you might find useful as a starting point: https://github.com/traceloop/openllmetry/commit/585062016736a8c9324b03360992277fa4bf136c

Avijeet-Mandal:
that helps😀. looking into it. @ronensc

PR #3673, opened 2026-02-11
feat(langchain): add OpenTelemetry GenAI semantic conventions

jemo21k:
This PR adds OpenTelemetry GenAI semantic convention support to LangChain/LangGraph instrumentation, aligned with the official [GenAI Agent Spans specification](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/).

**Fixes:** https://github.com/traceloop/openllmetry/issues/3439

**Spec references:**
- [GenAI Agent Spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/#spans)
- [GenAI Events](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/)

### Span Naming (GenAI convention)
- `invoke_agent {name}` - Graph/agent invocations
- `execute_task {name}` - Task spans
- `execute_tool {name}` - Tool spans  
- `create_agent {name}` - Agent factory calls
- `goto {target}` - Command routing

### New Attributes
- `gen_ai.operation.name` - Operation type
- `gen_ai.provider.name` - Provider (LangGraph/langchain)
- `gen_ai.agent.id/name` - Agent identification
- `gen_ai.task.id/name/status/kind` - Task tracking
- `gen_ai.conversation.id` - Thread/session tracking
- `gen_ai.tool.definitions` - Tool schemas (OpenAI format)
- `gen_ai.workflow.structure` - Graph topology JSON
- `gen_ai.system_instructions` - Agent system prompts
- `langgraph.command.source_node/goto_node` - Command routing

### Instrumentation Coverage
- LangGraph `Pregel.stream/astream` (graph invocation)
- LangGraph `Command.__init__` (routing)
- LangGraph `create_react_agent` factory
- LangChain `AgentMiddleware` hooks
- Retriever callbacks

**Acknowledgments:**
Special thanks to @galkleinman for guidance on the semantic conventions approach.

---

----

> [!IMPORTANT]
> Added comprehensive LangGraph instrumentation with GenAI semantic conventions, including graph invocation tracing, middleware hook wrapping, agent factory instrumentation, and retriever operation tracking.
> 
>   ## LangGraph Instrumentation
>   - **Core LangGraph Wrapping**: Added `_wrap_langgraph_components()` method to wrap `Pregel.stream` and `Pregel.astream` for graph invocation tracing, and `Command.__init__` to capture routing commands with source and destination node information.
>   - **Agent Factory Wrapping**: Implemented `_wrap_agent_factories()` to instrument both LangGraph prebuilt agents (`create_react_agent`) and LangChain agents (`create_agent`) at their module locations and re-export locations.
>   - **Middleware Hook Wrapping**: Added `_wrap_middleware_hooks()` to wrap synchronous and asynchronous `AgentMiddleware` hook methods (`before_model`, `after_model`, `before_agent`, `after_agent` and their async variants).
>   - **Uninstrumentation**: Extended `_uninstrument()` method to properly unwrap all LangGraph components, middleware hooks, and agent factory functions.
>   ## GenAI Semantic Conventions
>   - **New Attributes**: Added comprehensive GenAI semantic convention attributes to `SpanAttributes` class including operation names, provider, conversation ID, agent attributes (ID, name, system instructions), task attributes (ID, name, parent ID, input, output, status, kind), tool attributes (name, description, type, definitions, call arguments, results), and workflow structure.
>   - **Enums**: Created `GenAIOperationName` enum with values for `create_agent`, `invoke_agent`, `execute_task`, `execute_tool`, `llm_request`, and `vector_db_retrieve` operations, and `GenAITaskStatus` enum with `success` and `failure` values.
>   ## Callback Handler Enhancements
>   - **LangGraph Context Tracking**: Updated `on_chain_start()` to extract conversation ID from config and set LangGraph node context; updated `_create_span()` to check for LangGraph flow context and parent first child spans to graph spans.
>   - **Task Span Attributes**: Enhanced `_create_task_span()` with GenAI semantic convention attributes including provider name, operation name, and kind-specific attributes (agent name/ID, tool name/type/description, task name/ID/parent ID).
>   - **Retriever Operations**: Added `on_retriever_start()`, `on_retriever_end()`, and `on_retriever_error()` methods to handle retriever callbacks with `vector_db_retrieve` span naming and GenAI semantic convention attributes.
>   - **LLM Span Attributes**: Added GenAI operation name attribute to `_create_llm_span()`.
>   - **Status Tracking**: Updated `on_chain_end()`, `on_tool_end()`, and `_handle_error()` to set task status to success or failure.
>   ## Utilities and Helpers
>   - **Graph Extraction**: Added `langgraph_utils.py` module with `extract_graph_json()` function to extract nodes and edges from LangGraph Pregel instances, filtering special nodes and grouping edges by source.
>   - **Patch Utilities**: Implemented `patch.py` module with helper functions for generating unique agent IDs, managing span lifecycle and context, extracting graph names, setting GenAI attributes on spans, extracting tool definitions, and wrapping middleware hooks and agent factories.
>   ## Test Updates
>   - **Span Name Convention**: Updated all test files (`test_agents.py`, `test_chains.py`, `test_documents_chains.py`, `test_lcel.py`, `test_llms.py`) to use new span naming convention: changed from `"ComponentName.task"` to `"execute_task ComponentName"` format.
>   - **LangGraph Tests**: Added new test functions in `test_langgraph.py` including `test_create_react_agent_span`, `test_retriever_span_attributes`, `test_middleware_hook_span_attributes`, `test_langgraph_custom_name`, `test_command_with_goto`, `test_send_extraction`, `test_create_agent_with_system_prompt`, `test_async_middleware_hook`, and `test_middleware_failure_status` to verify GenAI semantic convention attributes and LangGraph-specific functionality.
>   - **Span Hierarchy**: Updated span hierarchy assertions to account for new `"invoke_agent LangGraph"` span as parent of workflow spans.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for 27907465a21bf9e6bec94701559833ddb0e0c603. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Expanded LangGraph/LangChain tracing: graph invocation spans, command routing, agent creation, middleware-hook instrumentation, retriever spans, and standardized "execute_task/execute_tool" span naming.
  * GenAI semantic attributes added for providers, agents, tasks, tools, LLM requests, retriever results, conversation IDs, and workflow structure.
  * Graph extraction utility to produce JSON representations.

* **Tests**
  * Updated and expanded tests for span names, GenAI attributes, graph spans, middleware hooks, and sync/async paths.

* **Bug Fixes**
  * Improved cleanup/uninstrumentation to reliably unwrap instrumented components.


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -418,21 +506,42 @@ def on_chain_start(
...
+        # Note: We intentionally don't store/detach this token because in async
+        # scenarios, detaching tokens created in a different context causes errors.
+        # The context is automatically cleaned up when the graph execution completes.
+        if metadata and metadata.get("langgraph_node") == name:
+            try:
+                context_api.attach(
+                    context_api.set_value(LANGGRAPH_CURRENT_NODE_KEY, name)
+                )
+            except Exception:
+                pass
```
galkleinman, line 550: Context token is attached but never detached

I understand the comment about async detach errors, but never detaching a context token is a leak — in long-running services this accumulates over time. Could we explore using a contextvars.ContextVar instead? That would sidestep the OTel attach/detach lifecycle entirely while still being async-safe

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
```diff
@@ -290,3 +325,24 @@ class TraceloopSpanKindValues(Enum):
...
+    EXECUTE_TOOL = "execute_tool"
+    LLM_REQUEST = "llm_request"
+    VECTOR_DB_RETRIEVE = "vector_db_retrieve"
+
+
+class GenAITaskStatus(Enum):
+    """Task execution status values for gen_ai.task.status attribute."""
+
+    SUCCESS = "success"
+    FAILURE = "failure"
```
galkleinman: why aren't we just importing from the official otel semconv the attributes that exist there?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
```diff
@@ -0,0 +1,492 @@
...
+    if config and isinstance(config, dict):
+        configurable = config.get("configurable", {})
+        thread_id = configurable.get("thread_id")
+        if thread_id:
+            graph_span.set_attribute(SpanAttributes.GEN_AI_CONVERSATION_ID, str(thread_id))
+
+    # Extract workflow structure (best-effort with debug logging)
+    try:
+        graph_structure = extract_graph_json(instance)
+        graph_span.set_attribute(SpanAttributes.GEN_AI_WORKFLOW_STRUCTURE, graph_structure)
```
galkleinman: The current approach serializes the graph structure as a JSON string in a single attribute.

OTel natively supports `string[]` attributes, which would be more idiomatic, more compact, and directly queryable by trace backends (no JSON parsing needed):


```python
span.set_attribute("gen_ai.workflow.nodes", ["node1", "node2", "node3"])
span.set_attribute("gen_ai.workflow.edges", ["node1 -> node2", "node1 -> node3"])
```

This eliminates the JSON overhead, the confusing grouped edge format, and makes the data first-class in any OTel-compatible backend. The extraction function also simplifies significantly - no json.dumps, no edge grouping needed.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
```diff
@@ -0,0 +1,492 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _generate_agent_id(name: str, instance: Any) -> str:
+    """Generate a unique agent ID by hashing the instance memory address."""
+    instance_id = str(id(instance))
+    hash_input = f"{name}_{instance_id}"
+    hashed = hashlib.sha256(hash_input.encode()).hexdigest()[:16]
+    return f"{name}_{hashed}"
```
galkleinman: Python reuses memory addresses after garbage collection, so id(instance) can produce the same value for different objects across invocations. 

Use `uuid.uuid4()` for a truly unique ID?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/langgraph_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        structure = {
+            "nodes": nodes,
+            "edges": edges,
+        }
+
+        return json.dumps(structure)
+
+    except Exception as e:
+        # Re-raise to let caller handle
+        raise Exception(f"Failed to extract graph structure: {e}") from e
```
galkleinman: Consider either letting the original exception propagate naturally, or defining a small custom exception if you want the wrapper message.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -326,9 +367,53 @@ def _create_task_span(
...
         )
 
+        # Set traceloop attributes for backwards compatibility
         _set_span_attribute(span, SpanAttributes.TRACELOOP_SPAN_KIND, kind.value)
         _set_span_attribute(span, SpanAttributes.TRACELOOP_ENTITY_NAME, entity_name)
 
+        # Set GenAI semantic convention attributes
+        # Check LangGraph flow context to set appropriate provider
+        langgraph_flow = context_api.get_value(LANGGRAPH_FLOW_KEY)
+        provider_name = "LangGraph" if langgraph_flow else "langchain"
```
galkleinman: should be lowercase according to otel semconv

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
```diff
@@ -0,0 +1,492 @@
...
+
+# Context key for marking LangGraph flow
+LANGGRAPH_FLOW_KEY = "langgraph_flow"
+# Context key for storing the graph span as parent for callback-created spans
+LANGGRAPH_GRAPH_SPAN_KEY = "langgraph_graph_span"
+# Context key for tracking if first child of graph span is pending (mutable list [bool])
+LANGGRAPH_FIRST_CHILD_PENDING_KEY = "langgraph_first_child_pending"
+# Context key for tracking current node (for Command source tracking)
+LANGGRAPH_CURRENT_NODE_KEY = "langgraph_current_node"
```
galkleinman: As I mentioned during our meeting, i'm not sure the context tricks will work here, or at least won't work for some cases, since we're using langchain callback mechanism which is called with different contexts. look at the `SpanHolder` examples, this is the only (ugly) way of passing data between different callbacks of the "same context".

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
jemo21k: Sure we I implement a solution where we detach it also.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
jemo21k: regarding the semantic conventions alignment - I checked the official package and the GenAI attributes are still in incubation (opentelemetry.semconv._incubating.attributes.gen_ai_attributes.GenAiOperationNameValues). They only have CHAT, EMBED, and TEXT_COMPLETION right now.

Our code uses agent-specific operations like CREATE_AGENT, INVOKE_AGENT, EXECUTE_TASK, etc. that aren't in the official spec yet.

Should I import and use the official values where they exist, or just keep our custom enum and document which values are extensions? What's your preference?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
jemo21k: About the agent ID generation - you're right that id(instance) can get reused after GC. We want the same graph instance to have the same gen_ai.agent.id across both the create_agent and invoke_agent spans so backends can correlate them.

Approach 1: Cache UUID on instance


def _generate_agent_id(name: str, instance: Any) -> str:
    if not hasattr(instance, '_otel_agent_id'):
        unique_id = uuid.uuid4().hex[:16]
        instance._otel_agent_id = f"{name}_{unique_id}"
    return instance._otel_agent_id
Approach 2: Use ContextVar mapping (avoids mutating instance)


import contextvars
from weakref import WeakKeyDictionary

# Module level
_agent_id_cache: WeakKeyDictionary = WeakKeyDictionary()

def _generate_agent_id(name: str, instance: Any) -> str:
    if instance not in _agent_id_cache:
        unique_id = uuid.uuid4().hex[:16]
        _agent_id_cache[instance] = f"{name}_{unique_id}"
    return _agent_id_cache[instance]
The WeakKeyDictionary automatically cleans up when instances are GC'd, and we don't mutate the graph instance.

Which approach do you prefer, or is there a better pattern you'd recommend?

jemo21k:
You mentioned on_retriever_error() should exist - it's already there at callback_handler.py lines 1030-1039. Did you miss it in the review, or were you referring to something else about that handler?

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
jemo21k: You mentioned that context tricks won't work with LangChain callbacks and that SpanHolder is the way to go. We're currently using 4 context keys:

LANGGRAPH_FLOW_KEY - Marks LangGraph vs langchain flow
LANGGRAPH_GRAPH_SPAN_KEY - Stores graph span for parent-child linking
LANGGRAPH_FIRST_CHILD_PENDING_KEY - Tracks first child with mutable list pattern
LANGGRAPH_CURRENT_NODE_KEY - Tracks current node for Command source tracking
Our tests (including async ones) are all passing, but I want to understand which context keys are problematic. Should we refactor all of them to use SpanHolder, or just specific ones? Is this a blocker for merging or something we can address in follow-up work?

jemo21k:
@galkleinman 

Thanks for the detailed review! I've pushed fixes for the items we agree on, and have a few questions on the remaining points.

Done:

- Graph structure: Changed from JSON string to native string arrays (gen_ai.workflow.nodes, gen_ai.workflow.edges with "source -> target" format)
- Provider name: Lowercased to "langgraph" / "langchain" everywhere
- Context token leak: Replaced the leaking context_api.attach() with contextvars.ContextVar - no manual detach needed, async-safe
- Exception wrapping: Removed the generic Exception wrapper in extract_graph_structure, exceptions now propagate naturally (caller already handles them with logger.debug)

galkleinman:
> You mentioned on_retriever_error() should exist - it's already there at callback_handler.py lines 1030-1039. Did you miss it in the review, or were you referring to something else about that handler?

missed it i guess, sorry :)

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
galkleinman: I would use the the official ones when exist, even if incubating... otherwise we're just copying the incubation here.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
galkleinman: honestly i'm not a fan of both, not because they're technically wrong, i just think that and agent id if such exist should be explicitly passed/declared. If we have to, and i strongly advise against it, I would go with ContextVar (which as mentioned in previous comment, not sure will work for real cases) since mutating the instance feels smelly.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
galkleinman: It's weird to me that everything work, maybe i'm missing something - context is not preserved in callbacks, because one callback isn't "nested" inside the other, and the whole context propagation of otel usually works on "cloning" the context when a span is being created in the context/nested of another one. I mean, you can use context, but you need the spanholder to hold a reference for the right one.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
jemo21k: Removed auto-generated gen_ai.agent.id entirely.  Relying on gen_ai.agent.name only for now.

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/patch.py
jemo21k: The graph span stored in context is now wrapped in a SpanHolder The callback reads graph_span_holder.context for correct span parenting instead of a raw Span reference.

Context keys: I believe the context keys work reliably because we're not passing context between independent callbacks. We instrument Pregel.stream/astream directly, so the flow is:

stream() wrapper [attaches context keys here]
  └── wrapped(*args, **kwargs)  [original graph execution]
        └── callback: on_chain_start  [reads context - same call stack]
        └── callback: on_llm_start    [reads context - same call stack]
        
The callbacks execute within the scope of the wrapped call, so they naturally inherit the OTel context. This is wrapper-to-callback propagation (which works), not callback-to-callback propagation (which doesn't). Using SpanHolder to replace these entirely would require bridging the graph span created in the stream wrapper (outside callbacks) into the callback handler's self.spans dict, which would need either passing the handler instance into patch.py or a global registry - both adding more coupling than OTel context.

packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py
jemo21k: Now importing official GenAiOperationNameValues from opentelemetry.semconv._incubating for CREATE_AGENT, INVOKE_AGENT, EXECUTE_TOOL. Renamed our custom enum to GenAICustomOperationName and it only contains extensions not yet in the official spec (EXECUTE_TASK, LLM_REQUEST, VECTOR_DB_RETRIEVE).

jemo21k:
@galkleinman Thanks for having the time to help with this PR. 
The PR is ready.

galkleinman:
tests are failing now 😞

jemo21k:
It was an import Issue. Should be fixed.

PR #3681, opened 2026-02-14
fix: preserve return values for RealtimeSession context manager methods

### LINKED ISSUE #3680: @dont_throw decorator breaks RealtimeSession async context manager protocol
## Bug

The `@dont_throw` decorator applied to `traced_aenter` and `traced_aexit` in `_realtime_wrappers.py` violates Python's [async context manager protocol](https://docs.python.org/3/reference/datamodel.html#asynchronous-context-managers).

When any exception occurs in the tracing/instrumentation code after the original method succeeds, `dont_throw` catches the exception and implicitly returns `None` instead of the session object.

### Expected behavior

`async with RealtimeSession(...) as session:` should always yield the session object, even if instrumentation code fails internally.

### Actual behavior

`session` is `None` when any tracing code throws, causing silent WebSocket connection failures with no indication that instrumentation is the root cause.

### Root cause

`dont_throw` in `utils.py` catches all exceptions and logs them, but the wrapper then falls through and returns `None`:

```python
async def async_wrapper(*args, **kwargs):
    try:
        return await func(*args, **kwargs)
    except Exception as e:
        _handle_exception(e, func, logger)
        # implicitly returns None
```

In `traced_aenter`, the original `__aenter__` succeeds and returns the session, but if any subsequent tracing code throws, the entire return value is lost. The inner `try/except` blocks already guard the tracing code, but the outer `@dont_throw` also wraps the original method call and the `return result` statement.

`@dont_throw` is the correct pattern for most instrumentation methods (e.g. `traced_put_event`, `traced_send_message`) where returning `None` is harmless. But `__aenter__` must return `self` per the [async context manager protocol](https://docs.python.org/3/reference/datamodel.html#asynchronous-context-managers), and `__aexit__`'s return value controls [exception suppression behavior](https://peps.python.org/pep-0492/#asynchronous-context-managers-and-async-with).

### Proposed fix

Remove `@dont_throw` from `traced_aenter` and `traced_aexit`. The inner `try/except` blocks already protect against tracing failures -- replace the bare `pass` with debug logging to match the `dont_throw` logging pattern.

### Affected file

`packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_realtime_wrappers.py`

### Workaround

Restore original methods after initialization:

```python
from opentelemetry.instrumentation.openai_agents._realtime_wrappers import _original_methods
from agents.realtime.session import RealtimeSession
for method in ("__aenter__", "__aexit__"):
    if method in _original_methods:
        setattr(RealtimeSession, method, _original_methods[method])
```

EliJaghab:
Fixes #3680

## Why

`@dont_throw` on `traced_aenter` and `traced_aexit` violates Python's
[async context manager protocol](https://docs.python.org/3/reference/datamodel.html#asynchronous-context-managers).
If any tracing code throws after the original method succeeds, the decorator
swallows the exception and returns `None` instead of the session object.

This causes `async with RealtimeSession(...) as session:` to yield `None`,
leading to silent WebSocket connection failures with no indication that
instrumentation is the root cause.

## What changed

- Removed `@dont_throw` from `traced_aenter` and `traced_aexit`
- Replaced bare `except: pass` with `logger.debug(...)` so tracing failures
  are still observable
- `@dont_throw` remains on `traced_put_event` and `traced_send_message`
  where returning `None` is harmless

----

> [!IMPORTANT]
> Remove `@dont_throw` from `traced_aenter` and `traced_aexit` in `_realtime_wrappers.py` to fix context manager protocol violation, adding logging for exceptions.
> 
>   - **Behavior**:
>     - Removed `@dont_throw` from `traced_aenter` and `traced_aexit` in `_realtime_wrappers.py` to comply with async context manager protocol.
>     - Added `logger.debug(...)` to log exceptions in `traced_aenter` and `traced_aexit`.
>     - `@dont_throw` remains on `traced_put_event` and `traced_send_message` where returning `None` is harmless.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for cb4642a4876b4606f2c598afdefafdeff08d7efb. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Release Notes

* **Bug Fixes**
  * Improved error visibility in OpenAI agents instrumentation by adding logging for initialization and cleanup failures that were previously silent.


galkleinman (approved):
LGTM 💯

PR #3688, opened 2026-02-15
fix(openai-agents): fix realtime session event handling for prompts, completions, and usage

### LINKED ISSUE #3685: Realtime session spans missing prompt, completion, and token usage attributes
## Bug

`openai.realtime` spans have no `gen_ai.prompt`, `gen_ai.completion`, or `gen_ai.usage` attributes. The root cause is event type mismatches in `traced_put_event` in `_realtime_wrappers.py`.

## Specific problems

1. **`response.done` check never matches** (line ~578): The code checks for `type == "response.done"`, but the SDK wraps raw OpenAI API events as `RealtimeModelRawServerEvent` with `type="raw_server_event"`. The actual API event type is nested inside `data.data.type`. The check needs to first match `raw_server_event`, then inspect `data.data.type` for `"response.done"`.

2. **No handler for `history_updated` events**: The SDK emits `RealtimeHistoryUpdated` events for assistant transcripts (e.g., when the model finishes speaking), but only `history_added` is handled. This means assistant-side completions from transcript updates are never captured.

3. **`response` event type handler is dead code** (lines ~539-554): No SDK event has `type="response"`. This branch can never execute.

4. **`record_usage()` is never called**: Its only call site is inside the `response.done` branch (problem 1), so token usage is never captured for realtime sessions.

## SDK event type hierarchy (openai-agents v0.6.0+)

For context, the event nesting works like this:

- **Session-level events** (what `traced_put_event` sees): `history_added`, `history_updated`, `raw_model_event`, etc.
- **Model-level events** (inside `raw_model_event.data`): `item_updated`, `turn_ended`, `raw_server_event`, etc.
- **Raw API events** (inside `raw_server_event.data`): `response.done`, `conversation.item.created`, etc.

The instrumentation code is checking for raw API event type strings at the session event level, where they will never appear.

## Related

- #3680 (monkey-patching issues in the same package)

EliJaghab:
## Summary

- Add `history_updated` event handler to capture assistant transcript updates
- Fix dict-based data access in `response.done` handler (`getattr` on dicts silently returned `None` instead of using `.get()`)
- Fix dict-case event unwrapping where the `data` variable was not updated to the nested level
- Remove dead `response` event handler (no SDK session event has `type="response"`)

## Why

The realtime instrumentation had several event type mismatches with the openai-agents SDK (v0.6.0+):

1. `history_updated` events were completely ignored, so assistant transcript updates from multi-turn conversations were never captured as completions.
2. In the `response.done` handler, `getattr()` was used on dict objects instead of `.get()`, so completion extraction silently failed (usage was captured correctly via the dict path, but output/content was not).
3. The dict-case unwrapping in `raw_model_event` did not update the `data` reference after finding nested data, so subsequent dict-path code looked at the wrong nesting level.
4. The `response` event handler was dead code -- no SDK session event has `type="response"`.

Closes #3685

## Test plan

- [x] All 46 package tests pass
- [x] Added test for `history_updated` assistant completion capture via transcript
- [x] Added test for dict-based `response.done` usage and completion extraction
- [x] Ruff lint passes

----

> [!IMPORTANT]
> Fix event handling in OpenAI Agents SDK by adding `history_updated` handler, correcting `response.done` data extraction, and removing dead code.
> 
>   - **Behavior**:
>     - Add `history_updated` event handler in `_realtime_wrappers.py` to capture assistant transcript updates.
>     - Fix `response.done` handler to use `.get()` instead of `getattr()` for dicts, ensuring completion extraction.
>     - Correct dict-case event unwrapping in `raw_model_event` to update `data` to the nested level.
>     - Remove dead `response` event handler.
>   - **Tests**:
>     - Add test for `history_updated` event handling in `test_realtime_session.py`.
>     - Add test for dict-based `response.done` handling in `test_realtime_session.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for ba639deb6deea159f85ea98d0ee8530bb0689e7f. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved realtime tracing for OpenAI agent interactions: more reliable capture of assistant completions and usage metrics across varied event types and payload formats.

* **Tests**
  * Added tests validating history-based completion extraction and response payload parsing (with and without usage) to ensure tracing accuracy and robustness.

### DIFF
--- a/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_realtime_wrappers.py
+++ b/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_realtime_wrappers.py
@@ -27,6 +27,67 @@ _original_methods: Dict[str, Any] = {}
 _tracing_states: Dict[int, "RealtimeTracingState"] = {}
 
 
+def _extract_content_text(item_content) -> Optional[str]:
+    """Extract text or transcript from content parts list.
+
+    Handles both dict-based and object-based content parts,
+    as well as plain string content.
+    """
+    if isinstance(item_content, str):
+        return item_content
+    if not item_content or not isinstance(item_content, list):
+        return None
+    for part in item_content:
+        if isinstance(part, dict):
+            text = part.get("text") or part.get("transcript")
+        else:
+            text = getattr(part, "text", None) or getattr(part, "transcript", None)
+        if text:
+            return text
+    return None
+
+
+def _unwrap_raw_event_data(data) -> Tuple[Optional[str], Any]:
+    """Extract type and unwrap nested data from raw_model_event.
+
+    The SDK sometimes double-wraps events as data.data.type.
+    Returns (data_type, unwrapped_data).
+    """
+    if isinstance(data, dict):
+        data_type = data.get("type")
+        raw_data = data.get("data")
+        if raw_data and isinstance(raw_data, dict):
+            nested_type = raw_data.get("type")
+            if nested_type:
+                return nested_type, raw_data
+        return data_type, data
+    else:
+        data_type = getattr(data, "type", None)
+        nested_data = getattr(data, "data", None)
+        if nested_data:
+            if isinstance(nested_data, dict):
+                nested_type = nested_data.get("type")
+            else:
+                nested_type = getattr(nested_data, "type", None)
+            if nested_type:
+                return nested_type, nested_data
+        return data_type, data
+
+
+def _extract_response_and_usage(data) -> Tuple[Any, Optional[dict]]:
+    """Extract response object and usage dict from response.done data.
+
+    Returns (response, usage).
+    """
+    if isinstance(data, dict):
+        response = data.get("response", {})
+        usage = response.get("usage") if isinstance(response, dict) else None
+    else:
+        response = getattr(data, "response", None)
+        usage = getattr(response, "usage", None) if response else None
+    return response, usage
+
+
 class RealtimeTracingState:
     """Tracks OpenTelemetry spans for a realtime session."""
 
@@ -513,20 +574,8 @@ def wrap_realtime_session(tracer: Tracer):
                     item = getattr(event, "item", None)
                     if item:
                         role = getattr(item, "role", None)
-                        content = None
-
                         item_content = getattr(item, "content", None)
-                        if item_content:
-                            if isinstance(item_content, list):
-                                for part in item_content:
-                                    if hasattr(part, "text"):
-                                        content = part.text
-                                        break
-                                    elif hasattr(part, "transcript"):
-                                        content = part.transcript
-                                        break
-                            elif isinstance(item_content, str):
-                                content = item_content
+                        content = _extract_content_text(item_content)
 
                         if not content:
                             content = getattr(item, "text", None) or getattr(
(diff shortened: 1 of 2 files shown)


EliJaghab:
Tested locally in our production voice relay (OpenAI Agents SDK realtime mode) with Jaeger trace collection.

With this fix applied, `openai.realtime` LLM spans now correctly capture:
- `gen_ai.prompt.0.content` -- e.g. "I want the iPhone 17 Pro Max"
- `gen_ai.completion.0.content` -- e.g. "Got it. The iPhone 17 Pro Max is selected. Which color would you like..."
- `gen_ai.usage.input_tokens` -- e.g. 9192
- `gen_ai.usage.output_tokens` -- e.g. 27

Before this fix, all of these attributes were missing on realtime spans. Confirmed across a 3-message automated conversation with multiple tool calls, producing 8 `openai.realtime` spans total.

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_realtime_wrappers.py
```diff
@@ -529,33 +529,34 @@ async def traced_put_event(self, event):
...
-                                    elif isinstance(item_content, str):
-                                        state.record_completion(role, item_content)
+                                if item_content and isinstance(item_content, list):
+                                    for part in item_content:
+                                        text = getattr(part, "text", None) or getattr(
+                                            part, "transcript", None
+                                        )
+                                        if text:
+                                            state.record_completion(role, text)
+                                            break
```
galkleinman: Should this capture all assistant messages or just the most recent one?

galkleinman (approved):
Would be great if we could take the opportunity to extract some of the nested event types handling into dedicated helper methods. up to you tho.

LGTM

packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_realtime_wrappers.py
EliJaghab: Just the most recent one, intentionally.

There are three event paths that can capture an assistant completion:

1. `history_added` -- fires when a single new item is added to history
2. `history_updated` -- fires when the full cumulative history changes (contains ALL messages from the entire session)
3. `response.done` (inside `raw_model_event`) -- fires when the model finishes responding

Whichever fires first creates the LLM span via `record_completion` -> `create_llm_span`. The `seen_completions` hash set in `record_completion` deduplicates, so subsequent handlers that try to record the same text become no-ops.

For `history_updated` specifically: the `history` list is cumulative -- on turn N it contains all messages from turns 1 through N. Using `reversed()` + `break` grabs only the newest assistant message (the one relevant to the current turn). Without this, we'd iterate through all previous assistant messages on every turn. The dedup set would ultimately prevent duplicate spans, but it's unnecessary work.

Here is the actual event ordering from an E2E test against the agents SDK relay (one-turn conversation, "Tell me about the iPhone 17"):

```
# Session starts -- history_updated fires with system message only, no assistant content yet
[OTEL DEBUG] traced_put_event CALLED, event_type=history_updated

# ... ~750 raw_model_event events (audio deltas, transcription deltas, etc.) ...

# Model finishes responding -- history_updated fires with full history including the new assistant message
[OTEL DEBUG] traced_put_event CALLED, event_type=history_updated
[OTEL DEBUG] record_completion called: role=assistant, content=The iPhone 17 is Apple's latest flagship phone...

# response.done fires next -- record_usage captures tokens, record_completion is deduplicated
[OTEL DEBUG] traced_put_event CALLED, event_type=raw_model_event
[OTEL DEBUG] record_usage called: usage type=dict
[OTEL DEBUG] record_completion called: role=assistant, content=The iPhone 17 is Apple's latest flagship phone...
[OTEL DEBUG] record_completion: duplicate, skipping

# Agent ends
[OTEL DEBUG] traced_put_event CALLED, event_type=agent_end
```

This shows: `history_updated` captures the completion first, then `response.done` captures usage and correctly skips the duplicate completion.

PR #3875, opened 2026-03-29
fix(sdk): prefer model_dump() over json() in JSONEncoder for Pydantic v2

### LINKED ISSUE #3516: 🐛 Bug Report: JSONEncoder calling deprecated method `json` for Pydantic models

Traceloop SDK

### 📜 Description

We recently turned on deprecation warnings for our app for another purpose, and began getting lots of warnings from Pydantic about Traceloop sdk calling the deprecated `json` method of Pydantic models. Presumably these models are arguments to workflows and tasks that we have set up.

```
.../miniconda3/envs/bb3.12/lib/python3.12/site-packages/traceloop/sdk/utils/json_encoder.py:21: PydanticDeprecatedSince20: The `json` method is deprecated; use `model_dump_json` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.11/migration/
  return o.json()
```

### 👟 Reproduction steps

This contrived example raises the deprecation warning in question:

```python
# make_warning.py
from traceloop.sdk.decorators import task
from pydantic import BaseModel

Traceloop.init()


class SomeValue(BaseModel):
    value: int


@task(name="add_numbers")
def add_numbers(a: SomeValue, b: SomeValue) -> SomeValue:
    return SomeValue(value=a.value + b.value)


def main():
    result = add_numbers(SomeValue(value=3), SomeValue(value=5))
    print(f"Result: {result.value}")


if __name__ == "__main__":
    main()
```

This script assumes the presence of a Traceloop API token in the env.

Call the script with `$ python -Wd make_warning.py`

### 👍 Expected behavior

JSONEncoder should call `model_dump_json()` method of pydantic classes if present. 

Deprecation warning is raised

Image

### 🤖 Python Version

3.12.11

_No response_

Narendra099:
## Summary
Resolves #3516

`JSONEncoder.default()` calls `o.json()` when serialising Pydantic model 
instances. In Pydantic v2, `.json()` is deprecated in favour of 
`.model_dump_json()`, causing `PydanticDeprecatedSince20` warnings.

## Root Cause
The `JSONEncoder` in `traceloop/sdk/utils/json_encoder.py` checked for 
`hasattr(o, "json")` and called `o.json()`. Pydantic v2 deprecated this 
method — it will be removed in Pydantic v3.

## Changes
- Check for `model_dump()` first (Pydantic v2), then `dict()` (Pydantic v1)
- Uses `model_dump()`/`dict()` (returning dicts) instead of 
  `model_dump_json()`/`json()` (returning strings) to avoid double-encoding 
  in `json.JSONEncoder.default()`
- Added 8 unit tests covering all serialization paths

## Why `model_dump()` and not `model_dump_json()`?
`json.JSONEncoder.default()` must return **Python objects** (dict, list, etc.), 
not JSON strings. Using `model_dump_json()` would return a string that gets 
double-encoded (e.g., `'"{\"value\": 42}"'` instead of `'{"value": 42}'`).

## Testing
- [x] 8 new unit tests pass (`test_json_encoder.py`)
- [x] Verified with real Pydantic v2 BaseModel — zero deprecation warnings
- [x] Backward compatible with Pydantic v1 (falls back to `dict()`)

Fixes #3516

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for JSON serialization compatibility with Pydantic models (versions 1 and 2)

* **Bug Fixes**
  * Updated JSON encoder to improve serialization handling for Pydantic model objects and dataclasses

### DIFF
--- a/packages/traceloop-sdk/tests/test_json_encoder.py
+++ b/packages/traceloop-sdk/tests/test_json_encoder.py
@@ -1,0 +1,128 @@
+"""Tests for JSONEncoder Pydantic v2 compatibility (issue #3516)."""
+
+import dataclasses
+import json
+import warnings
+import pytest
+
+from traceloop.sdk.utils.json_encoder import JSONEncoder
+
+
+class FakePydanticV2Model:
+    """Simulates a Pydantic v2 BaseModel with model_dump()."""
+
+    def __init__(self, value: int, name: str):
+        self.value = value
+        self.name = name
+
+    def model_dump(self):
+        return {"value": self.value, "name": self.name}
+
+    def model_dump_json(self):
+        return json.dumps({"value": self.value, "name": self.name})
+
+    def json(self):
+        warnings.warn(
+            "The `json` method is deprecated; use `model_dump_json` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return json.dumps({"value": self.value, "name": self.name})
+
+
+class FakePydanticV1Model:
+    """Simulates a Pydantic v1 BaseModel with dict() and json()."""
+
+    def __init__(self, value: int, name: str):
+        self.value = value
+        self.name = name
+
+    def dict(self):
+        return {"value": self.value, "name": self.name}
+
+    def json(self):
+        return json.dumps({"value": self.value, "name": self.name})
+
+
+@dataclasses.dataclass
+class SampleDataclass:
+    x: int
+    y: str
+
+
+class ObjectWithToJson:
+    def to_json(self):
+        return {"custom": "json"}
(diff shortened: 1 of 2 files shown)


dvirski:
Thanks for the PR! Closing in favor of #4098, which replaces the deprecated .json() / double-encoding .model_dump_json() calls in JSONEncoders across sdk, llamaindex, openai-agents, langchain, and anthropic with a model_dump() + dict() fallback chain plus 8 unit tests. Appreciate the report!

PR #3948, opened 2026-04-06
fix(google-genai): migrate span attrs to gen_ai.input/output.messages (fixes #3515)

### LINKED ISSUE #3515: 🐛 Bug Report: span attributes gen_ai.prompt and gen_ai.completion are deprecated in the latest OpenTelemetry Semantic Conventions

LLM Semantic Conventions

### 📜 Description

In the OpenTelemetry instrumentations in this repository, prompt contents are currently recorded under the following span attributes:

* `gen_ai.prompt` for request prompts
* `gen_ai.completion` for response outputs

However, in the latest (v1.38.0) OpenTelemetry [GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#deprecated-genai-attributes), both of these attributes are explicitly marked as deprecated and removed from the specification.

### 👟 Reproduction steps

Here is sample code using langchain and gemini.

Note that langchain and gemini are just example, and same issue can be said in another tools.

```python
from langchain_google_genai import ChatGoogleGenerativeAI
from langchain_core.messages import SystemMessage
from langchain_core.prompts import ChatPromptTemplate, HumanMessagePromptTemplate
from langchain_core.runnables import RunnablePassthrough


from opentelemetry import trace, _logs, metrics
# Trace
from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter

# LLM
from opentelemetry.instrumentation.langchain import LangchainInstrumentor
from opentelemetry.instrumentation.google_generativeai import GoogleGenerativeAiInstrumentor


def setUpOpenTelemetry():
    otlp_endpoint = "http://otel-collector:4317/v1"

    # Trace
    provider = TracerProvider()
    processor = BatchSpanProcessor(OTLPSpanExporter(
        endpoint=f'{otlp_endpoint}/traces', insecure=True))
    provider.add_span_processor(processor)

    LangchainInstrumentor().instrument()
    GoogleGenerativeAiInstrumentor().instrument()


setUpOpenTelemetry()

template = ChatPromptTemplate.from_messages([
    SystemMessage(content="""Answer the question based on the context below. If the question cannot be answered using the information provided, answer with "I don't know"."""),
    HumanMessagePromptTemplate.from_template("Context: {context}"),
    HumanMessagePromptTemplate.from_template("Questtion: {question}"),
])

model = ChatGoogleGenerativeAI(model="gemini-2.0-flash-lite")


ramen_chatbot = (
    {
        "question": RunnablePassthrough(),
        "context": lambda _: "The capital of France, Japan, and Germany are Paris, Tokyo, and Berlin respectively.",
    }
    | template
    | model
)
response = ramen_chatbot.invoke("What is the capital of Japan?")

print(response.text)
```

### 👍 Expected behavior

Although the specification refers to these deprecated attributes as "no replacement at this time", I believe the following fields provide a suitable alternative according to the latest GenAI semantic conventions:

* [`gen_ai.input.messages`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/?utm_source=chatgpt.com) for representing model input content
* [`gen_ai.output.messages`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/?utm_source=chatgpt.com) for representing model outputs
* [`gen_ai.system_instructions`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/?utm_source=chatgpt.com) for system- or model-level instructions

These attributes follow the new message-structured model introduced in the GenAI specification and align with the recommended approach for capturing prompt and response data, instead of using the deprecated gen_ai.prompt and gen_ai.completion span attributes.

---
I believe we can replace constants such as `GEN_AI_PROMPT` with an appropriate alternative like `GEN_AI_INPUT_MESSAGES`.

https://github.com/traceloop/openllmetry/blob/9d8cef6b7573b405f8a92ee74088716d160dbbb3/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py#L168-L172

In addition, we should follow the updated value format specified by the GenAI semantic conventions (e.g., for `gen_ai.output.messages`). The expected structure looks like:

```
[
  {
    “role”: “assistant”,
    “parts”: [
      {
        “type”: “text”,
        “content”: “The weather in Paris is currently rainy with a temperature of 57°F."
      }
    ],
    “finish_reason”: “stop”
  }
]
```

Span with deprecated attributes are emitted.

Image

### 🤖 Python Version

3.14.0

These fields were deprecated in Semantic Conventions v1.28.0 and in the following commit:

[32b75a8d465ddea8af396666cd4020c15f4859e1](https://github.com/open-telemetry/semantic-conventions/commit/32b75a8d465ddea8af396666cd4020c15f4859e1)

Although that commit introduced several event-based attributes such as `gen_ai.user.message`, these event attributes were later re-organized into a structured messaging model in another commit, which was included in v1.37.0 of the specification.

[3e06ddb5fc940eff38f66766372fc7b458d03906](https://github.com/open-telemetry/semantic-conventions/commit/3e06ddb5fc940eff38f66766372fc7b458d03906)

This newer version consolidates prompt, output, and instruction data into attributes like `gen_ai.input.messages`, `gen_ai.output.messages`, and `gen_ai.system_instructions`, replacing earlier experimental event fields.

---
There are a few existing issues in this repository that discuss OpenTelemetry SemConv compliance, but most of them take a broad or high-level perspective. None of them focus specifically on the deprecation of `gen_ai.prompt` and `gen_ai.completion` or the migration toward the new structured message attributes defined in the latest GenAI semantic conventions.

Because those issues do not address this concrete and well-scoped problem, opening this issue is still meaningful and necessary to ensure proper alignment with the current specification.

like

* https://github.com/traceloop/openllmetry/issues/1455

None

abhyudayareddy:
## Summary

Fixes #3515.

Migrates `google_generativeai/span_utils.py` from deprecated indexed span attributes to the stable OTel GenAI Semantic Conventions:

- `gen_ai.prompt.{N}.content` / `gen_ai.prompt.{N}.role` -> `gen_ai.input.messages` (JSON array)
- `gen_ai.completion.{N}.content` / `gen_ai.completion.{N}.role` -> `gen_ai.output.messages` (JSON array)

The new attributes store a JSON-serialised list of `{"role": "...", "parts": [...]}` objects, consistent with the format used by the OpenAI instrumentation in this repo.

## Changes

- **`span_utils.py`**: Removed `_set_prompt_attributes`; `set_input_attributes` and `set_input_attributes_sync` now build an `input_messages` list and write it to `gen_ai.input.messages`; `set_response_attributes` builds an `output_messages` list and writes it to `gen_ai.output.messages`.
- **`tests/test_generate_content.py`**: Updated assertions to parse the new JSON blob attributes instead of the old indexed ones.

## Checklist

- [x] PR name follows conventional commits format: `fix(instrumentation): ...`

## Summary by CodeRabbit

* **Refactor**
  * Streamlined input/output handling: inputs are normalized to aggregated JSON messages; inline binary images are base64-encoded and represented as image_url objects for both sync/async paths. Response serialization unified (clear streaming vs non-streaming behavior, finish-reason handling), presence/frequency penalties remapped to LLM_* attributes, and deprecated per-index response attributes removed.

* **Tests**
  * Updated to validate aggregated JSON message attributes (roles and part shapes), assert total token usage, and ensure deprecated per-index prompt/completion attributes are not emitted.


abhyudayareddy:
Addressed both CodeRabbit review items in the latest two commits:

1. **`finish_reason` fix** (4862ab3): replaced the hardcoded `"finish_reason": "stop"` with dynamic extraction from `response.candidates[0].finish_reason` for non-streaming responses (using `.name.lower()` on the enum), and omitted `finish_reason` entirely for streaming paths where it isn't reliably available per-chunk.

2. **Stronger test assertions** (e6f86a8): added `parts` structure validation for both input and output messages — checks that `parts` is a non-empty list, that `parts[0]["type"] == "text"`, and that `"content"` is present.

abhyudayareddy:
Addressed all CodeRabbit round-2 feedback (commits 08f2f74 and 56f5e51):

**test_generate_content.py** (commit 08f2f74): Added negative assertions after the existing `gen_ai.input.messages` / `gen_ai.output.messages` checks to guard against regression to the deprecated indexed attributes:
```python
assert not any(key.startswith("gen_ai.prompt.") for key in attrs.keys())
assert not any(key.startswith("gen_ai.completion.") for key in attrs.keys())
```

**span_utils.py** (commit 56f5e51): Fixed `contents` normalization to correctly distinguish between input shapes:
- `str` → single user message (unchanged)
- `Content` object (has `.parts`): single turn, wrapped as one message
- `list[Content]` (each item has `.parts`): multi-turn conversation, each item becomes a separate message with its own role
- `list[Part]` or `list[str]`: single-turn, all items collected into one `role: user` message with multiple parts

The fix applies to both the async `set_input_attributes` and the sync `set_input_attributes_sync` paths.

abhyudayareddy:
Addressed CodeRabbit round-3 feedback in commit 6dc73dc:

**Item 1 (Major — Handle all response.parts types):** Added a `_serialize_response_part(part)` helper that serializes text parts, function_call parts (with name + args), and falls back to `str(part)` for other types. `set_response_attributes` now iterates over `response.candidates`, reads each candidate's `content.parts`, and serializes all parts through this helper — so tool-call and multimodal responses are captured in `gen_ai.output.messages` instead of being dropped.

**Item 2 (Minor — Per-candidate finish_reason):** Removed the single shared `_finish_reason` extracted from `candidates[0]`. Each candidate now reads its own `finish_reason` field inside the loop, so multi-candidate responses get the correct stop reason per message instead of all inheriting the first candidate's reason.

abhyudayareddy:
@coderabbitai resume

abhyudayareddy:
👋 Hi @nirga, @galkleinman, @dinmukhamedm — could you take a look at this PR when you get a chance? It migrates the Google GenAI instrumentation span attributes to the stable `gen_ai.input/output.messages` convention (fixes #3515). CodeRabbit review has been resumed and all conflicts are resolved. Thanks!

abhyudayareddy:
Fixed the Major issue flagged by CodeRabbit. The `_finish_reason` extraction in the non-streaming output path was using `fr.name.lower()` directly, which produces raw Gemini enum names like `"max_tokens"` and `"safety"` instead of the OTel GenAI semantic convention values (`"length"` and `"content_filter"` respectively).

The fix routes through the existing `_map_gemini_finish_reason()` helper that already has the correct STOP→stop, MAX_TOKENS→length, SAFETY/RECITATION/BLOCKLIST→content_filter, MALFORMED_FUNCTION_CALL→error mappings.

Commit: fix(google-genai): use _map_gemini_finish_reason() for OTel-compliant finish_reason values

abhyudayareddy:
Addressed the two Critical issues flagged by CodeRabbit:

**Fix 1 — `span_utils.py` line 244: NameError + undefined function**

The original line called `parts_acc.extend(await _parts_from_genai_part_async(...))` which had two problems:
- `parts_acc` is not defined in `_process_content_item` — the list is called `processed_content`
- - `_parts_from_genai_part_async` is never defined anywhere in the module
Replaced with the correct pattern using the existing `_process_content_part` function (which returns a single dict, not a list):
```python
result = await _process_content_part(part, span, part_index)
if result:
    processed_content.append(result)
```

**Fix 2 — `test_generate_content.py` line 76: legacy token attribute**

The test was asserting against `SpanAttributes.LLM_USAGE_TOTAL_TOKENS` (deprecated) while the implementation sets `SpanAttributes.GEN_AI_USAGE_TOTAL_TOKENS`. Aligned the test to match the implementation.

abhyudayareddy:
Hi team — just flagging that the failing `security/snyk` check appears to be a repository quota/infrastructure issue ("You have used your limit of pr...") rather than a code problem in this PR. The other CI checks are passing. Would appreciate a human review when you get a chance. Thanks!

abhyudayareddy:
Addressing all CodeRabbit inline review suggestions below. No human maintainer comments exist yet. Snyk failure on Apr 19 was a quota/rate-limit issue (not a vulnerability in this PR's code) — confirmed by CodeRabbit's pre-merge check status showing 4 pass / 1 fail which matches the known Snyk quota problem.

---

## CodeRabbit suggestion assessment

### CR-1 (Major): `response.text` fallback inside candidate loop can bleed across candidates

**Valid. This is a real semantic bug.**

When `candidate.content` is missing or `content.parts` is empty, the current fallback reaches for `response.text` — but that is a response-level property, not per-candidate. In a multi-candidate response, any candidate with empty parts will inherit whatever text happens to be in `response.text`, which could be text from a completely different candidate. This silently produces wrong span data for multi-candidate calls.

**Fix:** Guard the fallback with `if idx == 0:` or, better, scope it outside the candidate loop so it only fires when no candidates produced output at all. Do not use `response.text` as a per-candidate fallback.

---

### CR-2 (Major): `for item in response:` should be `for item in response.text:`

**Valid. This is a definite bug.**

Line 679 correctly checks `isinstance(response.text, list)`, but line 680 then iterates `for item in response:` — iterating the `GenerateContentResponse` object itself, not the list `response.text`. The Google GenAI SDK's response object may not be iterable in this way, and even if it is, iterating it does not yield text items. The fix is a one-character change: `for item in response.text:`.

---

### CR-3 (Critical): `_GCP_GEN_AI` and `_GEN_CONTENT` used at line 750 but not imported in `span_utils.py`

**Valid. This would cause a `NameError` at runtime.**

These constants are defined in `__init__.py` (the instrumentation entrypoint) but `span_utils.py` does not import them. The proposed fix in CodeRabbit's earlier review rounds was incomplete because it referenced the constants without adding the import. Additionally, line 750 uses the hardcoded string `"Google"` while line 740 uses the constant `_GCP_GEN_AI` — the two should be consistent.

**Fix:** Either import `_GCP_GEN_AI` and `_GEN_CONTENT` from the appropriate module into `span_utils.py`, or inline the string literals consistently (and remove the dead constant reference). Do not mix constant references with hardcoded strings for the same logical value.

---

### CR-4 (Critical): Missing opening parenthesis on import at line 14

**Valid. This is a `SyntaxError` that prevents the module from loading at all.**

```python
# Current (broken):
from opentelemetry.semconv_ai import 
    SpanAttributes,

# Fixed:
from opentelemetry.semconv_ai import (
    SpanAttributes,
)
```

This would cause an immediate `SyntaxError` on import, meaning none of the instrumentation works. This must be fixed before any other issue matters.

---

## Summary of required actions

| # | Severity | Issue | Action |
|---|----------|-------|--------|
| CR-4 | Critical | Missing `(` in import on line 14 | Fix syntax error |
| CR-3 | Critical | `_GCP_GEN_AI`/`_GEN_CONTENT` used but not imported | Add import or inline literal consistently |
| CR-1 | Major | `response.text` fallback bleeds across candidates in loop | Scope fallback outside candidate loop |
| CR-2 | Major | `for item in response:` should be `for item in response.text:` | Fix iteration target |

CR-4 and CR-3 are blockers — the module cannot load with a syntax error, and the token histogram code will throw `NameError` at runtime if those constants are not in scope. CR-1 and CR-2 affect correctness of span data for streaming/multi-candidate responses.

The core `gen_ai.input.messages` / `gen_ai.output.messages` migration (the primary purpose of this PR) is correct: using a JSON-serialized list of `{"role": "...", "parts": [...]}` objects aligns with the OTel GenAI semantic conventions spec and is consistent with the OpenAI instrumentation in this repo. The attribute naming itself does not need changes.

abhyudayareddy:
## Code Review: 4 Critical Bugs in `span_utils.py`

Reviewing the diff in `packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/span_utils.py` — found four bugs that will cause runtime failures.

---

### CR-4 (SyntaxError) — Missing `(` in multi-line import, line 14

**File:** `span_utils.py`  
**Line:** 14

```python
# BEFORE (broken — SyntaxError: invalid syntax)
from opentelemetry.semconv_ai import 
    SpanAttributes,
)

# AFTER (fixed)
from opentelemetry.semconv_ai import (
    SpanAttributes,
)
```

The opening parenthesis for the multi-line import was dropped. This is a hard `SyntaxError` — the module cannot be imported at all.

---

### CR-3 (NameError) — `_GCP_GEN_AI` / `_GEN_CONTENT` used but never defined; `"Google"` hardcoded inconsistently

**File:** `span_utils.py`  
**Lines:** 19–20 (missing module-level constants), 742–743, 752

The two module-level constants that existed in `main` were dropped in this PR:

```python
# MISSING — add back after the imports (before the logger line)
_GCP_GEN_AI = GenAIAttributes.GenAiProviderNameValues.GCP_GEN_AI.value
_GEN_CONTENT = GenAIAttributes.GenAiOperationNameValues.GENERATE_CONTENT.value
```

Lines 742–743 reference these undeclared names → instant `NameError` at runtime when `token_histogram` is truthy.

Line 752 hardcodes the string `"Google"` instead of using the constant — inconsistent with lines 742–743:

```python
# BEFORE (line 752 — inconsistent hardcoded string)
                GenAIAttributes.GEN_AI_PROVIDER_NAME: "Google",

# AFTER (use the same constant)
                GenAIAttributes.GEN_AI_PROVIDER_NAME: _GCP_GEN_AI,
```

---

### CR-2 (wrong iteration) — `for item in response:` should be `for item in response.text:`

**File:** `span_utils.py`  
**Line:** 682

```python
# BEFORE (iterates over the response object itself — wrong)
                    for item in response:

# AFTER (iterates over the list returned by response.text)
                    for item in response.text:
```

The guard on line 681 already checks `isinstance(response.text, list)`, so the loop must iterate over `response.text`, not the raw `response` object. Iterating over `response` will either raise a `TypeError` or silently produce wrong data.

---

### CR-1 (candidate bleed) — `response.text` fallback is inside the `for candidate` loop

**File:** `span_utils.py`  
**Lines:** 665–672

The `response.text` fallback sits inside `for candidate in _candidates:` (lines 655–677). When a candidate has no `content.parts`, it appends `response.text` once **per candidate** — duplicating the same output message N times.

```python
# BEFORE — fallback inside the loop (causes N-candidate bleed)
            for candidate in _candidates:
                ...
                if not parts:
                    try:
                        text = response.text
                        if text:
                            parts = [{"type": "text", "content": text}]
                    except Exception:
                        pass
                if parts:
                    output_messages.append(...)

# AFTER — remove the fallback from inside the loop; add it once after the loop
            for candidate in _candidates:
                ...
                # (remove the response.text fallback block here)
                if parts:
                    output_messages.append(...)

            # Fallback: only fires if no candidate yielded any parts at all
            if not output_messages:
                try:
                    text = response.text
                    if text:
                        output_messages.append({
                            "role": "assistant",
                            "parts": [{"type": "text", "content": text}],
                        })
                except Exception:
                    pass
```

---

**Summary:**

| ID | Line(s) | Issue | Severity |
|----|---------|-------|----------|
| CR-4 | 14 | Missing `(` in `from opentelemetry.semconv_ai import` | Critical — `SyntaxError`, module unimportable |
| CR-3 | 19–20 (missing), 752 | `_GCP_GEN_AI`/`_GEN_CONTENT` not defined; `"Google"` hardcoded | Critical — `NameError` at runtime |
| CR-2 | 682 | `for item in response:` instead of `for item in response.text:` | High — `TypeError` or wrong output |
| CR-1 | 665–672 | `response.text` fallback inside candidate loop | Medium — duplicated output messages |

abhyudayareddy:
Fixed all 4 issues identified in code review:
- **CR-4**: Added missing `(` in multi-line import (was causing `SyntaxError` on module load)
- **CR-3**: Added `_GCP_GEN_AI` and `_GEN_CONTENT` module-level constants; replaced hardcoded `"Google"` with `_GCP_GEN_AI` for consistency
- **CR-2**: Fixed loop variable — `for item in response:` → `for item in response.text:`
- **CR-1**: Moved `response.text` fallback outside the `for candidate in _candidates:` loop to prevent cross-candidate bleed

All fixes committed. Happy to iterate if anything needs adjustment.

abhyudayareddy:
Hi @traceloop/opentelemetry-instrumentation-google-generativeai-maintainers — just pushed a follow-up commit ([27fdf93](https://github.com/abhyudayareddy/openllmetry/commit/27fdf93)) that resolves the remaining runtime errors flagged by CodeRabbit:

**Fixed:**
- `NameError: name 'um' is not defined` in `set_model_response_attributes` — `um = response.usage_metadata` was missing before all four `um.xxx` references, crashing every non-streaming call.
- `ImportError` on module load — `_collect_finish_reasons_from_response` was imported by `__init__.py` but absent from `span_utils.py`. Implemented: returns one mapped finish-reason string per candidate, preserving 1:1 alignment.
- `NameError: _parts_from_genai_part_sync` — called in `_system_instruction_to_parts` but never defined. Implemented with thinking/reasoning block support (`part.thought=True` → `{"type":"reasoning",...}`).
- `AttributeError: 'str' object has no attribute 'text'` in the `response.text` list fallback — replaced with a unified `_output_messages_from_generate_response` helper that always populates `finish_reason` (empty string when unknown).
- Streaming path now includes `finish_reason` in every emitted message and suppresses empty-string messages when no `stream_last_chunk` context exists.
- `set_model_response_attributes` now emits `gen_ai.response.finish_reasons` only when at least one reason is non-empty, using `stream_finish_reasons` or `_collect_finish_reasons_from_response` as fallback.

All cases covered by the new `test_finish_reasons.py` suite. CI should be green once the Snyk infra quota is restored.

abhyudayareddy:
Addressing the new CodeRabbit suggestions from the Apr 26 review (commit 27fdf93).

---

## New CodeRabbit findings assessment

### Inline — Major: `item.text` AttributeError when `response.text` is `list[str]`

**Valid. Real runtime bug.**

When `isinstance(response.text, list)` is true the loop runs `for item in response.text:`, but the body then accesses `item.text`. Since `response.text` yields plain `str` objects, `item.text` raises `AttributeError` on every iteration. The exception is silently swallowed by `@dont_throw`, so the entire `gen_ai.output.messages` attribute is dropped for any response where `response.text` is a list.

Fix: replace `item.text` with `item` directly (items are already strings).

Will fix in next commit.

---

### Outside-diff — Critical: `um` undefined in `set_model_response_attributes`

**Valid. Real runtime bug.**

`um = response.usage_metadata` was lost during the refactor. All four `um.xxx` accesses inside the `if hasattr(response, "usage_metadata"):` block raise `NameError`, silently dropping all token metrics for non-streaming responses via `@dont_throw`.

Fix: add `um = response.usage_metadata` immediately after the `if hasattr(...)` guard.

Will fix in the same commit.

---

### Inline — Minor: image part format inconsistency (`_process_image_part` vs `_otel_image_part_from_genai_part`)

**Valid style/consistency issue — not a runtime bug.**

`_process_image_part` deliberately returns OpenAI-compatible format (`{"type": "image_url", ...}`) for cross-provider consistency. The OTel-spec helpers (`_otel_image_part_from_genai_part`) are dead code left over from an earlier iteration.

Will document the intentional OpenAI-compatible shape with a comment at the function site and remove the dead OTel helpers in the same commit.

---

Pushing a follow-up commit to fix the two runtime bugs (Major `item.text` + Critical `um` NameError) and clean up the dead helpers.

abhyudayareddy:
Hi @traceloop/maintainers — all four CodeRabbit issues (CR-4 SyntaxError, CR-3 missing constants, CR-2 wrong loop variable, CR-1 candidate bleed) are now fixed and committed. The branch is out of date with main but changes can be cleanly merged. Would appreciate a first human review when you get a chance. Thanks!

abhyudayareddy:
Friendly ping @traceloop/maintainers — this PR has been open for a while and CodeRabbit has completed its review. All inline review threads have been addressed. Would love a human review when you get a chance. Happy to make any further changes needed.

abhyudayareddy:
Rebased this branch onto the latest `main` — it had drifted out of date (`mergeStateStatus: DIRTY`) since main picked up cache-token-count reporting for this instrumentation (#4240) after this branch's last sync.

Conflict was isolated to `span_utils.py`'s `set_model_response_attributes`: main added emission of `GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS` from `usage_metadata.cached_content_token_count`, while this branch had its own fix for the `um` NameError in the same block. Merged both — `um` is now initialized once up front, the cache-token emission from main is preserved, and the token-histogram guard uses `um is not None` so it can't NameError if `usage_metadata` is ever absent.

While verifying with the local test suite, also caught and fixed a real regression from this branch's own history: the output-token histogram record call had lost its `gen_ai.operation.name` / `gen_ai.request.model` attributes somewhere along the way (present on the input-token record, silently dropped on the output one), which was failing `test_generate_metrics`. Restored those to match the input record and upstream's version — full suite is green again except for one pre-existing, merge-unrelated failure (`test_async_input.py::test_list_of_strings`) that predates this rebase: the test expects a list of input strings to become two separate messages, but the current code (correctly, per Google's `contents` semantics) treats a string list as multiple parts of a single turn. Flagging it separately rather than guessing at intent — happy to update the test's expectation if that's confirmed as the right call.

Branch is up to date with `main` and conflict-free now.

abhyudayareddy:
cc @nirga @galkleinman @dinmukhamedm — rebased this onto current main since it had drifted (conflict resolution + two regressions the merge would've silently introduced are detailed above). It's conflict-free and green now. No rush, just flagging it's current whenever you get a chance to take a first look.

abhyudayareddy:
@nirga @galkleinman @dinmukhamedm — friendly ping. All CodeRabbit review threads are resolved, CI is green, and it's still conflict-free against current `main`. Would appreciate a first human review whenever you get a chance. Thanks!

packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/span_utils.py
```diff
@@ -708,24 +767,10 @@ def set_model_response_attributes(
...
-    else:
-        reasons = _collect_finish_reasons_from_response(response)
-    if reasons and any(reason != "" for reason in reasons):
-        _set_span_attribute(
-            span, GenAIAttributes.GEN_AI_RESPONSE_FINISH_REASONS, reasons
-        )
-
-    um = getattr(response, "usage_metadata", None)
-    if um is not None:
+    um = None
```
RichardoMrMu, line 776: **[P2] Preserve `gen_ai.response.id` during the semconv migration**

`set_model_response_attributes()` previously guardedly copied `response.response_id` to `GenAIAttributes.GEN_AI_RESPONSE_ID`, but this patch removes that write while continuing to emit the other response metadata. `GenerateContentResponse` exposes `response_id`, and the current OTel GenAI client-span conventions list `gen_ai.response.id` as Recommended. This silently drops the provider-issued response identifier from every response that supplies one, outside the prompt/completion migration requested by #3515. Please restore the guarded assignment and add a regression assertion next to the new output-message assertion; the existing VCR response already contains a `responseId`.

abhyudayareddy: Good catch — fixed in 3317023e. Restored the guarded `response.response_id` → `GEN_AI_RESPONSE_ID` assignment in `set_model_response_attributes()` right after `GEN_AI_RESPONSE_MODEL`. The `getattr(response, "response_id", None)` guard covers both the non-streaming response and the streaming last-chunk path (both `GenerateContentResponse` objects expose it). Added a regression assertion in `test_client_spans` against the `responseId` (`-hk4afOSMZKkjuMPnJWGkAk`) already in the VCR cassette.

abhyudayareddy:
Pushed two commits:

- `3317023e` — restores `gen_ai.response.id` (addresses @RichardoMrMu's review; guarded `response.response_id` copy + regression assertion against the cassette's `responseId`).
- `bcdcefd9` — fixes `test_async_input.py::test_list_of_strings`, which expected a bare `contents=["first", "second"]` list to produce two messages. `google.genai._transformers.t_contents` groups a bare string list into one user turn with multiple parts, which is what both the sync and async input paths already emit — the test now asserts that shape.

Full `pytest tests/` (61) + `ruff` green locally.

packages/opentelemetry-instrumentation-google-generativeai/opentelemetry/instrumentation/google_generativeai/span_utils.py
RichardoMrMu: Thanks — verified on live head `bcdcefd9` that `3317023e` restores the guarded `response.response_id` → `GenAIAttributes.GEN_AI_RESPONSE_ID` assignment and adds the VCR-backed regression assertion. This fully addresses the finding.

abhyudayareddy:
cc @nirga @galkleinman @dinmukhamedm — @RichardoMrMu reviewed and confirmed the latest fix a couple weeks back, and it's been quiet since. All threads resolved, CI green, conflict-free against main. Would appreciate a maintainer approval when you get a chance. Thanks!

PR #3976, opened 2026-04-11
fix(anthropic): always set streaming token usage from API data (#3949)

carsonjc04:
The streaming _handle_completion() and _complete_instrumentation() methods gated all token-usage recording behind
`if Config.enrich_token_usage`. Since that flag defaults to False, streaming spans never received gen_ai.usage.input_tokens — even when the API provided real usage data in the SSE stream events.

Downstream tools (e.g. Langfuse) would then fall back to tokenizing the raw input content, which includes base64 image data, producing massively inflated token counts (e.g. 1,633 instead of 343).

Restructure the logic so that:
- API-provided usage is always read and recorded on the span
- The enrich_token_usage flag only gates the local estimation fallback (count_prompt_tokens_from_request) for when usage is absent from the response

This aligns the streaming path with the non-streaming create() path, which already sets token attributes unconditionally.

Closes #3949

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming token usage now prefers API-provided values; local estimation is used only as a fallback when enrichment is enabled. Completion counts are normalized, instrumentation completion is reliably marked once for both sync and async streams, and token-usage warning logs now include exception text.

* **Tests**
  * Added tests validating token-usage attributes and span completion behavior for streaming (sync and async) when API usage is present or absent.


packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -137,6 +137,35 @@ def _set_token_usage(
             )
 
 
+def _resolve_stream_token_usage(complete_response, instance, kwargs):
+    """Resolve prompt and completion token counts from a streaming response.
+
+    Returns (prompt_tokens, completion_tokens) when token data is available,
```
max-deygin-traceloop:
```suggestion
    Returns (input_tokens, output_tokens) when token data is available,
```

max-deygin-traceloop: Inconsistency in dockstring

max-deygin-traceloop (changes requested):
Minor dock string nitpick, otherwise LGTM

max-deygin-traceloop:
LGTM @carsonjc04 if you don't mind, please [sign the CLA](https://github.com/traceloop/openllmetry/pull/3976#issuecomment-4230335831) and we can merge it

max-deygin-traceloop (approved):
LGTM!

PR #4000, opened 2026-04-16
fix(langchain): Use instance-level wrapping for AgentMiddleware hooks

MaxStenklyft:
## Summary

`_wrap_middleware_hooks` uses `wrapt.wrap_function_wrapper` on `AgentMiddleware` base class methods (`before_model`, `after_model`, etc.). This replaces them with `FunctionWrapper` descriptors, which return a new `BoundFunctionWrapper` on every attribute access. This breaks the identity checks in LangGraph's `create_agent` (`factory.py`):

```python
if m.__class__.before_agent is not AgentMiddleware.before_agent:
    graph.add_node(...)
```

The `is not` check always evaluates to `True` after wrapping, causing LangGraph to add graph nodes for every middleware hook regardless of whether it's actually overridden. This scales graph size linearly with the number of hooks per middleware, easily exceeding recursion limits.

## Fix

Replace class-level wrapping with instance-level wrapping via `AgentMiddleware.__init__`. The wrapper patches hook methods on each instance after construction, leaving base class methods untouched so identity checks work correctly.

## Tests

- Added `test_middleware_identity.py` with two tests verifying base class identity is preserved and instance hooks are instrumented
- Updated `test_middleware_super_call_succeeds_despite_outer_failure` to match new instance-level wrapping semantics
- Full test suite passes (141 passed, 20 skipped, 0 failures)

## Summary by CodeRabbit

* **Bug Fixes**
  * Middleware instrumentation now applies at the instance level so telemetry spans reflect full execution outcomes (including subclass-raised failures) and instance-level patches are properly removed on uninstrument.

* **Tests**
  * New and updated tests cover hook identity, instance-level instrumentation behavior, correct failure status in spans, and cleanup after uninstrument.

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/__init__.py
@@ -271,30 +271,47 @@ class LangchainInstrumentor(BaseInstrumentor):
                 logger.debug("Failed to wrap langchain.agents.create_agent: %s", e)
 
     def _wrap_middleware_hooks(self, tracer):
-        """Wrap AgentMiddleware hook methods for instrumentation."""
-        # Sync hooks
-        sync_hooks = ["before_model", "after_model", "before_agent", "after_agent"]
-        for hook_name in sync_hooks:
-            try:
-                wrap_function_wrapper(
-                    module="langchain.agents.middleware.types",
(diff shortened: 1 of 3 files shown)


MaxStenklyft:
Noticed this when upgraded to langchain 1.2. We started hitting GRAPH_RECURSION_LIMIT due to no-op nodes introduced by this package. This change should eliminate the no-op nodes by fixing the identity check.

PR #4051, opened 2026-04-24
fix(anthropic): prevent KeyError in _process_response_item during streaming

### LINKED ISSUE #4050: 🐛 Bug: KeyError: 'input' in _process_response_item during Anthropic streaming
## Description

`_process_response_item` in the Anthropic streaming instrumentation raises `KeyError: 'input'` during streaming when a `content_block_delta` with `input_json_delta` type arrives for a content block that isn't `tool_use`.

## Root Cause

In `streaming.py`, the `content_block_start` handler only sets the `"input"` key on the event dict for `tool_use` content blocks (line 48):

```python
if item.content_block.type == "tool_use":
    complete_response["events"][index]["input"] = """"""
```

But the `content_block_delta` handler accesses `["input"]` unconditionally (line 56):

```python
elif item.delta.type == "input_json_delta":
    complete_response["events"][index]["input"] += item.delta.partial_json
```

This bare dict subscription raises `KeyError: 'input'` when the key doesn't exist.

Similarly, `complete_response["events"][index]["text"]` (lines 52-54) uses bare `+=` which would fail if `"text"` were ever missing from the event dict.

## Impact

While the function is decorated with `@dont_throw`, the exception still triggers `traceback.format_exc()` output to stderr on **every streaming chunk**. In our production environment (FastAPI + K8s), this tight loop of stderr output destabilized all pods and caused a crash loop affecting all users.

## Reproduction

1. Use `AnthropicInstrumentor` with a `TracerProvider`
2. Send a streaming request to Anthropic that produces `input_json_delta` events for a content block where the `content_block_start` wasn't processed (e.g., due to timing or event ordering)
3. Observe `KeyError: 'input'` on stderr for every delta event

## Fix

PR submitted: use `.get(key, "")` instead of bare `[key]` and add bounds-checking on the events list index.

## Environment

- Python 3.11
- Anthropic SDK (latest)

orihab-placer:
Fixes #4050

## Summary

`_process_response_item` in the Anthropic streaming instrumentation crashes with `KeyError: 'input'` because the `content_block_delta` handler accesses `complete_response["events"][index]["input"]` with a bare dict subscription, but the `"input"` key is only set for `tool_use` content blocks in the `content_block_start` handler.

## Changes

- Add bounds-check on `index` before accessing the events list (prevents `IndexError`)
- Use `.get(key, "")` instead of bare `[key]` for all dict accesses in the `content_block_delta` handler
- Use local `event` variable to avoid repeated list indexing

## Impact

This bug causes `KeyError: 'input'` on every streaming chunk. While `@dont_throw` catches the exception, `traceback.format_exc()` still writes to stderr in a tight loop, destabilizing production pods (we experienced a full crash loop across all K8s pods in production).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming delta processing to properly handle edge cases and prevent errors during content accumulation.

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
@@ -48,12 +48,14 @@ def _process_response_item(item, complete_response):
                 complete_response["events"][index]["input"] = """"""
     elif item.type == "content_block_delta":
         index = item.index
-        if item.delta.type == "thinking_delta":
-            complete_response["events"][index]["text"] += item.delta.thinking
-        elif item.delta.type == "text_delta":
-            complete_response["events"][index]["text"] += item.delta.text
-        elif item.delta.type == "input_json_delta":
-            complete_response["events"][index]["input"] += item.delta.partial_json
+        if index < len(complete_response.get("events", [])):
+            event = complete_response["events"][index]
+            if item.delta.type == "thinking_delta":
+                event["text"] = event.get("text", "") + item.delta.thinking
+            elif item.delta.type == "text_delta":
+                event["text"] = event.get("text", "") + item.delta.text
+            elif item.delta.type == "input_json_delta":
+                event["input"] = event.get("input", "") + item.delta.partial_json
     elif item.type == "message_delta":
         for event in complete_response.get("events", []):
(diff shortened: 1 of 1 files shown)


dvirski:
Thanks for the PR! Closing in favor of #4079, which adds the same bounds check plus .get() in _process_response_item along with two regression tests. CLA was blocking merge here. Really appreciate the report and the fix.

PR #4053, opened 2026-04-24
fix(anthropic): replace blocking thread.join() with asyncio.ensure_future in run_async

### LINKED ISSUE #4052: 🐛 Bug: run_async() blocks async event loop with thread.join() — causes liveness probe failures
## Description

`run_async()` in `utils.py` uses `threading.Thread` + `thread.join()` to run async functions from sync wrappers. When called from within an async event loop (e.g. FastAPI), `thread.join()` blocks the entire event loop thread, preventing any other coroutines from executing.

## Root Cause

```python
# utils.py:274
def run_async(method):
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None

    if loop and loop.is_running():
        thread = threading.Thread(target=lambda: asyncio.run(method))
        thread.start()
        thread.join()  # ← blocks the event loop thread
    else:
        asyncio.run(method)
```

This is called from `_handle_input()` which runs inside the sync `_wrap` wrapper — including for `AsyncMessages.stream()` which is wrapped as a sync method (see `WRAPPED_METHODS` comment: "this method is on an async resource, but is meant to be called as an async context manager").

## Impact

In async frameworks (FastAPI, Starlette, etc.):

1. Every `messages.stream()` or `messages.create()` call blocks the event loop for the duration of `aset_input_attributes` (serializing all messages + tools to span attributes)
2. When tool calls trigger nested Anthropic API calls (e.g. a tool handler that calls `messages.create()` internally), the blocking compounds — multiple sequential `thread.join()` calls
3. Health check endpoints can't respond while the event loop is blocked
4. In Kubernetes, liveness probe timeouts → pod restarts → crash loop

In our production environment (FastAPI + K8s), this caused pod crash loops whenever a conversation triggered tool calls with nested Anthropic API calls. The k8s liveness probe (`/health/live`, 2s timeout, 3 failures = kill) couldn't be served while `thread.join()` blocked the event loop.

## Reproduction

1. Use `AnthropicInstrumentor` with `AsyncAnthropic` in a FastAPI app
2. Configure a k8s liveness probe with a short timeout (2s)
3. Send a message that triggers a tool call where the tool handler also calls the Anthropic API
4. The event loop blocks during both the outer `messages.stream()` and the inner `messages.create()` — liveness probe fails

## Suggested Fix

Replace `thread.join()` with `asyncio.ensure_future()` when a running event loop is detected:

```python
def run_async(method):
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None

    if loop and loop.is_running():
        asyncio.ensure_future(method)  # non-blocking
    else:
        asyncio.run(method)
```

This schedules the span attribute setting as a background task on the event loop. The attributes may be set slightly after the span starts, but this is a minor telemetry tradeoff vs blocking the event loop.

PR submitted with the fix.

## Related

- #4050 — `KeyError: 'input'` in `_process_response_item` (separate bug, same package, same production incident)

## Environment

- Python 3.11
- FastAPI + uvicorn (async)
- Kubernetes with liveness probes

orihab-placer:
Fixes #4052

## Summary

`run_async()` uses `threading.Thread` + `thread.join()` to bridge sync/async code. When called from within a running async event loop (FastAPI, Starlette, etc.), `thread.join()` blocks the entire event loop thread, preventing health checks and other coroutines from executing.

## Changes

Replace `thread.join()` with `asyncio.ensure_future()` when a running event loop is detected. This schedules the span attribute setting as a non-blocking background task.

```python
# Before (blocks event loop):
thread = threading.Thread(target=lambda: asyncio.run(method))
thread.start()
thread.join()

# After (non-blocking):
asyncio.ensure_future(method)
```

## Impact

In our production environment (FastAPI + Kubernetes), `thread.join()` caused pod crash loops:
1. Every `messages.stream()` / `messages.create()` call blocks the event loop
2. Tool calls with nested Anthropic API calls compound the blocking
3. K8s liveness probe can't be served → pod killed after 3 consecutive timeouts

## Related

- #4050 / #4051 — `KeyError: 'input'` in `_process_response_item` (separate bug, same incident)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of asynchronous operations when invoked from within an existing event loop, eliminating unnecessary thread overhead and optimizing coroutine scheduling behavior.

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
@@ -278,9 +278,7 @@ def run_async(method):
         loop = None
 
     if loop and loop.is_running():
-        thread = threading.Thread(target=lambda: asyncio.run(method))
-        thread.start()
-        thread.join()
+        asyncio.ensure_future(method)
     else:
         asyncio.run(method)
 


orihab-placer:
Closing — the underlying issue (#4052) was misdiagnosed. The root cause was in our code, not in `run_async()`. See #4052 for details.

PR #4061, opened 2026-04-28
feat(openai-agents): record cache_read_input_tokens and reasoning_tokens

### LINKED ISSUE #4060: 🐛 Bug Report: openai-agents instrumentation doesn't record cached input tokens
## 📜 Description

The `opentelemetry-instrumentation-openai-agents` package captures `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens` on LLM spans but never records cached input tokens. This is notable because the underlying OpenAI Agents SDK `Usage` object *does* carry this information via `usage.input_tokens_details.cached_tokens` (and `cached_tokens` is a documented field on OpenAI's `Response.usage`).

The raw-OpenAI instrumentor (`opentelemetry-instrumentation-openai`) correctly records `gen_ai.usage.cache_read_input_tokens`, so when both instrumentors are installed the cache data shows up on the raw-client spans. But anyone using the Agents SDK instrumentor alone (e.g. to avoid double-instrumentation artifacts from two spans per LLM call) has no cache visibility — even though the information is present on every Response.

Downstream impact: Braintrust (and any other backend reading the OTel GenAI semantic convention) renders cache metrics as empty for Agents-SDK-instrumented runs, making it impossible to verify that OpenAI prompt caching is actually working. For high-cache-hit workloads (90%+), this is a >10x distortion of visible cost.

## 👟 Reproduction steps

1. Run an OpenAI Agents SDK `Runner.run` loop with multiple internal turns (a tool-calling agent that makes 2+ Responses API calls within a single run) with `auto_previous_response_id=True` so OpenAI actually returns `cached_tokens > 0`.
2. Enable only `Instruments.OPENAI_AGENTS` on Traceloop (`Traceloop.init(block_instruments={Instruments.OPENAI})`) to exclude raw-client spans.
3. Inspect the resulting `openai.response` spans' attributes.

## 👍 Expected behavior

Each `openai.response` span's metadata should include `gen_ai.usage.cache_read_input_tokens` whenever the underlying `response.usage.input_tokens_details.cached_tokens` is populated — matching what `opentelemetry-instrumentation-openai` already does for non-Agents OpenAI calls.

## 👎 Actual behavior

Metadata contains only:

```
gen_ai.usage.input_tokens
gen_ai.usage.output_tokens
```

`cache_read_input_tokens` is never set, even when `response.usage.input_tokens_details.cached_tokens > 0`.

## 🔧 Suggested fix

In [`packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py`](https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py), right after the existing `input_tokens` handler (~line 292):

```python
details = getattr(usage, "input_tokens_details", None)
if details is not None:
    cached = getattr(details, "cached_tokens", None)
    if cached is not None:
        otel_span.set_attribute(
            "gen_ai.usage.cache_read_input_tokens", cached
        )
```

The `gen_ai.usage.cache_read_input_tokens` attribute matches the OTel GenAI semantic convention used by the existing OpenAI client instrumentor. Analogous handling could be added for `output_tokens_details.reasoning_tokens` to capture reasoning-model token breakdowns.

## Environment

- `opentelemetry-instrumentation-openai-agents==0.52.3` (latest on PyPI `0.60.0` confirmed still affected; `_hooks.py` on `main` still reads only `input_tokens`/`output_tokens`)
- Python 3.12
- `openai-agents==0.9.1`
- `openai>=1.50`

hansmire:
## Summary

The OpenAI Agents SDK `Usage` object carries two additional token-count fields not currently captured by this instrumentor:

- `usage.input_tokens_details.cached_tokens` — prompt-cache hits returned by the Responses API.
- `usage.output_tokens_details.reasoning_tokens` — reasoning-model reasoning-token breakdown.

Users who install only the openai-agents instrumentor (e.g. via `Traceloop.init(block_instruments={Instruments.OPENAI})` to avoid duplicate spans with the raw-client instrumentor) currently have **no cache visibility**, even when caching is working at 99%+. That makes prompt-caching verification impossible and, downstream, causes backends like Braintrust to render empty `prompt_cached_tokens` columns.

Fixes #4060.

## Changes

`packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py` — in `_extract_response_attributes`, after the existing `total_tokens` handler, read the two `_tokens_details` sub-objects and record them under the conventional OTel GenAI attribute names:

- `gen_ai.usage.cache_read_input_tokens` via `SpanAttributes.LLM_USAGE_CACHE_READ_INPUT_TOKENS` (same constant already used by the anthropic and langchain instrumentors in this repo).
- `gen_ai.usage.reasoning_tokens` (string literal — no corresponding constant in `semconv_ai` yet; happy to add one in a follow-up if preferred).

Diff: 18 insertions, 0 deletions in `_hooks.py`; 61 insertions, 0 deletions in tests.

## Tests

Added two unit tests in `tests/test_openai_agents.py` that exercise `_extract_response_attributes` directly (no VCR needed):

- `test_extract_response_cached_and_reasoning_tokens` — asserts both new attributes are recorded when the Usage object carries the detail fields.
- `test_extract_response_cache_tokens_absent_when_missing` — regression guard: when the detail objects are absent, the new attributes must not be set.

Both pass locally:
```
uv run pytest tests/test_openai_agents.py::test_extract_response_cached_and_reasoning_tokens tests/test_openai_agents.py::test_extract_response_cache_tokens_absent_when_missing -v
======================== 2 passed in 0.25s ========================
```

`uv run ruff check` on both modified files is clean.

## End-to-end verification

I installed this branch into a downstream agent repo that uses `Traceloop.init(block_instruments={Instruments.OPENAI})` and re-ran a two-turn agent case against the OpenAI Responses API with `auto_previous_response_id=True`. The resulting `openai.response` span metadata now includes:

```
gen_ai.usage.input_tokens = 22026
gen_ai.usage.output_tokens = 347
gen_ai.usage.cache_read_input_tokens = 20608   ← new
gen_ai.usage.reasoning_tokens = 0              ← new
```

Before the patch, the same run produced a span with `input_tokens` and `output_tokens` only — no way to see that 93% of the prompt was served from cache.

## Summary by CodeRabbit

* **New Features**
  * Instrumentation now records cached input-token and reasoning-token metrics from responses when available, improving observability of token usage.
* **Tests**
  * Added unit tests to confirm the new token attributes are emitted when present and omitted when absent, preserving existing token reporting behavior.

### DIFF
--- a/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
+++ b/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
@@ -307,6 +307,24 @@ def _extract_response_attributes(otel_span, response, trace_content: bool):
                 SpanAttributes.LLM_USAGE_TOTAL_TOKENS, usage.total_tokens
             )
 
+        input_details = getattr(usage, "input_tokens_details", None)
+        if input_details is not None:
+            cached_tokens = getattr(input_details, "cached_tokens", None)
+            if cached_tokens is not None:
+                otel_span.set_attribute(
+                    SpanAttributes.LLM_USAGE_CACHE_READ_INPUT_TOKENS,
+                    cached_tokens,
+                )
+
+        output_details = getattr(usage, "output_tokens_details", None)
+        if output_details is not None:
+            reasoning_tokens = getattr(output_details, "reasoning_tokens", None)
+            if reasoning_tokens is not None:
+                otel_span.set_attribute(
+                    SpanAttributes.LLM_USAGE_REASONING_TOKENS,
+                    reasoning_tokens,
+                )
+
     return model_settings
 
 
--- a/packages/opentelemetry-instrumentation-openai-agents/tests/test_openai_agents.py
+++ b/packages/opentelemetry-instrumentation-openai-agents/tests/test_openai_agents.py
@@ -501,3 +501,64 @@ def test_tool_call_and_result_attributes(exporter):
 
     assert tool_call_found, "No assistant message with tool_calls found in second response span"
     assert tool_result_found, "No tool message found in second response span"
+
+
+def test_extract_response_cached_and_reasoning_tokens():
+    """Verify cache-read and reasoning tokens are recorded when the Agents SDK
+    Usage object carries them via input_tokens_details / output_tokens_details.
+    """
+    from types import SimpleNamespace
+    from opentelemetry.instrumentation.openai_agents._hooks import (
+        _extract_response_attributes,
+    )
+
+    recorded: dict[str, object] = {}
+
+    class _FakeSpan:
+        def set_attribute(self, key, value):
+            recorded[key] = value
+
+    response = SimpleNamespace(
+        usage=SimpleNamespace(
+            input_tokens=20_750,
+            output_tokens=97,
+            total_tokens=20_847,
+            input_tokens_details=SimpleNamespace(cached_tokens=20_608),
+            output_tokens_details=SimpleNamespace(reasoning_tokens=64),
+        ),
+    )
+
+    _extract_response_attributes(_FakeSpan(), response, trace_content=False)
+
+    assert recorded[SpanAttributes.LLM_USAGE_CACHE_READ_INPUT_TOKENS] == 20_608
+    assert recorded[SpanAttributes.LLM_USAGE_REASONING_TOKENS] == 64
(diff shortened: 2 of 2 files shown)


dvirski:
Thanks for the PR! Closing in favor of #4130, which lands the same fix for cached input tokens plus reasoning_tokens emission per the OTel GenAI spec. Appreciate the contribution!

PR #4078, opened 2026-05-06
fix(openai): guard against AsyncAPIResponse without .id in async responses wrapper

### LINKED ISSUE #4058: 🐛 Bug Report: OpenAI Agents instrumentation

OpenAI Instrumentation

### 📜 Description

When running the [OpenAI Customer Service Agents demo](https://github.com/openai/openai-cs-agents-demo), the instrumentation crashes at this line: 
https://github.com/traceloop/openllmetry/blob/3efa0eb4e21df170f51aa0f85383e75be2fca171/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py#L739

```
async_responses_get_or_create_wrapper
    existing_data = responses.get(parsed_response.id) 
                                  ^^^^^^^^^^^^^^^^^^
AttributeError: 'AsyncAPIResponse' object has no attribute 'id'
```

### 👟 Reproduction steps

1. Instrument https://github.com/openai/openai-cs-agents-demo with OpenLLMetry
2. Run the demo

### 👍 Expected behavior

No crashes

it crashes

```
async_responses_get_or_create_wrapper
    existing_data = responses.get(parsed_response.id) 
                                  ^^^^^^^^^^^^^^^^^^
AttributeError: 'AsyncAPIResponse' object has no attribute 'id'
```

### 🤖 Python Version

3.13

_No response_

dvirski:
Fixes #4058 — OpenAI Agents crash when AsyncAPIResponse without .id is returned by parse_response 

## Summary by CodeRabbit

* **Bug Fixes**
  * Response parsing now correctly unwraps legacy, modern, and async response wrappers, returning the underlying response objects to avoid wrapped results and improve compatibility.

* **Tests**
  * Added unit tests verifying synchronous and asynchronous unwrapping for legacy and modern wrappers, and that plain response-like objects remain unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4078)

### DIFF
--- a/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py
+++ b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py
@@ -736,6 +736,9 @@ async def async_responses_get_or_create_wrapper(
         raise
     parsed_response = parse_response(response)
 
+    if not hasattr(parsed_response, "id"):
+        return response
+
     existing_data = responses.get(parsed_response.id)
     if existing_data is None:
         existing_data = {}
--- a/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py
+++ b/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py
@@ -234,10 +234,16 @@ def _derive_finish_reason(traced_data: TracedData) -> str:
     return ""
 
 
-def parse_response(response: Union[LegacyAPIResponse, APIResponse, AsyncAPIResponse, Response]) -> Response:
-    if isinstance(response, (LegacyAPIResponse, APIResponse, AsyncAPIResponse)):
+def parse_response(response: Union[LegacyAPIResponse, APIResponse, Response]) -> Response:
+    if isinstance(response, (LegacyAPIResponse, APIResponse)):
         return response.parse()
     return response
+
+
+async def async_parse_response(response: Union[LegacyAPIResponse, APIResponse, AsyncAPIResponse, Response]) -> Response:
+    if isinstance(response, AsyncAPIResponse):
+        return await response.parse()
+    return parse_response(response)
 
 
 def get_tools_from_kwargs(kwargs: dict) -> list[ToolParam]:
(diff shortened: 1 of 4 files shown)


doronkopit5:
2 suggestions toward a simpler/more robust shape before this lands:

### 1. The same `parse_response(response).id` pattern is unguarded in 4 other places

Only the one site you patched has the guard. The others will crash the same way under `with_raw_response` (sync) or in the cancel paths:

| Line | Function | Guarded? |
|---|---|---|
| 570 | `responses_get_or_create_wrapper` (sync) | ❌ |
| 737 | `async_responses_get_or_create_wrapper` | ✅ (this PR) |
| 827 | `responses_cancel_wrapper` | ❌ |
| 859 | `async_responses_cancel_wrapper` | ❌ |
| ~1050 | `ResponseStream` finalize path | ❌ |

The reporter only hit the async path because the cs-agents-demo is async, but the sync caller using `client.responses.with_raw_response.create(...)` trips the identical `AttributeError` on line 572.

### 2. The cleanest fix is in `parse_response` itself

`parse_response` (line 236) currently only unwraps `LegacyAPIResponse`:

```python
def parse_response(response):
    if isinstance(response, LegacyAPIResponse):
        return response.parse()
    return response
```

But `APIResponse` and `AsyncAPIResponse` (the non-legacy variants returned by `with_raw_response()`) also expose a `.parse()` method that yields the typed `Response` — which has the `id` everything downstream needs. So the one-line fix is to unwrap them too:

```python
def parse_response(response):
    if hasattr(response, "parse") and callable(getattr(response, "parse", None)):
        return response.parse()
    return response
```

or, more explicit:

```python
from openai._response import APIResponse, AsyncAPIResponse  # alongside LegacyAPIResponse

def parse_response(response):
    if isinstance(response, (LegacyAPIResponse, APIResponse, AsyncAPIResponse)):
        return response.parse()
    return response
```

Benefits over the per-site `hasattr` guard:
- Fixes all 5 call sites at once.
- **Preserves telemetry** — the current early-return drops the span entirely, so users who instrument with `with_raw_response` silently get no observability for those calls. Unwrapping at the source means they get a proper span with the real `id`.
- Matches the function's declared return type (`-> Response`).

If unwrapping in `parse_response` feels too broad for this PR, then at minimum please mirror the `hasattr` guard to the other 4 sites — a guard in only one of five identical places is harder to reason about than no guard at all.

dvirski:
@doronkopit5 You are absolutely right. made that change plus fixed the below suggested code rabbit change

doronkopit5 (approved):

PR #4079, opened 2026-05-06
fix(anthropic): prevent KeyError/IndexError in streaming tool use delta handler

### LINKED ISSUE #4050: 🐛 Bug: KeyError: 'input' in _process_response_item during Anthropic streaming

dvirski:
Fixes #4050 — KeyError/IndexError crash in Anthropic streaming tool use delta handler 

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous streaming instrumentation for non-blocking monitoring.
* **Bug Fixes**
  * Improved robustness of streaming response handling to tolerate out-of-order content deltas and missing fields without failures.
* **Tests**
  * Added regression tests validating resilience of streaming handling in these edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4079)

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
@@ -48,12 +48,14 @@ def _process_response_item(item, complete_response):
                 complete_response["events"][index]["input"] = """"""
     elif item.type == "content_block_delta":
         index = item.index
-        if item.delta.type == "thinking_delta":
-            complete_response["events"][index]["text"] += item.delta.thinking
-        elif item.delta.type == "text_delta":
-            complete_response["events"][index]["text"] += item.delta.text
-        elif item.delta.type == "input_json_delta":
-            complete_response["events"][index]["input"] += item.delta.partial_json
+        if index < len(complete_response.get("events", [])):
+            event = complete_response["events"][index]
+            if item.delta.type == "thinking_delta":
+                event["text"] = event.get("text", "") + item.delta.thinking
+            elif item.delta.type == "text_delta":
+                event["text"] = event.get("text", "") + item.delta.text
+            elif item.delta.type == "input_json_delta":
+                event["input"] = event.get("input", "") + item.delta.partial_json
     elif item.type == "message_delta":
         for event in complete_response.get("events", []):
             event["finish_reason"] = item.delta.stop_reason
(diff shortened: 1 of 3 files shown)


doronkopit5:
One thing I'd like to investigate further before merging, and I'd appreciate your input since I'm not deeply familiar with this part of the Anthropic streaming protocol:

In the second test (`test_process_response_item_no_crash_on_missing_input_key`), an `input_json_delta` arriving for a `type: "text"` block ends up writing an `"input"` key onto the text event. That's better than crashing, but semantically it feels off — `input_json_delta` is conceptually tied to `tool_use` blocks (that's why `content_block_start` only initializes `"input": ""` for `tool_use`).

Could you do a bit more research on the Anthropic streaming spec / SDK behavior here and confirm:

1. Can `input_json_delta` legitimately arrive for non-`tool_use` blocks? (My guess is no, but I'd like confirmation.)
2. If it can't, would it be more robust to guard the branch with `event.get("type") == "tool_use"` and silently drop mismatched deltas instead of writing an `"input"` field onto a text block?

If that holds up, the test should assert that no `"input"` key was added to a text event, and the code change becomes:

```python
elif item.delta.type == "input_json_delta":
    if event.get("type") == "tool_use":
        event["input"] = event.get("input", "") + item.delta.partial_json
```

Let me know what you find in the spec — happy to iterate.

dvirski:
@doronkopit5 you were right, good catch. changed the test and adde the guard you suggested.

doronkopit5 (approved):

PR #4084, opened 2026-05-07
fix(instrumentation): pinecone span attributes for upsert and delete (#2688)

### LINKED ISSUE #2688: 🐛 Bug Report: No Span Attributes defined for upsert and delete

Pinecone Instrumentation

### 📜 Description

Currently we don't get span attributes for upsert and delete methods . Only query() has span attributed defined . 

### 👟 Reproduction steps

implement pinecone.Index class query , upsert and delete methods . 
We will get span attributes only for query().

### 👍 Expected behavior

We should get span attributes for upsert and delete methods as well like chromadb and Milvus Db. 

Image

### 🤖 Python Version

3.11

_No response_

WatchTree-19:

---

Fixes #2688

Pinecone instrumentation only set span attributes for `query()`. `upsert()` and `delete()` produced spans with the bare vendor tag and nothing operation-specific.

Following the pattern from `chromadb` (`_set_upsert_attributes`, `_set_delete_attributes` in chromadb's wrapper), this PR adds:

- 7 new `SpanAttributes` constants in `semconv_ai`:
  - `PINECONE_UPSERT_VECTORS_COUNT`, `PINECONE_UPSERT_NAMESPACE`, `PINECONE_UPSERT_BATCH_SIZE`
  - `PINECONE_DELETE_IDS_COUNT`, `PINECONE_DELETE_NAMESPACE`, `PINECONE_DELETE_DELETE_ALL`, `PINECONE_DELETE_FILTER`
- `set_upsert_input_attributes` and `set_delete_input_attributes` in `query_handlers.py`. Both `@dont_throw`. Both check `kwargs` then fall back to `args[0]` for `vectors` / `ids` since Pinecone callers usually pass them positionally. `len()` is wrapped in `try / except TypeError` so a generator-typed `vectors` doesn't get materialised just to count.
- Dispatch in `_wrap` switched to a small `if/elif` on `to_wrap.get("method")`.

Did not touch response-side attributes. The existing `_set_response_attributes` already records `read_units` / `write_units` for any response with a `usage` block, which covers the metric side for upsert. Pure upsert/delete-specific response attributes (vector IDs returned, etc.) can be a follow-up.

No tests added in this PR; cassette-based tests for the new setters would be a small follow-up. Happy to add if you'd like.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Pinecone tracing to record additional span details for **upsert** and **delete**, including vectors/IDs counts (when available), namespaces, upsert batch size, the delete-all flag, and filter information.
* **Semantic Conventions**
  * Added new Pinecone-specific span attribute keys to standardize these telemetry fields.
* **Bug Fixes**
  * Improved when certain response/usage attributes are captured by treating `None` responses differently from other responses.
* **Tests**
  * Added unit tests for attribute extraction from keyword and positional inputs, including generator handling and JSON-serialization for dict filters.


WatchTree-19:
gentle bump on this when you get a chance. PR adds pinecone span attributes for upsert/delete following the chromadb pattern (#2688). CLA cleared and coderabbit-bot reviewed positively. happy to address any feedback or rebase if it's gone stale.

WatchTree-19:
rebased onto main and resolved the conflict in `semconv_ai/__init__.py` (kept both the new `db.pinecone.query.*` constants and these `pinecone.upsert.*` / `pinecone.delete.*` ones - no name overlap). also added unit tests for the upsert/delete attribute helpers. conflict-free and CLA green now.

one thing worth a call: i kept the `pinecone.*` prefix to match the existing `pinecone.query.*` / `pinecone.usage.*` attributes, but i see newer additions are moving to `db.pinecone.*`. happy to rename these to `db.pinecone.upsert.*` / `db.pinecone.delete.*` if that's the direction you'd prefer.

WatchTree-19:
thanks - addressed the coderabbit note: a successful pinecone delete returns an empty dict, which the truthiness guard skipped, so switched to `if response is not None:` (901ba84) and the delete span attributes are now set.

still one open question for a maintainer when someone has a moment: i kept the `pinecone.upsert.*` / `pinecone.delete.*` prefix to match the existing `pinecone.query.*` / `pinecone.usage.*` attributes, but main also added newer `db.pinecone.query.*` constants. happy to rename to `db.pinecone.*` if that's the preferred direction - just didn't want to make that call unilaterally in a rebase.

PR #4098, opened 2026-05-10
fix(sdk): replace deprecated Pydantic .json() and double-encoding .model_dump_json() in JSONEncoders

### LINKED ISSUE #3516: 🐛 Bug Report: JSONEncoder calling deprecated method `json` for Pydantic models

dvirski:
Fixes #3516. Applies across sdk, llamaindex, openai-agents, langchain, and anthropic packages.

Problem:                                                                                                                                                                                                                                                                                                 
Several JSONEncoder classes across the repo were serializing Pydantic models using either .json() (deprecated in Pydantic v2) or .model_dump_json() (returns a JSON string). Using .model_dump_json() inside JSONEncoder.default() causes double-encoding — the string gets JSON-encoded again by the
outer encoder, turning {"name": "Alice"} into "{\"name\": \"Alice\"}".                                                                                                                                                                                                                                   
   
Fix:                                                                                                                                                                                                                                                                                                     
Replace both with .model_dump() (Pydantic v2) which returns a plain dict — exactly what JSONEncoder.default() expects. A .dict() fallback covers Pydantic v1, and .json() is kept at the end as a last resort for non-Pydantic objects that happen to have that method. Applied consistently across 5
packages: sdk, llamaindex, openai-agents, langchain, anthropic.    

## Summary by CodeRabbit

* **Refactor**
  * Improved JSON serialization across instrumentations: prefers modern model representations, avoids double-encoding, and handles non-standard or asynchronous object outputs more safely to produce stable, predictable JSON.

* **Tests**
  * Added comprehensive tests covering serialization ordering, dataclass and model support, async-safe handling, legacy fallbacks, and regression safeguards.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4098?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import json
 import logging
 import os
@@ -302,7 +303,12 @@ class JSONEncoder(json.JSONEncoder):
             return o.model_dump()
 
         if hasattr(o, "dict"):
-            return o.dict()
+            dict_method = o.dict
+            if callable(dict_method) and not inspect.iscoroutinefunction(dict_method):
+                result = dict_method()
+                if not inspect.iscoroutine(result):
+                    return result
+                result.close()
 
         try:
             return str(o)
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/utils.py
@@ -298,8 +298,11 @@ class JSONEncoder(json.JSONEncoder):
         if hasattr(o, "to_json"):
             return o.to_json()
 
-        if hasattr(o, "model_dump_json"):
-            return o.model_dump_json()
+        if hasattr(o, "model_dump"):
+            return o.model_dump()
+
+        if hasattr(o, "dict"):
+            return o.dict()
 
         try:
             return str(o)
--- a/packages/opentelemetry-instrumentation-anthropic/tests/test_json_encoder.py
+++ b/packages/opentelemetry-instrumentation-anthropic/tests/test_json_encoder.py
@@ -1,0 +1,50 @@
+import json
+
+import pytest
+
+from opentelemetry.instrumentation.anthropic.utils import JSONEncoder
+
+
+class PydanticV2Model:
+    def model_dump(self):
+        return {"name": "Alice", "age": 30}
+
+
+class PydanticV1Model:
+    def dict(self):
+        return {"name": "Bob", "age": 25}
+
+
+def test_pydantic_v2_uses_model_dump():
+    assert json.loads(json.dumps(PydanticV2Model(), cls=JSONEncoder)) == {"name": "Alice", "age": 30}
+
+
+def test_pydantic_v2_no_double_encoding():
+    # Regression: model_dump_json() would emit a JSON-encoded string literal.
+    parsed = json.loads(json.dumps(PydanticV2Model(), cls=JSONEncoder))
+    assert isinstance(parsed, dict), "result must decode to a dict, not a string"
+
+
+def test_pydantic_v1_uses_dict():
+    assert json.loads(json.dumps(PydanticV1Model(), cls=JSONEncoder)) == {"name": "Bob", "age": 25}
+
+
+@pytest.mark.parametrize("model_factory", [
+    pytest.param(lambda: type("HasDictDataAttribute", (), {"dict": {"not": "callable"}})(),
+                 id="non-callable-dict"),
+])
+def test_non_pydantic_dict_attribute_does_not_crash(model_factory):
+    # Non-Pydantic objects with a `.dict` attribute must not raise TypeError.
+    # The anthropic encoder falls back to str(o) — just assert it doesn't crash.
+    result = json.dumps(model_factory(), cls=JSONEncoder)
+    assert isinstance(json.loads(result), str)
+
+
+def test_coroutine_dict_method_does_not_crash():
+    class AsyncDictModel:
+        async def dict(self):
+            return {"async": "result"}
+
+    # Must not emit un-encodable garbage; falls back to str(o).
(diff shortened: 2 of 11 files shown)


doronkopit5:
Thanks for the fix — `model_dump_json()` → `model_dump()` correctly addresses both the deprecation and the double-encoding bug. A few suggestions for making it more robust:

### 1. Inconsistent robustness across the 5 encoders

The SDK version is defensive (callable check + coroutine handling on `dict()` and `json()`), but the other four packages call `o.dict()` blindly:

```python
# anthropic / llamaindex / openai-agents
if hasattr(o, "dict"):
    return o.dict()
```

Many non-Pydantic objects expose a `.dict` attribute that isn't a v1 model (`enum.DynamicClassAttribute`, descriptor-style attribute holders, etc.). A non-callable `.dict` raises `TypeError`; a coroutine-returning `.dict` produces un-encodable garbage. `langchain` got this right by gating on `isinstance(o, BaseModel)` — would be great to apply the same gate in the other three, or (preferred) factor a single shared encoder out of the SDK and import it from the four instrumentations. Five near-identical encoders is exactly what allowed this bug to live in multiple places to begin with.

### 2. Test file has dead code and a misleading test

In `packages/traceloop-sdk/tests/test_json_encoder.py`:

- `LegacyJsonModel` is defined but never referenced — the `.json()` fallback path (the original bug from #3516) has no coverage. Worth adding an explicit test for it.
- `test_dict_with_callbacks_stripped` defines a `DictWrapper` class then doesn't use it; the test passes a plain `dict` and directly invokes `encoder.default(data)`, which bypasses the real `json.dumps` flow. The comment in the test even acknowledges that plain dicts never hit `default()` in practice. Either drop `DictWrapper` and rename the test to reflect what it actually exercises, or exercise the `dict`-subclass path properly through `json.dumps`.

### 3. Tests only cover the SDK encoder

The PR changes four other encoders (`anthropic`, `langchain`, `llamaindex`, `openai-agents`) without adding any test for them. The excellent `test_pydantic_v2_no_double_encoding` regression test should be replicated in each — otherwise future drift will reintroduce the same bug in one of them, which is exactly what this PR is trying to prevent.

dvirski:
@doronkopit5 

Summary of what has been added:

Encoder fixes (defensive guards on .dict() / .json() calls):
  - packages/opentelemetry-instrumentation-anthropic/.../utils.py — guarded .dict()
  - packages/opentelemetry-instrumentation-llamaindex/.../utils.py — guarded .dict()
  - packages/opentelemetry-instrumentation-openai-agents/.../utils.py — guarded .dict() + .json()
  - packages/traceloop-sdk/.../json_encoder.py — fixed .json() double-encoding (parses JSON string before returning)
  
  SDK test cleanup (packages/traceloop-sdk/tests/test_json_encoder.py):
  - Removed unused pytest import and dead DictWrapper class
  - Replaced misleading test_dict_with_callbacks_stripped with honest test_default_strips_callbacks_from_dict
  - Added test_legacy_json_method_fallback, test_non_callable_dict_attribute_does_not_crash, test_coroutine_dict_method_does_not_crash
  
  New regression test files (one per instrumentation package):
  - packages/opentelemetry-instrumentation-anthropic/tests/test_json_encoder.py
  - packages/opentelemetry-instrumentation-langchain/tests/test_json_encoder.py
  - packages/opentelemetry-instrumentation-llamaindex/tests/test_json_encoder.py
  - packages/opentelemetry-instrumentation-openai-agents/tests/test_json_encoder.py
  
  Each includes test_pydantic_v2_no_double_encoding regression plus defensive-guard tests tailored to that encoder's fallback behavior.

Verification: 32 tests passing across all 5 packages, ruff clean.

We didn't create a traceloop-utils package because Python wheels are independent units, so a shared utility would require a full new PyPI package with build config, release workflow changes, and cross-package wiring — too much overhead for ~30 lines of code when per-package duplication of the defensive guards is honest and low-risk.

doronkopit5:
you didnt commit the new test files
take a look at the 2 coderabbit comments just added

doronkopit5 (approved):

dvirski:
@doronkopit5 

added missing test files 

+ fixed code rabbit major comment (the major one):

packages/opentelemetry-instrumentation-llamaindex/.../utils.py — added a guarded .json() fallback branch after .dict(), mirroring the SDK + openai-agents implementation. Includes:
  - callable + coroutine-function guards
  - coroutine-result close
  - JSON-string parse via json.loads to avoid double-encoding

  packages/opentelemetry-instrumentation-llamaindex/tests/test_json_encoder.py — added two tests:                                                                                                                          
  - test_legacy_json_method_fallback — locks in correct (non-double-encoded) serialization of .json()-only objects
  - test_non_callable_json_attribute_does_not_crash — locks in the callable guard

PR #4099, opened 2026-05-10
fix(anthropic): fix double-counted input tokens in non-streaming create() paths

dvirski:
Fixes #3949. Complements PR #3976 which fixed the same arithmetic in                                                                                                                                                                                                                                     
the streaming path.                 

The problem was that Cache sub-tokens (cache_read, cache_creation) were being added to                                                                                                                                                                                                                                        
input_tokens, which already includes them per the Anthropic API spec.

## Summary by CodeRabbit

## Bug Fixes
* Fixed token usage accounting in Anthropic instrumentation where input tokens were incorrectly including cached prompt tokens. Input tokens are now calculated using only prompt tokens, with cached token metrics available as separate attributes.

## Tests
* Updated test cases to reflect the corrected token counting behavior for Anthropic prompt caching scenarios.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4099)

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -229,7 +229,7 @@ async def _aset_token_usage(
         cache_read_tokens = 0
         cache_creation_tokens = 0
 
-    input_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens
+    input_tokens = prompt_tokens
 
     if token_histogram and isinstance(input_tokens, int) and input_tokens >= 0:
         token_histogram.record(
@@ -345,7 +345,7 @@ def _set_token_usage(
         cache_read_tokens = 0
         cache_creation_tokens = 0
 
-    input_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens
+    input_tokens = prompt_tokens
 
     if token_histogram and isinstance(input_tokens, int) and input_tokens >= 0:
         token_histogram.record(
--- a/packages/opentelemetry-instrumentation-anthropic/tests/test_prompt_caching.py
+++ b/packages/opentelemetry-instrumentation-anthropic/tests/test_prompt_caching.py
@@ -127,7 +127,7 @@ def test_anthropic_prompt_caching_legacy(
     assert json.loads(cache_creation_span.attributes[GenAIAttributes.GEN_AI_OUTPUT_MESSAGES])[-1]["role"] == "assistant"
     assert json.loads(cache_read_span.attributes[GenAIAttributes.GEN_AI_OUTPUT_MESSAGES])[-1]["role"] == "assistant"
 
-    _verify_caching_attributes(cache_creation_span, cache_read_span, 1167, 187, 202, 1163)
+    _verify_caching_attributes(cache_creation_span, cache_read_span, 4, 187, 202, 1163)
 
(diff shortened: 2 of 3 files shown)


doronkopit5 (changes requested):
I want to push back on this before we land it, because the premise rests on a name collision that I think led to a misreading of what the attribute means.

**`gen_ai.usage.input_tokens` is the OTel-spec rename of `gen_ai.usage.prompt_tokens`, not Anthropic's `usage.input_tokens` field.** This is documented right in our own `packages/opentelemetry-semantic-conventions-ai/MIGRATION.md:36`:

```
SpanAttributes.LLM_USAGE_PROMPT_TOKENS → GenAIAttributes.GEN_AI_USAGE_INPUT_TOKENS
```

The semconv migration was a rename, not a semantic change. The attribute still means "total prompt tokens" — provider-neutral. Anthropic's API just happens to use the same string `input_tokens` for a *narrower* concept (the non-cached delta only). Those are two different things that share a name. You can confirm Anthropic's field is the non-cached portion by decompressing our own cassettes:

```
# test_anthropic_prompt_caching_legacy.yaml
[0] "usage":{"input_tokens":4,"cache_creation_input_tokens":1163,"cache_read_input_tokens":0,"output_tokens":187}
[1] "usage":{"input_tokens":4,"cache_creation_input_tokens":0,"cache_read_input_tokens":1163,"output_tokens":202}
```

Three disjoint fields. The old code's `prompt_tokens + cache_read_tokens + cache_creation_tokens` was deliberately reconstructing the spec's "total prompt" from Anthropic's three pieces. It wasn't double-counting — it was doing the right thing. The OpenAI side gets this for free because OpenAI's `usage.prompt_tokens` is already pre-summed (cached is a subset); we have to sum ourselves for Anthropic.

**After this PR, `gen_ai.usage.input_tokens` means different things across providers** — total-prompt for OpenAI, non-cached-only for Anthropic. That violates the cross-provider neutrality the semconv exists to provide.

**A few more issues with the PR as it stands:**

- **Doesn't actually fix #3949.** That issue is about *streaming* with base64 images, and its own reproduction shows non-streaming `create()` was already correct (343 → 343). #3976 fixed the real root cause. The non-streaming path this PR changes wasn't the broken one.

- **Streaming vs non-streaming now diverge.** `streaming.py:89` still does `input_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens`. After this PR, `messages.create()` and `messages.stream()` emit different `gen_ai.usage.input_tokens` for the same call when caching is in use.

- **`total_tokens` is broken either way and should be fixed.** Today it's `input_tokens + completion_tokens` (lines 270/386), which is only correct if `input_tokens` already encompasses the cache portions. The robust form is to sum all token types explicitly:

  ```python
  total_tokens = prompt_tokens + cache_read_tokens + cache_creation_tokens + completion_tokens
  ```

  That way it's independent of whatever convention we pick for `input_tokens`, and any future Anthropic usage fields (thinking tokens, etc.) get added in one obvious spot.

**Recommended path: revert this PR**, and as a follow-up tighten `total_tokens` to explicitly sum all the token-type fields rather than relying on `input_tokens` carrying the full count. That gets us: consistent with OpenAI semantics, consistent between our streaming and non-streaming paths, and robust against future per-provider usage additions. CodeRabbit's nit about adding a `gen_ai.usage.total_tokens` assertion to `_verify_caching_attributes` is also worth taking — it would have caught this drift.

PR #4005, opened 2026-04-16
fix(ollama,together,replicate,sagemaker,bedrock): record exceptions on error spans

Koushik-Salammagari:
## Summary

Fixes #412 (partial — LLM instrumentation packages)

When API calls to LLM providers raise exceptions, spans were left in `UNSET` status with no error information attached. This PR adds proper error recording to five packages:

- **ollama**: sync `_wrap` and async `_awrap` — exceptions now call `span.record_exception(e)`, `span.set_status(ERROR)`, `span.end()` before re-raising
- **together**: `_wrap` — same pattern
- **replicate**: `_wrap` — same pattern
- **sagemaker**: both `_instrumented_endpoint_invoke` and `_instrumented_endpoint_invoke_with_response_stream`
- **bedrock**: all four `_instrumented_model_invoke`, `_instrumented_model_invoke_with_response_stream`, `_instrumented_converse`, `_instrumented_converse_stream` — adds `Status, StatusCode` import

This follows the same approach as traceloop/openllmetry#3970 (anthropic/groq/mistralai).

## Test plan

- [ ] Trigger an API error (e.g. invalid API key or model name) in each provider
- [ ] Verify the resulting span has `status=ERROR` and an attached exception event
- [ ] Verify successful calls still produce `status=OK` spans
- [ ] Run existing test suites: `npx nx run-many -t test --projects=opentelemetry-instrumentation-ollama,opentelemetry-instrumentation-together,opentelemetry-instrumentation-replicate,opentelemetry-instrumentation-sagemaker,opentelemetry-instrumentation-bedrock`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling across OpenTelemetry instrumentations (Bedrock, Ollama, Replicate, SageMaker, Together): exceptions are now explicitly recorded on spans, spans are marked with error status and ended promptly on exceptions (including streaming paths), and automatic exception-to-span behavior was adjusted to use explicit handling for more reliable and consistent trace visibility.

PR #3970, opened 2026-04-10
fix: record errors on spans for anthropic, groq, and mistralai

Dheeraj-Bhaskaruni:
## Summary

Fixes #412

When API calls to foundation model providers raise exceptions, the OpenTelemetry spans were not being properly marked as failed. This PR adds error recording to three instrumentation packages:

- **anthropic**: Sync and async wrappers caught exceptions for metrics but never called `span.record_exception()`, `span.set_status(ERROR)`, or `span.end()`
- **groq**: Same issue — exceptions were caught for duration metrics but span was left in UNSET state
- **mistralai**: No `try/except` at all around `wrapped()` calls, so exceptions propagated with spans never ended or marked

### Changes

All three packages now follow the same pattern used by `opentelemetry-instrumentation-openai`:

```python
except Exception as e:
    span.record_exception(e)
    span.set_status(Status(StatusCode.ERROR, str(e)))
    span.end()
    raise
```

### Files changed

- `packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py`
- `packages/opentelemetry-instrumentation-groq/opentelemetry/instrumentation/groq/__init__.py`
- `packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py`

## Test plan

- [x] `ruff check` passes on all changed files
- [x] Error spans appear correctly in trace exporters when API calls fail
- [x] Exceptions still propagate to callers after being recorded
- [x] Streaming and non-streaming paths both handle errors correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved observability for Anthropic, Groq, and Mistral integrations: exceptions are now recorded on spans, spans are marked as error, and spans are ended immediately to ensure accurate failure reporting and metrics.

* **Tests**
  * Updated tests to filter out finished spans with error status when asserting non-error trace outcomes, reducing noise from failed spans.

PR #4101, opened 2026-05-10
fix(langchain,anthropic,groq,mistralai,bedrock,ollama,sagemaker,together): record exceptions and set ERROR status on failed spans

### LINKED ISSUE #412: 🐛 Bug Report: Errors are not logged

All Packages

### 📜 Description

If an HTTP error is returned from a foundation model API, we don't properly log it as a failed span.

### 👟 Reproduction steps

N/A

### 👍 Expected behavior

N/A

N/A

### 🤖 Python Version

_No response_

_No response_

None

dvirski:
Fixes #412.
Applies consistent error handling pattern for:
langchain, anthropic, groq, mistralai, bedrock, ollama, sagemaker, together.                                                                                                                                                    
record_exception + set_status(ERROR) + ERROR_TYPE attribute on all
instrumented call sites. Also fixes pre-existing bug in LangChain where                                                                                                                                                                                                                                  
set_status() was called with wrong argument order, silently dropping the                                                                                                                                                                                                                                 
error description from spans. Adds regression tests for each package.                                                                                                                                                                                                                                    
Supersedes PRs #3970 and #4005. 




## Summary by CodeRabbit

* **Bug Fixes**
  * Instrumentations now consistently record exception type and message on spans, set span status to error, preserve proper exception propagation, and ensure streaming spans are finalized on failures.

* **Tests**
  * Added coverage asserting spans include error details and are finished on failure; removed silent exception-workarounds and updated assertions to match explicit error handling.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4101?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
@@ -259,7 +259,7 @@ class AnthropicStream(ObjectProxy):
             attributes = error_metrics_attributes(e)
             if self._exception_counter:
                 self._exception_counter.add(1, attributes=attributes)
-            raise e
+            raise
         _process_response_item(item, self._complete_response)
         return item
 
--- a/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
+++ b/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
@@ -62,6 +62,7 @@ from opentelemetry.semconv._incubating.attributes.gen_ai_attributes import (
     GenAiOperationNameValues,
     GenAiSystemValues,
 )
+from opentelemetry.semconv.attributes.error_attributes import ERROR_TYPE
 from opentelemetry.semconv_ai import (
     SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY,
     Meters,
@@ -211,7 +212,7 @@ def _wrap(
             if metric_params.exception_counter:
                 metric_params.exception_counter.add(1, attributes=attributes)
 
-            raise e
+            raise
 
     return wrapped(*args, **kwargs)
 
@@ -236,6 +237,7 @@ def _instrumented_model_invoke(fn, tracer, metric_params, event_logger):
             try:
                 response = fn(*args, **kwargs)
             except Exception as e:
+                span.set_attribute(ERROR_TYPE, e.__class__.__name__)
                 span.record_exception(e)
                 span.set_status(Status(StatusCode.ERROR, str(e)))
                 raise
@@ -269,6 +271,7 @@ def _instrumented_model_invoke_with_response_stream(
         try:
             response = fn(*args, **kwargs)
         except Exception as e:
+            span.set_attribute(ERROR_TYPE, e.__class__.__name__)
             span.record_exception(e)
             span.set_status(Status(StatusCode.ERROR, str(e)))
             span.end()
@@ -304,6 +307,7 @@ def _instrumented_converse(fn, tracer, metric_params, event_logger):
             try:
                 response = fn(*args, **kwargs)
             except Exception as e:
+                span.set_attribute(ERROR_TYPE, e.__class__.__name__)
                 span.record_exception(e)
                 span.set_status(Status(StatusCode.ERROR, str(e)))
                 raise
@@ -334,6 +338,7 @@ def _instrumented_converse_stream(fn, tracer, metric_params, event_logger):
         try:
             response = fn(*args, **kwargs)
         except Exception as e:
+            span.set_attribute(ERROR_TYPE, e.__class__.__name__)
             span.record_exception(e)
             span.set_status(Status(StatusCode.ERROR, str(e)))
             span.end()
--- a/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
+++ b/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
@@ -67,6 +67,7 @@ from opentelemetry.semconv_ai import (
     Meters,
 )
 from opentelemetry.trace import Span, SpanKind, get_tracer
+from opentelemetry.trace.status import Status, StatusCode
 from wrapt import wrap_function_wrapper
 
 
@@ -229,9 +230,15 @@ def _instrumented_model_invoke(fn, tracer, metric_params, event_logger):
             GenAIAttributes.GEN_AI_REQUEST_MODEL: _model,
         }
         with tracer.start_as_current_span(
-            _span_name(operation_name, _model), kind=SpanKind.CLIENT, attributes=span_attributes
+            _span_name(operation_name, _model), kind=SpanKind.CLIENT, attributes=span_attributes,
+            record_exception=False, set_status_on_exception=False,
         ) as span:
-            response = fn(*args, **kwargs)
+            try:
+                response = fn(*args, **kwargs)
+            except Exception as e:
+                span.record_exception(e)
+                span.set_status(Status(StatusCode.ERROR, str(e)))
+                raise
             _handle_call(span, kwargs, response, metric_params, event_logger)
             return response
 
@@ -259,7 +266,13 @@ def _instrumented_model_invoke_with_response_stream(
             attributes=span_attributes,
         )
 
-        response = fn(*args, **kwargs)
+        try:
+            response = fn(*args, **kwargs)
+        except Exception as e:
+            span.record_exception(e)
+            span.set_status(Status(StatusCode.ERROR, str(e)))
+            span.end()
+            raise
         _handle_stream_call(span, kwargs, response, metric_params, event_logger)
 
         return response
@@ -286,8 +299,14 @@ def _instrumented_converse(fn, tracer, metric_params, event_logger):
             _span_name(GenAiOperationNameValues.CHAT.value, _model),
             kind=SpanKind.CLIENT,
             attributes=span_attributes,
+            record_exception=False, set_status_on_exception=False,
         ) as span:
-            response = fn(*args, **kwargs)
+            try:
+                response = fn(*args, **kwargs)
+            except Exception as e:
+                span.record_exception(e)
+                span.set_status(Status(StatusCode.ERROR, str(e)))
+                raise
             _handle_converse(span, kwargs, response, metric_params, event_logger)
 
             return response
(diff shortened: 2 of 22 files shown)


doronkopit5:
Thanks for consolidating #3970 and #4005 — the langchain `set_status()` arg-order fix is a great catch. A few things I'd like to see tightened before merge so we end up with one consistent pattern across the 8 packages rather than 8 slightly-different ones:

### 1. `ERROR_TYPE` is inconsistent across packages
The PR description says "consistent error handling pattern", but only 3 of 8 actually set the semconv `error.type` attribute:

| Package | record_exception | set_status | ERROR_TYPE |
|---|---|---|---|
| anthropic | ✅ | ✅ | ✅ |
| groq | ✅ | ✅ | ✅ |
| mistralai | ✅ | ✅ | ✅ |
| bedrock | ✅ | ✅ | ❌ |
| ollama | ✅ | ✅ | ❌ |
| sagemaker | ✅ | ✅ | ❌ |
| together | ✅ | ✅ | ❌ |
| langchain `_handle_error` | ✅ | ✅ | ❌ |

`error.type` is the canonical attribute backends use to bucket failures per the OTel GenAI semconv. The tests already reflect this drift — groq/mistralai assert it, bedrock/ollama/sagemaker/together don't. Please add `span.set_attribute(ERROR_TYPE, e.__class__.__name__)` everywhere (and to `_handle_error` in the langchain callback handler).

### 2. Sagemaker streaming wrapper has no error handling
Only `_instrumented_endpoint_invoke` got the try/except — `_instrumented_endpoint_invoke_with_response_stream` is untouched. Errors from streaming invocations will still produce UNSET-status spans, so issue #412 is only partly closed for sagemaker. Bedrock's streaming wrappers correctly got the same treatment; sagemaker's should too.

### 3. Together test patches `requests.Session.request` while everything else patches `httpx.Client.send`
Together SDK ≥1.2 does use `requests`, so this is probably correct — but please confirm by running the test locally and verifying it actually fails without the instrumentation change (a green test against an unpatched path is a silent regression risk). If it does exercise the right path, leave it; otherwise switch to `httpx`.

### 4. `raise e` vs bare `raise` inconsistency
- `raise e`: anthropic, groq
- `raise`: mistralai, ollama, sagemaker, bedrock, together

Bare `raise` is the idiomatic re-raise and preserves the original traceback. Recommend `raise` everywhere.

### 5. Groq's `is_recording()` guard is only partial
```python
span.set_attribute(ERROR_TYPE, e.__class__.__name__)   # unguarded
span.record_exception(e)                                # unguarded
if span.is_recording():
    span.set_status(Status(StatusCode.ERROR, str(e)))   # guarded
span.end()
```
Either guard all three or none. OpenAI's reference (the pattern this PR claims to follow) guards none — recommend dropping the guard for symmetry with the rest of the PR.

### 6. Anthropic test churn — defensive filter added to ~25 unrelated tests
```python
spans = [s for s in span_exporter.get_finished_spans() if s.status.status_code != StatusCode.ERROR]
```
This pattern is repeated across `test_messages.py`, `test_prompt_caching.py`, `test_thinking.py`. The reason it's necessary: many of those tests have `try: ... except Exception: pass` blocks that previously hid failures silently, and now produce error spans the assertions trip over. The filter works but masks the real question — are those tests actually exercising the success path they claim? Two ways to make this less smelly:
- Extract a `successful_spans(exporter)` helper to DRY the filter
- Better: fix the tests that are silently swallowing exceptions, so they don't need a filter at all

---

Re: the goal of "simple and robust" — items 1, 2, and 4 are the ones that matter most. Once they're addressed, the pattern across all 8 packages becomes a single rule: *on exception → set ERROR_TYPE, record_exception, set_status(ERROR, msg), end span, bare raise.* That's the version that's easy to reason about and easy to extend to the next instrumentation.

packages/opentelemetry-instrumentation-together/tests/test_error_handling.py
```diff
@@ -0,0 +1,22 @@
+from unittest.mock import patch
+
+import pytest
+from opentelemetry.trace.status import StatusCode
+
+
+def test_together_error_sets_span_status(instrument_legacy, together_client, span_exporter):
+    with patch("requests.Session.request", side_effect=Exception("API connection error")):
```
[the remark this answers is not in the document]
dvirski: this is a false positive for the version under test. This instrumentation pins together >= 1.2.0, < 2 (see _instruments in __init__.py and the together>=1.2.0,<2 entry in pyproject.toml), and together v1.x uses requests.Session() as its HTTP backend (see 
  ▎ together/abstract/api_requestor.py in the installed package: import requests and requests.Session() in _make_session). The httpx switch you cited landed in together v2.0, which is excluded from our supported range here. The patch target is correct, and the test passes because requests.Session.request 
  ▎ does intercept the real HTTP call. Happy to revisit if/when this instrumentation adds v2 support.

dvirski:
@doronkopit5 Addressed all PR comments.
This is a summary of the changes that have been done:

Production code — error handling
  1. bedrock — added ERROR_TYPE import + set on all 4 exception blocks; changed raise e → raise
  2. ollama — added ERROR_TYPE import + set on both sync/async exception blocks 
  3. together — added ERROR_TYPE import + set on exception block
  4. sagemaker — added ERROR_TYPE import + set on non-streaming path; added full error handling (try/except, ERROR status, span.end) to the streaming wrapper
  5. langchain — moved ERROR_TYPE set from on_tool_error into shared _handle_error so all 5 error callbacks set it 
  6. groq — removed inconsistent is_recording() guard around set_status in error blocks; raise e → raise
  7. anthropic — restored lost ERROR_TYPE import (was dropped in a merge conflict); raise e → raise in both __init__.py exception blocks
  8. anthropic streaming — raise e → raise

  Tests — new dedicated error-handling coverage
  9. anthropic — created tests/test_error_handling.py with sync + async tests asserting ERROR status, error.type, exception event, exception.type
  10. langchain — created tests/test_error_handling.py building a real RunnableSequence (prompt | llm), patching httpx to raise, asserting at least one langchain span has ERROR + error.type +
  exception event (verified via regression check that it actually exercises _handle_error)

  Tests — assertion additions
  11. bedrock — added error.type assertions to both existing error tests                                              
  12. ollama — added error.type to both tests + exception.message parity assertion to async test
  13. sagemaker — added error.type assertion + new streaming-path error test
  14. together — added error.type assertion to the existing error test

  Tests — anthropic cleanup (the big one)
  
  15. test_messages.py — removed 24 try/except: pass bad-kwarg blocks + 27 defensive filters + unused StatusCode import
  16. test_prompt_caching.py — removed 12 bad-kwarg blocks + 12 filters + unused import
  17. test_thinking.py — removed 12 bad-kwarg blocks + 12 filters + unused import
  18. test_completion.py — removed 3 bad-kwarg blocks
  19. tests/utils.py — cleaned verify_metrics: removed exception-metric enforcement and the or error.type=="TypeError" tolerance clause

  Why the anthropic test cleanup was needed

  Doron flagged that the original PR added a defensive filter ([s for s in spans if s.status.status_code != StatusCode.ERROR]) in 65 places across the anthropic test files. The filter was bridging a
   side effect of this PR — but the deeper issue was a pre-existing test idiom that had been there for years:

  try:
      anthropic_client.messages.create(unknown_parameter="unknown")
  except Exception:
      pass

  The original intent was a smoke test — "verify our instrumentation doesn't crash when the SDK rejects bad input." But on inspection it didn't actually test that, because:
  - except Exception: pass swallows everything — the SDK's expected TypeError, any bug in our wrapper, and the case where nothing raises at all. All three scenarios produced identical test behavior.
  - Pre-PR, the bad call's span was started but never .end()-ed (the wrapper re-raised before ending it), so it never reached the exporter. The existing tests passed because get_finished_spans()
  returned only the success span.
  - This PR's instrumentation now correctly ends the span and marks it ERROR on the exception path → those previously-invisible spans suddenly become visible to get_finished_spans(), breaking tests
  that assumed one call = one finished span. The filter was added as a quick fix to hide them.
  
  The cleanup removes both the cause and the symptom: the bad-call blocks are deleted (they verified nothing), so no error spans get produced, so no filter is needed, and the legacy len(spans) == N
  assertions remain accurate without modification. The new dedicated test_error_handling.py (item 9) provides the proper test coverage for the error-span behavior — explicitly asserting on ERROR
  status, error.type, and exception events using a meaningful failure mode (mocked HTTP failure), rather than the prior accidental smoke test that swallowed everything.

  verify_metrics was also touched (item 19) because it had grown a coupling to the bad-call pattern: it required the LLM_ANTHROPIC_COMPLETION_EXCEPTIONS metric to have been incremented and had a
  tolerance clause for error.type=="TypeError" duration data points. Both existed only because every caller happened to make a failing call. Now that the bad calls are gone, those assertions become
  dead requirements that would block the cleanup.
  
  Verified but not changed

  20. together test patch target — confirmed requests.Session.request is correct for together<2 (the pinned range); ran a regression check showing the test fails without our instrumentation change

  Total: 51 bad-kwarg blocks removed, 51 defensive filters removed, ~440 net lines deleted in test cleanup, plus the production fixes and new error-span coverage.

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
```diff
@@ -259,7 +259,7 @@ def __next__(self):
             attributes = error_metrics_attributes(e)
             if self._exception_counter:
                 self._exception_counter.add(1, attributes=attributes)
-            raise e
+            raise
```
nina-kollman, line 262: Why this change ? 
Isnt is safer to rase the exception itself ?

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -210,7 +212,7 @@ def _wrap(
             if metric_params.exception_counter:
                 metric_params.exception_counter.add(1, attributes=attributes)
 
-            raise e
+            raise
```
nina-kollman, line 215: Same

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/streaming.py
dvirski: @nina-kollman 
We need bare raise here so the logged exception exact real line will show (if for example it fails in the providers sdk code), and not in our instrumentation wrapper.

safety wise they are identical. e is just a name bound to the exception that's already active. Bare raise re-raises that same exception.

this follows Doron suggestion 

4. raise e vs bare raise inconsistency
raise e: anthropic, groq
raise: mistralai, ollama, sagemaker, bedrock, together
Bare raise is the idiomatic re-raise and preserves the original traceback. Recommend raise everywhere.

nina-kollman (approved):

PR #4103, opened 2026-05-10
fix(mcp): set error.type attribute on protocol-level tool errors

### LINKED ISSUE #4037: 🐛 Bug Report: MCP instrumentation: Missing `error.type` attribute for protocol-level errors

Traceloop SDK

### Summary
The MCP instrumentation in `opentelemetry-instrumentation-mcp` does not set the `error.type` attribute for MCP protocol-level errors (when `result.isError=True`), only for Python exceptions. This creates inconsistent error reporting and makes it difficult to filter, categorize, and alert on MCP tool errors in observability platforms.

### Proposed Solution

Add `error.type` attribute for MCP protocol-level errors:

```python
if hasattr(result, "isError") and result.isError:
    # Set error.type for MCP protocol errors
    span.set_attribute(ERROR_TYPE, "MCPToolError")
    if len(result.content) > 0:
        span.set_status(Status(StatusCode.ERROR, f"{result.content[0].text}"))
```

Alternatively, extract more specific error types from the MCP response if available (e.g., `"InvalidInput"`, `"NotFound"`, `"PermissionDenied"`).

### Impact

**Current impact:**
- Users cannot filter MCP protocol errors by type in observability platforms (e.g., Instana, Datadog)
- Inconsistent error reporting between exception-based and protocol-level errors
- Difficult to create targeted alerts for specific MCP error conditions
- Reduced observability for MCP tool failures

### Environment
- Package: `opentelemetry-instrumentation-mcp`
- File: `opentelemetry/instrumentation/mcp/instrumentation.py`
- Method: `_execute_and_handle_result()` (lines 318-353)

---

### 👟 Reproduction steps

This issue was discovered when analyzing MCP tool call traces in Instana UI, where some error spans had the `error.type` tag while others (with MCP protocol errors) did not, making it difficult to understand and categorize failures consistently.

### 👍 Expected behavior

According to OpenTelemetry semantic conventions, the `error.type` attribute should be set for all error spans, regardless of whether the error is an exception or application-level error.

MCP protocol errors should also set the `error.type` attribute to enable:
- Consistent error categorization across all error types
- Filtering and grouping by error type in observability platforms
- Alerting based on specific error categories
- Better root cause analysis

**Exception-based errors** (lines 350-353 in `instrumentation.py`):
```python
except Exception as e:
    span.set_attribute(ERROR_TYPE, type(e).__name__)  # ✅ Sets error.type
    span.record_exception(e)
    span.set_status(Status(StatusCode.ERROR, str(e)))
    raise
```
- Sets `error.type` attribute (e.g., `"ConnectionError"`, `"ValueError"`)
- Visible in observability UI with error type tag

**MCP protocol-level errors** (lines 343-348 in `instrumentation.py`):
```python
if hasattr(result, "isError") and result.isError:
    if len(result.content) > 0:
        span.set_status(Status(StatusCode.ERROR, f"{result.content[0].text}"))
```
- Does NOT set `error.type` attribute ❌
- Only sets span status to ERROR
- Error visible in UI but without error type categorization

### 🤖 Python Version

_No response_

_No response_

dvirski:
Fixes #4037.
When a tool returns isError=True, the span was marked ERROR but missing                                                                                                                       
the error.type attribute required by OTel semconv.  

## Summary by CodeRabbit

* **New Features**
  * Classifies protocol-level tool-call failures in traces by setting the error type to "tool_error" while preserving exception-type reporting for raised exceptions.

* **Tests**
  * Added an async test to verify tool-call failures set error.type on client spans and that error.type is present for ERROR spans.
  * Adjusted test fixtures to enable instrumentation for the whole session and clear spans per-test for reliable span inspection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4103?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
@@ -341,6 +341,7 @@ class McpInstrumentor(BaseInstrumentor):
                 )
             # Handle errors
             if hasattr(result, "isError") and result.isError:
+                span.set_attribute(ERROR_TYPE, "tool_error")
                 if len(result.content) > 0:
                     span.set_status(
                         Status(StatusCode.ERROR, f"{result.content[0].text}")
@@ -569,6 +570,7 @@ class InstrumentedStreamWriter(ObjectProxy):  # type: ignore
                 )
                 if "isError" in request.result:
                     if request.result["isError"] is True:
+                        span.set_attribute(ERROR_TYPE, "tool_error")
                         span.set_status(
                             Status(
                                 StatusCode.ERROR,
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
@@ -570,7 +570,6 @@ class InstrumentedStreamWriter(ObjectProxy):  # type: ignore
                 )
                 if "isError" in request.result:
                     if request.result["isError"] is True:
-                        span.set_attribute(ERROR_TYPE, "tool_error")
                         span.set_status(
                             Status(
                                 StatusCode.ERROR,
--- a/packages/opentelemetry-instrumentation-mcp/tests/conftest.py
+++ b/packages/opentelemetry-instrumentation-mcp/tests/conftest.py
@@ -21,12 +21,19 @@ def fixture_tracer_provider(span_exporter):
     provider.shutdown()
 
 
-@pytest.fixture(autouse=True)
-def instrument_mcp(tracer_provider, span_exporter):
+# Session-scoped: McpInstrumentor wraps process-level imports (BaseSession.send_request,
+# post-import hooks on fastmcp.client, etc.). Re-instrumenting per test stacks wrappers
+# because _uninstrument doesn't fully tear them down — that breaks tests whose assertions
+# depend on a single wrapping layer (e.g. trace-id sharing in test_fastmcp.py).
+@pytest.fixture(scope="session", autouse=True)
+def instrument_mcp(tracer_provider):
     instrumenter = McpInstrumentor()
     instrumenter.instrument(tracer_provider=tracer_provider)
-    try:
-        yield
-    finally:
-        instrumenter.uninstrument()
-        span_exporter.clear()
+    yield
+    instrumenter.uninstrument()
+
+
+@pytest.fixture(autouse=True)
+def clear_spans(span_exporter):
+    yield
+    span_exporter.clear()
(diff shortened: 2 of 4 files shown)


doronkopit5:
Thanks for the fix — the intent is right and the change at `instrumentation.py:344` is exercised correctly by the new test (verified locally: a `fail_tool.tool` client-side span now carries `error.type='tool_error'`). A few small things to tighten before merge:

1. **Test docstring is misleading.** It describes the asserted span as "server-side", but the new fix actually lives on the **client-side** path (`_handle_tool_call` → `_execute_and_handle_result`). The server-side `fail_tool.tool` span already has `error.type` set via `fastmcp_instrumentation.py`'s exception handler (it captures `type(e).__name__`, e.g. `"ToolError"`). Worth rewording the docstring/comments so future readers aren't sent in the wrong direction.

2. **Extract the magic string.** `"tool_error"` appears in two places (`instrumentation.py:344` and `:573`). Hoisting it to a module-level constant (e.g. `_MCP_TOOL_ERROR_TYPE = "tool_error"`) keeps the two sites in lockstep if the value ever changes. The lowercase namespace style is fine per OTel semconv (matches the spec's `auth_error` example, low cardinality) — just want it defined once.

3. **No test for the second fix site.** The patch in `ResponseStreamWriter.send` (line 573) isn't covered by `test_error_type.py` — the in-process FastMCP `Client(server)` scenario doesn't produce a `ResponseStreamWriter` span. Either add a test that exercises that path (e.g. via a stdio/streamable transport where the response-stream writer is actually used), or, if you can't reach it from a test, consider whether that branch is reachable in practice and drop it if not. Untested dead branches are how regressions slip in later.

dvirski:
@doronkopit5 

 Changes made following your comment:

- Test docstring corrected. Renamed test_tool_error_sets_error_type_on_server_span → ..._on_client_span. Rewrote docstring and inline comments to correctly identify instrumentation.py:344 as the client-side fix site, with a note explaining both client- and server-side fail_tool.tool spans exist.

 - Dropped the line 573 change. Verified against issue #4037 that the bug is scoped to _execute_and_handle_result() (client side). The ResponseStreamWriter.send patch was scope creep, untestable from in-process Client(server), and the FastMCP server path is already covered elsewhere.
 
- Magic string clarified. Briefly extracted "tool_error" to a module-level constant, then inlined it back after comment #3 left only one call site (avoiding premature abstraction). Added an inline comment explaining why "tool_error" is used rather than type(e).__name__.

Fixed CI failure discovered during the session:
- Diagnosed test-pollution bug in test_fastmcp.py::test_fastmcp_instrumentor: the autouse instrument_mcp fixture was re-wrapping BaseSession.send_request on every test because _uninstrument doesn't fully tear it down — stacking wrappers broke trace-id sharing assertions. The new
test_error_type.py exposed it by being alphabetically ordered before test_fastmcp.py.

- Patched tests/conftest.py — moved McpInstrumentor setup/teardown to session scope (matching OTel's process-level design contract), added a function-scoped clear_spans fixture for per-test span isolation. Deeper _uninstrument fix left as a separate concern.

Validation: 5/5 tests pass, ruff clean.

doronkopit5 (approved):

Issue #1870, opened 2024-08-19
🐛 Bug Report: Inconsistency in recorded data across different vector databases


LakshmiN5:
### Which component is this bug for?

All Packages

### 📜 Description

Tried using Traceloop version 0.26.4 with different vector databases while trying to run a RAG application (watsonx + langchain) and observed some difference in behaviour. I was expecting the behaviour across all vector dbs to be uniform with regards to the span information. I tested using Milvus, Pinecone and Chroma where Milvus and Chroma were both tested using in memory option with langchain and for pinecone I tried the managed instance.
**Observations** :

- Chroma - captures less information in the vector db related spans - captures embedding count , gives similarity value but does not give all 4 retrieved chunks information. It just returns one chunk and  specifying the parameters in the as_retriever() method does not seem to have any effect on the span information collected. reference for as_retriever: [https://api.python.langchain.com/en/v0.1/vectorstores/langchain_astradb.vectorstores.AstraDBVectorStore.html#langchain_astradb.[…]Store.as_retriever](https://api.python.langchain.com/en/v0.1/vectorstores/langchain_astradb.vectorstores.AstraDBVectorStore.html#langchain_astradb.vectorstores.AstraDBVectorStore.as_retriever)
- Pinecone - Does not capture the embedding count nor does it give similarity value but I could see the top 4 retrieved documents as part of another span.
- Milvus - does not capture the embedding count nor does it give the similarity value , there seems to be some problem with the retrieved context also.

### 👟 Reproduction steps

Steps can be reproduced by trying the RAG sample from langchain using different vector databases. LLM used is from watsonx with langchain framework.

https://python.langchain.com/v0.1/docs/use_cases/question_answering/quickstart/

### 👍 Expected behavior

Ideally the following information should be captured consistently across all vector dbs

1. embeddings details - such as count (for the stored knowledge base) and any additional information 
2. query embeddings and other details
3. retrieved context information - no of chunks matched, should return all matched chunks as per the configuration parameters set for the retriever (mentioned in number 5)
4. retrieval parameters configured should influence the actual results generated , for eg : similarity algorithm to use for searching query against the stored docs, number of documents to retrieve , similarity threshold etc
5. any insights on the chunk(s) used for the final answer generation.

### 👎 Actual Behavior with Screenshots

Most information is missing and the behaviour is definitely not consistent.

### 🤖 Python Version

3.10

### 👀 Have you spent some time to check if this bug has been raised before?

[...]

PR #4105, opened 2026-05-11
fix(chromadb): emit one result event per document across all queries

dvirski:
Related to #1870

problem:                                                                                                                                                                        
ChromaDB's query() returns results as a list-of-lists (one inner list per query embedding). The instrumentation only unzipped the outer list, producing one db.query.result     
event per query instead of one per result document. A second bug also indexed into each attribute value with [0], so string IDs were silently truncated to their first          
character.                                                                                                                                                                      
                                                                                                                                                                                  
fix:                                                                                                                                                                            
Added an inner loop over each query's result list so every document gets its own span event. Removed the erroneous [0] indexing so attribute values are used as-is.


## Summary by CodeRabbit

* **Bug Fixes**
  * Query telemetry now emits one event per returned document (N×K for multi-query), with event attributes included only when present and metadata serialized when it's a dictionary.

* **Tests**
  * Added and updated tests to assert per-result event counts and validate each event's attributes for single and multi-embedding queries.

* **Documentation**
  * Clarified the nested query-result structure and the resulting per-result event behavior.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4105?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-chromadb/opentelemetry/instrumentation/chromadb/wrapper.py
+++ b/packages/opentelemetry-instrumentation-chromadb/opentelemetry/instrumentation/chromadb/wrapper.py
@@ -223,68 +223,50 @@ def _add_segment_query_embeddings_events(span, kwargs, args=None):
 @dont_throw
 def _add_query_result_events(span, kwargs):
     """
-    There's a lot of logic here involved in converting the query result
-    format from ChromaDB into the canonical format (taken from Pinecone)
-
-    This is because Chroma query result looks like this:
+    ChromaDB query results have a nested structure — one inner list per query:
 
         {
-           ids: [1, 2, 3...],
-           distances: [0.3, 0.5, 0.6...],
-           metadata: ["some metadata text", "another metadata text",...],
-           documents: ["retrieved text", "retrieved text2", ...]
+            "ids":       [["id1", "id2", "id3"]],   <- outer: per query, inner: per result
+            "distances": [[0.1,   0.2,   0.3  ]],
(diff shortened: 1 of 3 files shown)


doronkopit5:
PLease take this small comment and add
> > packages/opentelemetry-instrumentation-chromadb/tests/test_query_results.py (1)> `54-76`: _⚡ Quick win_
> > > **Use multi-character IDs so this test actually exercises the `[0]` truncation fix.**
> > > This PR fixes two bugs: (1) event count and (2) the previous `[0]` indexing that truncated string IDs to a single character. With single-character IDs (`"1"`, `"2"`), `"1"[0] == "1"`, so the id-equality assertion below would still pass against the _old_ buggy code — i.e., this test only protects against the count regression, not the truncation regression. Switching to multi-character IDs makes the assertion meaningfully cover the truncation fix.

dvirski:
@doronkopit5 

Updated the third test in test_query_results.py to use multi-character IDs ("doc-id-aaa", "doc-id-bbb") instead of single-character ones ("1", "2").
  
Why: the PR fixed an ids[0] truncation bug, but single-character IDs hid it — "1"[0] == "1" so the assertion passed even against the broken code. Multi-character IDs make ids[0] collapse to just "d", so the test now actually fails under the old bug and genuinely covers the truncation fix (not
just the event-count fix).

doronkopit5 (approved):

PR #4130, opened 2026-05-12
fix(openai-agents): emit cache_read.input_tokens and reasoning_tokens

### LINKED ISSUE #4060: 🐛 Bug Report: openai-agents instrumentation doesn't record cached input tokens

dvirski:
Fixes #4060

Problem: 
gen_ai.usage.reasoning_tokens was the wrong attribute name — the OTel GenAI spec defines it as gen_ai.usage.reasoning.output_tokens. Similarly, the LLM_USAGE_CACHE_READ_INPUT_TOKENS constant resolved to gen_ai.usage.cache_read_input_tokens (underscore) instead of   
the spec-correct gen_ai.usage.cache_read.input_tokens (dot). The openai-agents instrumentor also wasn't recording cache read or reasoning tokens at all. The opentelemetry-semantic-conventions floor was too old (0.60b1) to provide the official upstream cache token constants. 

Fix: 
Bumped opentelemetry-semantic-conventions to >=0.62b1 so cache token constants come from the upstream OTel package. Added GEN_AI_USAGE_REASONING_OUTPUT_TOKENS = "gen_ai.usage.reasoning.output_tokens" to semconv_ai and removed the old incorrect constant. Updated the     
openai and openai-agents instrumentors to use the corrected constants. Added two new tests with recorded cassettes for cache read and reasoning token recording in openai-agents.        

## Summary by CodeRabbit

* **New Features**
  * Instrumentation now records cached input token counts and reasoning token counts for OpenAI agent calls.

* **Dependencies**
  * Minimum required opentelemetry-semantic-conventions updated to 0.62b1.

* **Tests**
  * Added unit tests to verify the new token-tracking attributes are emitted.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4130?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


doronkopit5:
Reviewed the change — direction is right (aligning attribute names with the OTel GenAI spec). A few things worth addressing:

### 1. Two sources of truth for the cache constants
The same constants exist in both places, resolving to the same string today:
- `opentelemetry.semconv._incubating.attributes.gen_ai_attributes.GEN_AI_USAGE_CACHE_*_INPUT_TOKENS` (upstream OTel)
- `opentelemetry.semconv_ai.SpanAttributes.GEN_AI_USAGE_CACHE_*_INPUT_TOKENS` (this repo)

The PR mixes both: `_hooks.py` and `manual.py` use `GenAIAttributes` (upstream), while pre-existing code in `chat_wrappers.py`, `responses_wrappers.py`, `anthropic`, `langchain` uses `SpanAttributes` from semconv_ai. Both work, but it's nicer to have one source of truth.

I'd lean toward standardizing on the **upstream OTel-native** `GenAIAttributes` here — that's the spec source — but happy to leave the call to you. Whatever you pick, please apply it consistently across the touched files so callers don't have to know which path leads where.

### 2. `traceloop-sdk` needs an explicit upstream semconv floor
`manual.py` now uses `GenAIAttributes.GEN_AI_USAGE_CACHE_*_INPUT_TOKENS`, but `packages/traceloop-sdk/pyproject.toml` doesn't pin `opentelemetry-semantic-conventions`. With `<0.62b1` installed transitively, that attribute either doesn't exist (→ `AttributeError`) or has the old underscore string (silently wrong telemetry).

Please add `opentelemetry-semantic-conventions>=0.62b1` to `packages/traceloop-sdk/pyproject.toml` (same bump you did for `openai-agents`).

### 3. Removing `GEN_AI_USAGE_REASONING_TOKENS` is a breaking change with no alias
Deleting the class attribute from `SpanAttributes` will break any downstream code importing `SpanAttributes.GEN_AI_USAGE_REASONING_TOKENS` with `AttributeError`. The repo already has prior art for graceful renames (see the "Legacy LLM_* aliases" block in `semconv_ai/__init__.py`). Suggest keeping a deprecation alias:

```python
GEN_AI_USAGE_REASONING_OUTPUT_TOKENS = "gen_ai.usage.reasoning.output_tokens"
# Deprecated alias; use GEN_AI_USAGE_REASONING_OUTPUT_TOKENS
GEN_AI_USAGE_REASONING_TOKENS = GEN_AI_USAGE_REASONING_OUTPUT_TOKENS
```

Also note that the **string value** changes too (`gen_ai.usage.reasoning_tokens` → `gen_ai.usage.reasoning.output_tokens`), which is a wire-level break for dashboards. Worth a line in `MIGRATION.md` under the "Changed string values" section.

### 5. New test assertions don't actually verify the feature
```python
assert response_span.attributes[GenAIAttributes.GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS] >= 0
assert response_span.attributes[SpanAttributes.GEN_AI_USAGE_REASONING_OUTPUT_TOKENS] >= 0
```

Both `>= 0` assertions are trivially true once the attribute is recorded. The cache test is intentionally structured (first call writes cache → `clear()` → second call reads) precisely to exercise caching — so it should assert `> 0` to prove the second call actually got a cache hit. Same for o3-mini reasoning tokens, which per the docstring always reports `> 0`. Tightening to `> 0` makes these tests actually catch regressions.

dvirski:
@doronkopit5 

 Changes in response to review feedback

  - Reverted out-of-scope changes — Removed edits to traceloop-sdk/manual.py, openai instrumentor (chat_wrappers.py, responses_wrappers.py, test_azure.py, test_chat.py), and the
  GEN_AI_USAGE_REASONING_TOKENS rename in semconv_ai. PR now contains only changes required for #4060.
  - Resolves CR #4 / Doron #3 (reasoning rename breaking change) — Reverted; constant name and string value unchanged from main. No alias needed.
  - Resolves Doron #2 (traceloop-sdk semconv floor) — Moot after reverting manual.py.
  - Resolves CR #5 / Doron #5 (weak >= 0 assertions) — Replaced both VCR tests with two deterministic unit tests that call _extract_response_attributes directly with a documented fake Response
   and assert exact emitted values (1024, 256). Fake-response shape is documented inline against the openai-agents SDK source (v0.14.2 pinned floor), a real Responses API payload in an
  existing cassette in this repo, and the OpenAI API reference.
  - Resolves CR #1, #2, #3 (cookie/sensitive headers in cassettes) — Deleted both new cassettes. Unit tests need no VCR, no API key.
  - Resolves CR nitpick #6 (brittle response_spans[0]) — Gone with the VCR tests; new tests use a MagicMock span directly.
  - Partially addresses Doron #1 (single source of truth) — _hooks.py uses upstream GenAIAttributes.GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS for cache. Reasoning continues to use
  SpanAttributes.GEN_AI_USAG

doronkopit5:
Re-reviewed after the latest push (`86e2d74` + `d77ffe4e`). Major improvements — happy with the direction:

### ✅ Resolved
- **Scope tightened**: PR is now 82 lines of code change in 3 files (excluding lock-file churn), focused squarely on #4060. The `manual.py` change, `chat_wrappers/responses_wrappers` rename, and `SpanAttributes` rename are all out — that addresses my #2, #3, and #4 cleanly.
- **Tests are now meaningful**: `SimpleNamespace` fakes with exact value assertions (1024, 256) are much better than the `>= 0` VCR tests. The docstring linking the fake shape to the SDK source, the live cassette, and the OpenAI docs is nice — easy to keep in sync. Also drops the cassettes with `Cookie` / `set-cookie` headers, which is a separate win.
- **Cache fix using upstream constant**: `GenAIAttributes.GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS` is the right pick — matches the OTel-native preference and the `>=0.62b1` pyproject floor backs it correctly (verified: that constant only exists in upstream from 0.62b1 onward, with the dot-form string).

### ⚠️ Worth a small follow-up
1. **PR title and description are now stale**. The title still says "*align reasoning and cache token attributes with OTel GenAI spec*" and the description still claims "*Added GEN_AI_USAGE_REASONING_OUTPUT_TOKENS …*" — but the reasoning rename was reverted. After this PR, `_hooks.py` will emit `gen_ai.usage.reasoning_tokens` (the non-spec string from `SpanAttributes.GEN_AI_USAGE_REASONING_TOKENS`), not `gen_ai.usage.reasoning.output_tokens`. Please retitle to something like `fix(openai-agents): record cached input + reasoning tokens (#4060)` and trim the description to match.

2. **Reasoning attribute string is still non-spec** — for awareness, not blocking this PR. The OTel GenAI spec calls for `gen_ai.usage.reasoning.output_tokens`, and the upstream 0.62b1 package doesn't yet define a constant for it (verified). Once the semconv_ai constant is corrected (or the upstream catches up), `_hooks.py` will pick up the right string automatically since it uses the named constant. Worth tracking as a follow-up issue so it's not forgotten.

3. **Tests don't exercise the None-guard paths** — minor. Both `input_tokens_details is None` and `cached_tokens is None` short-circuits go untested. Two extra one-liners would lock in the guard behavior:
   ```python
   def test_extract_response_attributes_no_cache_when_details_missing():
       response = _make_fake_response()  # both None
       span = MagicMock()
       _extract_response_attributes(span, response, trace_content=False)
       for call in span.set_attribute.call_args_list:
           assert call.args[0] != GenAIAttributes.GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS
           assert call.args[0] != SpanAttributes.GEN_AI_USAGE_REASONING_TOKENS
   ```
   Not blocking — the existing tests cover the positive paths well.

LGTM once the title/description are updated. Nice cleanup.

doronkopit5 (approved):
just fix the last comment form code rabbit about the non zero attribute

dvirski:
> just fix the last comment form code rabbit about the non zero attribute

@doronkopit5 

isn't zero cache token also meaningful? like it doesn't mean "no cache hit".


Adding a check summary from otel spec:

Spec check (semconv reference (https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)): gen_ai.usage.cache_read.input_tokens is classified as Recommended with no "when applicable" qualifier — the same level as gen_ai.usage.input_tokens and gen_ai.usage.output_tokens. The spec 
doesn't suggest omitting it when zero. Suppressing 0 would make "no cache hit" indistinguishable from "instrumentor doesn't capture it," and break aggregations like cache-hit-rate. Matching how the existing openai instrumentor emits this attribute is also the safer choice for trace consistency.
Keeping is not None as the guard.

doronkopit5:
> > just fix the last comment form code rabbit about the non zero attribute
> 
> @doronkopit5
> 
> isn't zero cache token also meaningful? like it doesn't mean "no cache hit".
> 
> Adding a check summary from otel spec:
> 
> Spec check (semconv reference (https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)): gen_ai.usage.cache_read.input_tokens is classified as Recommended with no "when applicable" qualifier — the same level as gen_ai.usage.input_tokens and gen_ai.usage.output_tokens. The spec doesn't suggest omitting it when zero. Suppressing 0 would make "no cache hit" indistinguishable from "instrumentor doesn't capture it," and break aggregations like cache-hit-rate. Matching how the existing openai instrumentor emits this attribute is also the safer choice for trace consistency. Keeping is not None as the guard.

Get, your point
LGTM

PR #4133, opened 2026-05-12
feat(sdk): expose use_legacy_attributes via Traceloop.init()

### LINKED ISSUE #3236: 🐛 Bug Report: No way to set `use_legacy_attributes` using TraceLoop.init()

OpenAI Instrumentation

## What's the issue? 

Although most of the instrumentors support the flag `use_legacy_attributes` to emit prompts input&output as events, there is no way to pass this flag through `TraceLoop.init()`.  

People have to init traceProvider, logProvider, metricProvider manually, and list down all the instrumentors that are needed. 

```
# 1. Define Resource
resource = Resource(attributes={SERVICE_NAME: "ying-test"})

# 2. configure trace provider
tracer_provider = TracerProvider(resource=resource)
trace.set_tracer_provider(tracer_provider)
span_exporter = OTLPSpanExporterHTTP(endpoint="http://127.0.0.1:4320/v1/traces")
tracer_provider.add_span_processor(BatchSpanProcessor(span_exporter))

# 3 configure log provider
logger_provider = LoggerProvider(resource=resource)
_logs.set_logger_provider(logger_provider)
log_exporter = OTLPLogExporterHTTP(endpoint="http://127.0.0.1:4320/v1/logs")
logger_provider.add_log_record_processor(BatchLogRecordProcessor(log_exporter))

OpenAIInstrumentor(
    use_legacy_attributes=False,
).instrument(
    event_logger_provider=EventLoggerProvider()
)
```

## What's the root cause?

1. The function `Traceloop.init()` does not allow this flag as an argument.
2. When we init instrumentors [here](https://github.com/traceloop/openllmetry/blob/main/packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py#L494), there is no parameter for `use_legacy_attributes`.  

## Why is this a Bug rather than feature request?

As we have closed this [issue](https://github.com/traceloop/openllmetry/issues/2456), people expect this feature is supported through TraceLoop whereas it is not due to no way to configure it. 

### 👟 Reproduction steps

When we tried to send input&output as event, we met this issue. 

### 👍 Expected behavior

```
Traceloop.init(use_legacy_attributes=False)
```

```
# 1. Define Resource
resource = Resource(attributes={SERVICE_NAME: "ying-test"})

# 2. configure trace provider
tracer_provider = TracerProvider(resource=resource)
trace.set_tracer_provider(tracer_provider)
span_exporter = OTLPSpanExporterHTTP(endpoint="http://127.0.0.1:4320/v1/traces")
tracer_provider.add_span_processor(BatchSpanProcessor(span_exporter))

# 3 configure log provider
logger_provider = LoggerProvider(resource=resource)
_logs.set_logger_provider(logger_provider)
log_exporter = OTLPLogExporterHTTP(endpoint="http://127.0.0.1:4320/v1/logs")
logger_provider.add_log_record_processor(BatchLogRecordProcessor(log_exporter))

OpenAIInstrumentor(
    use_legacy_attributes=False,
).instrument(
    event_logger_provider=EventLoggerProvider()
)
```

### 🤖 Python Version

_No response_

Please also advice how should we create the EventLoggerProvider? should we set it as an argument for `Traceloop.init` or, we just create it when `use_legacy_attributes` is set to False. 

dvirski:
Fixes #3236.
 
 Problem: use_legacy_attributes existed on every instrumentation but was
 unreachable through the SDK — users were locked to legacy gen_ai.prompt/
 gen_ai.completion span attributes with no way to opt into the new
 event-based format.

 Fix: thread use_legacy_attributes=True (default, no behaviour change) from
 Traceloop.init() through TracerWrapper, init_instrumentations(), and all
 8 instrumentations that support the flag: openai, anthropic, bedrock,
 sagemaker, groq, langchain, together, watsonx.

## Summary by CodeRabbit

* **New Features**
  * Added use_attributes (default: true) to control whether prompt/completion data is emitted as span attributes or as events; this setting now propagates across all instrumentations and the deprecated use_legacy_attributes alias is supported.

* **Bug Fixes**
  * Passing both use_attributes and use_legacy_attributes now raises a TypeError; using use_legacy_attributes emits a DeprecationWarning.

* **Tests**
  * Added initialization tests for defaults, propagation, deprecation warning, and mutual-exclusion error.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4133?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -3,7 +3,10 @@
 import logging
 import os
 import time
+import warnings
 from typing import Callable, Collection, Optional
+
+_USE_ATTRIBUTES_UNSET = object()
 
 from opentelemetry import context as context_api
 from opentelemetry._logs import Logger, get_logger
@@ -796,18 +799,31 @@ class AnthropicInstrumentor(BaseInstrumentor):
         self,
         enrich_token_usage: bool = False,
         exception_logger=None,
-        use_legacy_attributes: bool = True,
+        use_attributes: bool = True,
         get_common_metrics_attributes: Callable[[], dict] = lambda: {},
         upload_base64_image: Optional[
             Callable[[str, str, str, str], Coroutine[None, None, str]]
         ] = None,
+        use_legacy_attributes=_USE_ATTRIBUTES_UNSET,
     ):
         super().__init__()
+        if use_legacy_attributes is not _USE_ATTRIBUTES_UNSET:
+            warnings.warn(
+                "`use_legacy_attributes` is deprecated and will be removed in a "
+                "future release; use `use_attributes` instead. The current OTel "
+                "GenAI spec emits prompts/completions as span attributes "
+                "(`gen_ai.input.messages` / `gen_ai.output.messages`), which is "
+                "what `use_attributes=True` (the default) does. "
+                "`use_attributes=False` opts into the events path instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            use_attributes = use_legacy_attributes
         Config.exception_logger = exception_logger
         Config.enrich_token_usage = enrich_token_usage
         Config.get_common_metrics_attributes = get_common_metrics_attributes
         Config.upload_base64_image = upload_base64_image
-        Config.use_legacy_attributes = use_legacy_attributes
+        Config.use_legacy_attributes = use_attributes
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
--- a/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
+++ b/packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
@@ -6,8 +6,6 @@ import time
 import warnings
 from typing import Callable, Collection, Optional
 
-_USE_ATTRIBUTES_UNSET = object()
-
 from opentelemetry import context as context_api
 from opentelemetry._logs import Logger, get_logger
 from opentelemetry.instrumentation.anthropic.config import Config
@@ -799,15 +797,20 @@ class AnthropicInstrumentor(BaseInstrumentor):
         self,
         enrich_token_usage: bool = False,
         exception_logger=None,
-        use_attributes: bool = True,
+        use_attributes: Optional[bool] = None,
         get_common_metrics_attributes: Callable[[], dict] = lambda: {},
         upload_base64_image: Optional[
             Callable[[str, str, str, str], Coroutine[None, None, str]]
         ] = None,
-        use_legacy_attributes=_USE_ATTRIBUTES_UNSET,
+        use_legacy_attributes: Optional[bool] = None,
     ):
         super().__init__()
-        if use_legacy_attributes is not _USE_ATTRIBUTES_UNSET:
+        if use_attributes is not None and use_legacy_attributes is not None:
+            raise TypeError(
+                "Cannot pass both `use_attributes` and `use_legacy_attributes`; "
+                "`use_legacy_attributes` is deprecated, use `use_attributes` instead."
+            )
+        if use_legacy_attributes is not None:
             warnings.warn(
                 "`use_legacy_attributes` is deprecated and will be removed in a "
                 "future release; use `use_attributes` instead. The current OTel "
@@ -819,6 +822,8 @@ class AnthropicInstrumentor(BaseInstrumentor):
                 stacklevel=2,
             )
             use_attributes = use_legacy_attributes
+        if use_attributes is None:
+            use_attributes = True
         Config.exception_logger = exception_logger
         Config.enrich_token_usage = enrich_token_usage
         Config.get_common_metrics_attributes = get_common_metrics_attributes
--- a/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
+++ b/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
@@ -7,9 +7,7 @@ import time
 import traceback
 import warnings
 from functools import partial, wraps
-from typing import Collection
-
-_USE_ATTRIBUTES_UNSET = object()
+from typing import Collection, Optional
 
 from opentelemetry import context as context_api
 from opentelemetry._logs import get_logger
@@ -1146,11 +1144,16 @@ class BedrockInstrumentor(BaseInstrumentor):
         self,
         enrich_token_usage: bool = False,
         exception_logger=None,
-        use_attributes: bool = True,
-        use_legacy_attributes=_USE_ATTRIBUTES_UNSET,
+        use_attributes: Optional[bool] = None,
+        use_legacy_attributes: Optional[bool] = None,
     ):
         super().__init__()
-        if use_legacy_attributes is not _USE_ATTRIBUTES_UNSET:
+        if use_attributes is not None and use_legacy_attributes is not None:
+            raise TypeError(
+                "Cannot pass both `use_attributes` and `use_legacy_attributes`; "
+                "`use_legacy_attributes` is deprecated, use `use_attributes` instead."
+            )
+        if use_legacy_attributes is not None:
             warnings.warn(
                 "`use_legacy_attributes` is deprecated and will be removed in a "
                 "future release; use `use_attributes` instead. The current OTel "
@@ -1162,6 +1165,8 @@ class BedrockInstrumentor(BaseInstrumentor):
                 stacklevel=2,
             )
             use_attributes = use_legacy_attributes
+        if use_attributes is None:
+            use_attributes = True
         Config.enrich_token_usage = enrich_token_usage
         Config.exception_logger = exception_logger
         Config.use_legacy_attributes = use_attributes
--- a/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
+++ b/packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
@@ -5,8 +5,11 @@ import logging
 import os
 import time
 import traceback
+import warnings
 from functools import partial, wraps
 from typing import Collection
+
+_USE_ATTRIBUTES_UNSET = object()
 
 from opentelemetry import context as context_api
 from opentelemetry._logs import get_logger
@@ -1143,12 +1146,25 @@ class BedrockInstrumentor(BaseInstrumentor):
         self,
         enrich_token_usage: bool = False,
         exception_logger=None,
-        use_legacy_attributes: bool = True,
+        use_attributes: bool = True,
+        use_legacy_attributes=_USE_ATTRIBUTES_UNSET,
     ):
         super().__init__()
+        if use_legacy_attributes is not _USE_ATTRIBUTES_UNSET:
+            warnings.warn(
+                "`use_legacy_attributes` is deprecated and will be removed in a "
+                "future release; use `use_attributes` instead. The current OTel "
+                "GenAI spec emits prompts/completions as span attributes "
+                "(`gen_ai.input.messages` / `gen_ai.output.messages`), which is "
+                "what `use_attributes=True` (the default) does. "
+                "`use_attributes=False` opts into the events path instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            use_attributes = use_legacy_attributes
         Config.enrich_token_usage = enrich_token_usage
         Config.exception_logger = exception_logger
-        Config.use_legacy_attributes = use_legacy_attributes
+        Config.use_legacy_attributes = use_attributes
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
--- a/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
+++ b/packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
@@ -7,8 +7,6 @@ import types
 import warnings
 from typing import Collection, Optional, Union
 
-_USE_ATTRIBUTES_UNSET = object()
-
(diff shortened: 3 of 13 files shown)


doronkopit5:
The PR cleanly threads the flag through the SDK, and the mechanical scope is right (all 8 instrumentors that accept the kwarg are covered). Before merging though, I want to flag that the world shifted under issue #3236 in a way that affects what this flag should be named and what it should do.

### The OTel semconv decision reversed

When #3236 was filed, the assumption was: "legacy attributes (`gen_ai.prompt`/`gen_ai.completion`) → events" is the migration path. That's no longer true. [Semconv issue #2010](https://github.com/open-telemetry/semantic-conventions/issues/2010) was re-litigated and closed in favor of **attributes on spans, not events**. The new spec-compliant attributes are `gen_ai.input.messages` / `gen_ai.output.messages` / `gen_ai.system_instructions` — and events failed to gain adoption across the GenAI tracing ecosystem.

### OpenLLMetry has already migrated

Spot-checking the instrumentors, they all emit the new attributes today by default:

- `opentelemetry-instrumentation-openai/.../shared/chat_wrappers.py:552, 612` — `GEN_AI_INPUT_MESSAGES` / `GEN_AI_OUTPUT_MESSAGES`
- `opentelemetry-instrumentation-anthropic/.../span_utils.py:147, 318`
- `opentelemetry-instrumentation-groq/.../span_utils.py:140, 183`
- `opentelemetry-instrumentation-bedrock/.../span_utils.py:185, 376`
- `opentelemetry-instrumentation-langchain/.../span_utils.py:231, 379`

So `use_legacy_attributes=True` (the default) is already the **current spec-compliant** behavior.

### What `use_legacy_attributes=False` actually does

Looking at `opentelemetry-instrumentation-openai/.../utils.py:191-198` and the call sites in `chat_wrappers.py:273-287`:

```python
if should_emit_events():        # not Config.use_legacy_attributes AND event_logger set
    emit_event(...)             # emit prompts as OTel log events
elif should_send_prompts():
    await _set_prompts(...)     # emit gen_ai.input.messages on the span
```

So the real semantics today are:

| Flag value | What gets emitted | Status |
|---|---|---|
| `use_legacy_attributes=True` (default) | `gen_ai.input.messages` on span | Current spec |
| `use_legacy_attributes=False` (this PR exposes) | OTel log events instead | Abandoned direction |

The name is now actively misleading: a user reading the SDK signature would reasonably assume `False` means "non-legacy / modern", but it actually opts them **off** the current spec and **onto** the events path.

### Proposal: rename to `use_attributes`

This matches @nirga's original suggestion in #3236 (\"`use_attributes` (or `use_events`)\") and reflects reality without flipping defaults:

| Old | New |
|---|---|
| `use_legacy_attributes: bool = True` | `use_attributes: bool = True` |

Default stays `True` (no behavior change for existing users). The "legacy" word — which is the inaccurate part — goes away. `False` continues to mean "emit as events instead."

To keep the SDK and instrumentor-level APIs consistent, I'd suggest renaming at **both** layers in this PR:

1. Rename the kwarg on the 8 instrumentor `__init__`s to `use_attributes`, accepting `use_legacy_attributes` as a deprecated alias that emits a `DeprecationWarning` and forwards to the new name.
2. Same on `Traceloop.init()`.
3. Remove the alias in the next major version.

This adds maybe ~15 lines of deprecation-shim code but gives us a clean, accurate public API without breaking anyone.

### Two related items worth deciding before merge

1. **`EventLoggerProvider` plumbing.** The original issue also asked how to configure the event logger when opting into events. Today, `use_legacy_attributes=False` silently relies on the global logger provider being configured. Worth documenting clearly that users must configure one themselves — otherwise setting the flag silently does nothing useful.

2. **Test robustness.** The new test in `tests/test_sdk_initialization.py:232-253` manually mutates `TracerWrapper.instance` instead of using the existing fixture pattern from `conftest.py`. It also only verifies one instrumentor (OpenAI). Worth parametrizing across 2-3 instrumentors and using the fixture pattern to avoid global-state leakage if `Traceloop.init()` ever raises mid-call.

dvirski:
@doronkopit5  Your suggestions were on point.

What has changed:

1. Renamed kwarg use_legacy_attributes → use_attributes across 8 instrumentor __init__ methods (openai, anthropic, bedrock, sagemaker, groq, langchain, together, watsonx).

2. Added deprecation shim — each __init__ still accepts use_legacy_attributes via a sentinel, emits DeprecationWarning, and forwards to use_attributes.

3. Same rename + deprecation shim on Traceloop.init() at the SDK layer.

4. Threaded use_attributes through the SDK call chain — TracerWrapper.__new__, init_instrumentations, and all 8 init_*_instrumentor helpers in tracing.py.

5. Added docstring to Traceloop.init() documenting use_attributes semantics and the requirement that use_attributes=False needs an EventLoggerProvider configured.

6. Rewrote the test using a reusable isolated_tracer_wrapper fixture (replaces manual TracerWrapper.instance mutation) and expanded coverage to 3 tests: default-is-True across OpenAI + Anthropic, use_attributes=False propagates to OpenAI + Anthropic, deprecated alias still works and emits
DeprecationWarning.

Default behavior unchanged; existing callers using use_legacy_attributes keep working (with a warning)

doronkopit5:
Thanks @dvirski — the rename, the deprecation shim, the docstring, the fixture-based tests are all on point. Two concrete issues with the implementation though:

### 1. CI lint will fail — `_USE_ATTRIBUTES_UNSET = object()` placement triggers E402

In 8 of the 9 modified files the sentinel sits **between** stdlib imports and the rest, which ruff flags as `E402 Module level import not at top of file`. I ran ruff locally on the branch:

| File | Ruff errors |
|---|---|
| `opentelemetry-instrumentation-anthropic/__init__.py` | 19 |
| `opentelemetry-instrumentation-bedrock/__init__.py` | 19 |
| `opentelemetry-instrumentation-groq/__init__.py` | 17 |
| `opentelemetry-instrumentation-langchain/__init__.py` | 15 |
| `opentelemetry-instrumentation-sagemaker/__init__.py` | 14 |
| `opentelemetry-instrumentation-together/__init__.py` | 14 |
| `opentelemetry-instrumentation-watsonx/__init__.py` | 15 |
| `traceloop-sdk/sdk/__init__.py` + `tracing.py` | 41 combined |
| `opentelemetry-instrumentation-openai/__init__.py` | 0 (sentinel placed correctly below imports) |

### 2. The whole sentinel machinery is overengineering — `Optional[bool] = None` is the idiomatic fix

`None` isn't a valid value for a bool flag, so it's a natural sentinel. Drops 9 module-level objects, sidesteps the import-ordering trap entirely, and reads cleaner:

```python
def __init__(
    self,
    ...
    use_attributes: bool = True,
    use_legacy_attributes: Optional[bool] = None,
):
    if use_legacy_attributes is not None:
        warnings.warn(...)
        use_attributes = use_legacy_attributes
```

This also addresses a subtle behavior issue with the current shim: if a caller passes **both** kwargs — `Traceloop.init(use_attributes=False, use_legacy_attributes=True)` — the deprecated one silently clobbers the explicit new one. Worth deciding which should win (I'd vote: the new kwarg wins, or raise `TypeError`), and either way add a one-line test.

### Nit

The same ~12-line `warnings.warn(...)` block is copy-pasted 9 times. A small helper (e.g. `traceloop.sdk._deprecation.warn_use_legacy_attributes()`) would keep the message in one place if it ever needs to change.

Everything else looks good to merge once these are addressed.

packages/opentelemetry-instrumentation-watsonx/opentelemetry/instrumentation/watsonx/__init__.py
```diff
@@ -642,10 +645,27 @@ class WatsonxSpanAttributes:
...
+                "(`gen_ai.input.messages` / `gen_ai.output.messages`), which is "
+                "what `use_attributes=True` (the default) does. "
+                "`use_attributes=False` opts into the events path instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            use_attributes = use_legacy_attributes
         Config.exception_logger = exception_logger
-        Config.use_legacy_attributes = use_legacy_attributes
+        Config.use_legacy_attributes = use_attributes
```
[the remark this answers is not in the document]
dvirski: Valid finding but pre-existing — not introduced by this PR. _emit_response_events had the broken signature on main before this branch was cut, and the bug only fires when use_attributes=False (events path) + a 2-arg call site is hit; even then it's silently swallowed by the surrounding 
@dont_throw. This PR is scoped to the kwarg rename, so I'll leave the watsonx event-emission bug for a follow-up that can also fix the matching issue inside _emit_response_events itself (the emit_event(ChoiceEvent(...)) call also passes only one arg and would TypeError on its own).

dvirski:
@doronkopit5 

Fixed last comments:

summary of what was fixed according to your comment numbers:

1. CI lint failure — _USE_ATTRIBUTES_UNSET = object() triggering E402
Fixed. Removed all 9 module-level _USE_ATTRIBUTES_UNSET = object() sentinels by switching to the Optional[bool] = None pattern (see #2). ruff check now passes clean across all 9 modified files. While I was in tests/test_sdk_initialization.py I also cleaned up two pre-existing unused/redefined
imports (SimpleSpanProcessor, BatchSpanProcessor) that were also tripping ruff.

2. Sentinel machinery is overengineering — Optional[bool] = None is idiomatic
Fixed. Both kwargs are now Optional[bool] = None. Resolution logic in all 9 sites:
if use_attributes is not None and use_legacy_attributes is not None:
      raise TypeError(
          "Cannot pass both `use_attributes` and `use_legacy_attributes`; "
          "`use_legacy_attributes` is deprecated, use `use_attributes` instead."
      )
  if use_legacy_attributes is not None:
      warnings.warn(..., DeprecationWarning, stacklevel=2)
      use_attributes = use_legacy_attributes
  if use_attributes is None:
      use_attributes = True

On the both-kwargs question: went with raise TypeError rather than "new wins". Passing both is a programmer error, and a loud failure beats either kwarg silently winning. Added a test (test_passing_both_kwargs_raises_type_error) asserting pytest.raises(TypeError, match="Cannot pass both").


Nit — duplicated 12-line warning across 9 sites
Not extracted, intentionally. Two reasons:
1. Dependency direction. Today the dependency flow is one-way: traceloop-sdk depends on the instrumentor packages. Putting a shared helper in traceloop.sdk._deprecation would force the 8 instrumentor packages to depend on the SDK, which inverts the current dependency direction and would create a
cycle.
2. Package overhead for a new package only for this,
The duplication isn't free, but in this case the cure (new files or a circular dep) is worse than the disease. Once the alias is removed in the next major, all 9 copies go away together.

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
```diff
@@ -787,18 +788,38 @@ def __init__(
...
+                "future release; use `use_attributes` instead. The current OTel "
+                "GenAI spec emits prompts/completions as span attributes "
+                "(`gen_ai.input.messages` / `gen_ai.output.messages`), which is "
+                "what `use_attributes=True` (the default) does. "
+                "`use_attributes=False` opts into the events path instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            use_attributes = use_legacy_attributes
+        if use_attributes is None:
```
nina-kollman, line 825: This means the use_attributes should have a default value of True - it should not be optional

packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py
```diff
@@ -1113,12 +1114,32 @@ def __init__(
...
+                "future release; use `use_attributes` instead. The current OTel "
+                "GenAI spec emits prompts/completions as span attributes "
+                "(`gen_ai.input.messages` / `gen_ai.output.messages`), which is "
+                "what `use_attributes=True` (the default) does. "
+                "`use_attributes=False` opts into the events path instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            use_attributes = use_legacy_attributes
+        if use_attributes is None:
```
nina-kollman, line 1168: Same

packages/opentelemetry-instrumentation-anthropic/opentelemetry/instrumentation/anthropic/__init__.py
dvirski: ▎ Good catch on the signature — fair to question why use_attributes is typed Optional[bool] when conceptually it's a bool defaulting to True.
  ▎
  ▎ The None default isn't because None is a meaningful runtime value — it's a marker for "caller didn't pass this kwarg." I need that marker to detect when a caller passes both use_attributes and the deprecated use_legacy_attributes in the same call, and raise TypeError (per Doron's earlier ask — 
  ▎ silent clobber of an explicit kwarg is sneaky). The check is:
  ▎
  ▎ if use_attributes is not None and use_legacy_attributes is not None:
  ▎     raise TypeError("Cannot pass both ...")
  ▎
  ▎ If I default use_attributes to True, I lose the ability to tell apart __init__(use_legacy_attributes=False) from __init__(use_attributes=True, use_legacy_attributes=False) — both look the same inside the method body, and the both-passed TypeError becomes impossible to detect cleanly.
  ▎
  ▎ The alternatives I considered:
  ▎ 1. Module-level _UNSET = object() sentinel — what I originally had; Doron flagged it as overengineering (and it triggered E402 in 8 of 9 files).
  ▎ 2. **kwargs to absorb the deprecated alias — kills IDE autocomplete + discoverability of use_legacy_attributes.
  ▎ 3. Lie in the annotation (bool = True but actually default to a sentinel) — breaks inspect.signature and mypy.
  ▎
  ▎ So Optional[bool] = None ended up being the least-bad option: it's a known Python idiom for "unset", costs no extra symbols, and matches what Optional semantically means here ("either a bool, or the absence of one"). Once we remove the use_legacy_attributes alias in the next major, this can 
  ▎ collapse to use_attributes: bool = True.
  ▎
  ▎ Happy to add an inline code comment explaining this if you think it'd help future readers.

nina-kollman: I get the argument. But still I think you are aming to the final situation after the deprecation that param looks like:
        use_attributes: bool = True

If this is the case then the logic should be to check if the legacy is defined, warn about it and deal with it.. 

Your choice

nina-kollman (approved):

PR #4135, opened 2026-05-13
feat(bedrock): add async support for aioboto3

### LINKED ISSUE #3510: 🐛 Bug Report: Add support to aioboto3 + bedrock converse

Bedrock Instrumentation

### 📜 Description

we are not getting the out of the box span attributes (gen_ai.usage.output_tokens, gen_ai.usage.input_token etc..), not the bedrock.converse span when working with bedrock converse and aiboto3

### 👟 Reproduction steps

 aioboto3_session = aioboto3.Session(
            aws_access_key_id=xx,
            aws_secret_access_key=xx,
            aws_session_token=xx,
            region_name=REGION
        )
        
async with aioboto3_session.client('bedrock-runtime') as bedrock_client:

         response = await bedrock_client.converse(
                modelId=model,
                system=system,
                messages=messages)

### 👍 Expected behavior

get the bedrock.converse span as well as all out of the box span attributes

- no bedrock.converse span
- no out of the box attributes llm span attributes

### 🤖 Python Version

3.12

_No response_

None

dvirski:
Screenshot 2026-05-13 at 15 44 38

 
 Fixes #3510

Bedrock’s OpenTelemetry instrumentation worked for boto3 (sync) but produced zero spans when users called Bedrock through aioboto3. This PR adds full async support — all four Bedrock Runtime methods (converse, converse_stream, invoke_model, invoke_model_with_response_stream)
now emit OTel-semconv-compliant spans, prompts, completions, token counts, and finish reasons.

The fix is symmetric to the existing sync path: same downstream helpers, same span attributes, same event-emitter pipeline, same semconv compliance — async just adds the missing async hook + four async-aware wrappers.

Tests: Added tests/traces/test_aioboto3.py — 12 mock-based tests covering all 4 async methods × 3 attribute modes (legacy / events+content / events+no-content), matching the sync test_nova.py coverage. Uses mocks instead of VCR cassettes because vcrpy's aiohttp stub can't replay aiobotocore responses (vcrpy#927 (https://github.com/kevin1024/vcrpy/issues/927)).

Example: Added packages/sample-app/sample_app/async_bedrock_example.py — a research-assistant @workflow chaining all 4 async methods (converse → converse_stream → invoke_model → invoke_model_with_response_stream), so the resulting Traceloop trace shows a real-life async Bedrock flow.

Stress test (local-only, not committed): Wrote async_bedrock_stress_paths.py exercising 4 edge-case paths the doc example doesn't cover — invalid model (exception recording), concurrent asyncio.gather (no state leakage between simultaneous calls), early break from streaming
(span still ends with partial content, no ERROR status per OTel guidance), and SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY (instrumentation correctly skipped). This caught a real bug in converse_stream where the span would leak on early break, which we then fixed.

Code review (adversarial subagent): Spawned a code-reviewer subagent with explicit instructions to find bugs in the new async code. It flagged ~17 concerns; we fixed the ones introduced by this PR and deferred the ones that pre-exist in the sync path. Fixed:
AsyncStreamingWrapper span leak on early break (try/finally + _done guard), BufferedAsyncBody buffer-mutation bug (immutable buffer + cursor), aiobotocore body never closed (now released before replacement), InstrumentedClientContext.__aenter__ not exception-safe (wrapped
patching block so the inner ctx is properly exited on failure), _handle_async_call passing a fake response dict to emit_choice_events (now passes the real response), and _handle_async_call swallowing exceptions at debug log level (now error).

## Summary by CodeRabbit

* **New Features**
  * Async Bedrock instrumentation for aioboto3/aiobotocore: tracing for invoke/converse (streaming and non-streaming), buffered async responses, and robust async stream handling to ensure metrics/events and connection cleanup.
  * Added an async sample workflow demonstrating classification, streaming answer generation, summarization, and follow-ups.

* **Bug Fixes**
  * Improved extraction of token usage from nested response fields and headers.

* **Tests**
  * Comprehensive async tests for legacy/event-stream modes, edge cases, and resource handling.

* **Chores**
  * Test config updated to support asyncio and async AWS test packages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4135?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


doronkopit5:
### 1. Async exception logging is asymmetric with sync (`__init__.py:683-689`)

```python
except Exception as e:
    logger.error(
        "OpenLLMetry failed to trace in _handle_async_call, error: %s", e, exc_info=True
    )
```

The sync path uses `@dont_throw` (`utils.py:11-33`), which logs at `logger.debug`. Promoting async-only to `logger.error` with `exc_info=True` means async users will see stack traces in their logs that sync users don't — an instrumentation bug becomes user-visible noise. Either bump both, or align async to `logger.debug` for parity.

### 2. Only 1 of 12 tests asserts token attributes (`tests/traces/test_aioboto3.py`)

`test_aioboto3_converse_legacy` (lines 232-233) is the only test asserting `GEN_AI_USAGE_INPUT_TOKENS` / `GEN_AI_USAGE_OUTPUT_TOKENS`. The other 11 (including `test_aioboto3_invoke_model_legacy`, `test_aioboto3_invoke_stream_legacy`, `test_aioboto3_converse_stream_legacy`) only assert span name + `PROVIDER_NAME` / `OPERATION_NAME` / `REQUEST_MODEL`. Issue #3510 explicitly calls out token attributes as missing — a regression in token extraction for invoke_model or streaming would not be caught. Add token-attribute assertions to all 12 tests (mock values are deterministic: `input=11`, `output=16`).

### 3. No committed tests for the claimed edge-case fixes

The PR description enumerates fixes for:
- Early `break` out of `async for` (AsyncStreamingWrapper + converse_stream span leak)
- Exception during iteration
- Concurrent `asyncio.gather` (no state leakage)
- `SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY` short-circuit

None of these are in `test_aioboto3.py` — the PR description notes they were tested in a local-only stress harness. Those are the very bugs the author identified and fixed; without committed tests they can silently regress. At minimum:

- `test_aioboto3_converse_stream_early_break` — break after first chunk, assert `len(spans) == 1` with partial content.
- `test_aioboto3_invoke_stream_exception` — raise from inside the `async for` body, assert the span still ends.
- `test_aioboto3_suppress_language_model_instrumentation` — set the context key, assert `len(spans) == 0`.

### 7. Mocks bypass the real aiobotocore body lifecycle (`test_aioboto3.py:87-98`)

`_AsyncReadableBody` does not implement `close()`. The instrumentation (`__init__.py:634-641`) tries `getattr(original_body, \"close\", None)` — with the mock it returns `None` and the close-branch is silently skipped, so the test never exercises the close path. The mock also doesn't simulate aiobotocore's \"body is consumed after first read\" semantic, so the `BufferedAsyncBody` swap-in isn't truly proven necessary by these tests.

Add:
- `close()` (and `aclose()` if applicable) on `_AsyncReadableBody` that records invocation, and assert it gets called.
- A test that calls `await response[\"body\"].read()` twice and asserts the second call returns `b\"\"` — proves the buffered-body cursor semantic.

dvirski:
@doronkopit5 

Topics addressed:

  #1 — Async exception logging asymmetric with sync
  - bedrock/__init__.py:683-690 — replaced logger.error(..., exc_info=True) with logger.debug(..., traceback.format_exc()) to match @dont_throw.
  - Added import traceback.

  #2 — Only 1 of 12 tests asserts token attributes
  - Added deterministic GEN_AI_USAGE_INPUT_TOKENS == 11 / GEN_AI_USAGE_OUTPUT_TOKENS == 16 assertions to all 12 existing async tests in test_aioboto3.py.
  - Tightened the existing > 0 assertion in test_aioboto3_converse_legacy to == 11 / == 16.
  
  #3 — No committed tests for the claimed edge-case fixes
  - test_aioboto3_invoke_stream_early_break — drives iterator with __anext__() + aclose(), asserts span ends. Locks the try/finally + _done guard in AsyncStreamingWrapper.
  - test_aioboto3_invoke_stream_exception_during_iteration — uses stream.athrow(RuntimeError("boom")) to inject mid-iteration exception, asserts span still ends.
  - test_aioboto3_suppress_language_model_instrumentation — sets SUPPRESS_LANGUAGE_MODEL_INSTRUMENTATION_KEY, asserts len(spans) == 0.

  #7 — Mocks bypass real aiobotocore body lifecycle
  - _AsyncReadableBody — added async def close() with close_calls counter (exercises the await-if-coroutine branch in the instrumentation).
  - test_aioboto3_invoke_model_closes_underlying_body — captures original body via factory wrapper, asserts close_calls == 1.
  - test_aioboto3_invoke_model_body_is_rereadable — reads response["body"] twice, asserts both calls return identical bytes (proves BufferedAsyncBody swap-in is necessary).

  ---
  Bonus — preexisting sync bug found while addressing #2
  - bedrock/span_utils.py:_set_amazon_span_attributes — fixed Nova token extraction:
    - Non-streaming path was reading top-level inputTokens (always missing → 0 in prod). Now reads usage.inputTokens (where Nova actually puts it).
    - Streaming-accumulated path now also reads metadata.usage.{input,output}Tokens.
    - HTTP header (x-amzn-bedrock-input-token-count) used as fallback for both.
  - Verified live against AWS Bedrock Nova-lite via probe script — confirmed response shape matches the cassette and the fix.
  
  Bonus — sync regression locks for the above fix
  - test_nova.py::test_nova_completion — added GEN_AI_USAGE_INPUT_TOKENS == 30 / OUTPUT_TOKENS == 51 (values from VCR cassette).
  - test_nova.py::test_nova_invoke_stream — added > 0 token assertions to guard the streaming-branch fix.
  
  Verification: 380/380 bedrock package tests pass, ruff clean.

doronkopit5 (approved):

PR #4137, opened 2026-05-13
 fix(sdk): warn when both exporter and processor are passed to Traceloop.init()

### LINKED ISSUE #3046: 🐛 Bug Report: `exporter` is ignored when `processor` is also defined in `Traceloop.init()`

Traceloop SDK

### 📜 Description

When both a custom `SpanProcessor` and `SpanExporter` are passed to `Traceloop.init()`, the exporter is silently ignored. This leads to unexpected behavior: the processor is invoked, but the exporter is never called.

### 👟 Reproduction steps

1. Create a fresh virtual environment.
2. Install the SDK
```
pip install traceloop-sdk==0.40.14
```
3. Run the following code:

```python
from typing import Sequence
from opentelemetry.sdk.trace import ReadableSpan
from opentelemetry.sdk.trace.export import SpanExporter, SpanProcessor, SpanExportResult
from opentelemetry.trace import Span


class MySpanProcessor(SpanProcessor):
    def __init__(self):
        self.call_count = 0

    def on_start(self, span: Span, parent_context):
        self.call_count += 1
        print("MySpanProcessor")


class MySpanExporter(SpanExporter):
    def __init__(self):
        self.call_count = 0

    def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
        self.call_count += 1
        print("MySpanExporter")
        return SpanExportResult.SUCCESS


@workflow()
def run_workflow():
    pass


my_proc = MySpanProcessor()
my_export = MySpanExporter()
Traceloop.init(processor=my_proc, exporter=my_export)

run_workflow()

assert my_proc.call_count > 0, "Expected MySpanProcessor to be called"
assert my_export.call_count > 0, "Expected MySpanExporter to be called"
```

4. Observe that the second assert fails:
```
AssertionError: Expected MySpanExporter to be called
```

### 👍 Expected behavior

- Either support combining a custom processor and exporter by internally wiring them,

    or

- Raise an error or warning when both are provided to avoid confusion.

See Reproduction steps

### 🤖 Python Version

_No response_

_No response_

None

dvirski:
Fixes #3046.
Problem: passing both exporter and processor is a mistake — the processor already wraps the exporter
internally. The exporter was silently ignored with no feedback to the user.

Fix: emit a UserWarning pointing at the caller's line. Also removes the redundant
exporter= from the exporter_with_custom_span_processor test fixture which had the same bug.

## Summary by CodeRabbit

* **New Features**
  * Traceloop.init now emits a UserWarning when both an exporter and a processor are supplied, clarifying the exporter will be ignored and advising wrapping the exporter with a processor instead.

* **Tests**
  * Added a test confirming the warning is raised and that span export behavior follows the processor-wrapped exporter.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4137?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/traceloop-sdk/tests/test_sdk_initialization.py
+++ b/packages/traceloop-sdk/tests/test_sdk_initialization.py
@@ -231,19 +231,39 @@ def test_get_default_span_processor():
 
 def test_both_exporter_and_processor_warns():
     """Passing both exporter and processor is a mistake — the processor already wraps
-    the exporter internally. We warn instead of silently dropping the exporter."""
-    if hasattr(TracerWrapper, "instance"):
-        _instance = TracerWrapper.instance
+    the exporter internally. We warn instead of silently dropping the exporter, and
+    verify the standalone exporter receives nothing while the processor's wrapped
+    exporter receives the span."""
+    saved_instance = getattr(TracerWrapper, "instance", None)
+    if saved_instance is not None:
         del TracerWrapper.instance
 
-    exporter = InMemorySpanExporter()
-    processor = SimpleSpanProcessor(exporter)
+    try:
+        standalone_exporter = InMemorySpanExporter()
+        processor_exporter = InMemorySpanExporter()
+        processor = SimpleSpanProcessor(processor_exporter)
 
-    with pytest.warns(UserWarning, match="exporter.*ignored"):
-        Traceloop.init(
-            exporter=exporter,
-            processor=processor,
+        with pytest.warns(UserWarning, match="exporter.*ignored"):
+            Traceloop.init(
+                exporter=standalone_exporter,
+                processor=processor,
+            )
+
+        @workflow(name="probe_workflow")
+        def probe():
+            return "ok"
+
+        probe()
+
+        assert len(standalone_exporter.get_finished_spans()) == 0, (
+            "The standalone exporter must receive no spans — the warning promises it is ignored."
         )
-
-    if "_instance" in dir():
-        TracerWrapper.instance = _instance
+        processor_spans = processor_exporter.get_finished_spans()
+        assert any(span.name == "probe_workflow.workflow" for span in processor_spans), (
+            "The processor's wrapped exporter must receive the emitted span."
+        )
+    finally:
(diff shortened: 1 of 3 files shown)


packages/traceloop-sdk/traceloop/sdk/__init__.py
```diff
@@ -90,6 +91,14 @@ def init(
 
         enable_content_tracing = is_content_tracing_enabled()
 
+        if exporter and processor:
+            warnings.warn(
+                "Both 'exporter' and 'processor' were provided to Traceloop.init(). "
+                "The 'exporter' will be ignored — your processor should already wrap the exporter.",
```
doronkopit5:
```suggestion
                "Both 'exporter' and 'processor' were provided to Traceloop.init(). "
                "The 'exporter' will be ignored — your processor should already wrap the exporter. "
                "Wrap your exporter inside the processor, e.g.: SimpleSpanProcessor(my_exporter).",
```

packages/traceloop-sdk/tests/test_sdk_initialization.py
```diff
@@ -223,3 +227,23 @@ def test_get_default_span_processor():
     assert isinstance(processor, BatchSpanProcessor)
     assert hasattr(processor, "_traceloop_processor")
     assert getattr(processor, "_traceloop_processor") is True
+
+
+def test_both_exporter_and_processor_warns():
```
doronkopit5, line 232: the test asserts the warning fires but doesn’t assert what actually happens (processor wins, exporter is never called). If someone later changes TracerWrapper.__new__ to wire both together (option (a) from the issue), this test won’t catch that the warning is now lying. Consider adding one more assertion: emit a span, confirm the standalone exporter received zero spans while the processor’s wrapped exporter received them.

dvirski:
@doronkopit5 

Fixed both your comments (in resolved state now)

nina-kollman (approved):

Issue #4150, opened 2026-05-14
Best-practice: anonymize secrets in VCR test cassettes (+ minor SDK jinja2 trust-boundary comment)


elfrost:
Hi Traceloop team,

While testing [AI PatchLab](https://github.com/elfrost/ai-patchlab) (an open-source local-first security scanner) on a few mid-popularity Python AI projects, I scanned openllmetry at approximately `72fc45e` and wanted to flag one best-practice improvement plus one minor SDK code-clarity note. Filing as a single courtesy issue.

Full curated write-up of the scan (with FP analysis, methodology, and the findings AI PatchLab got wrong): https://elfrost.github.io/ai-patchlab/scans/traceloop-openllmetry.html

## 1. Anonymize secrets in VCR cassettes before recording

Of 26 high-severity findings on the scan, 25 are Gitleaks matches in `packages/**/tests/cassettes/**.yaml`:

- 11× `aws-access-token` matches in `opentelemetry-instrumentation-anthropic/tests/cassettes/test_bedrock_*/`
- 8× `jwt` matches in `opentelemetry-instrumentation-watsonx/tests/`
- 6× `generic-api-key` matches (including PostHog `phc_…` public keys in haystack cassettes)

**None of these are credential leaks today**: the AWS findings are access key IDs without their corresponding secret keys (the Sigv4 signature in the cassette is only valid for that one already-replayed request), the JWTs have transparently placeholder claims (`sub: noone@ibm.com`, `account.bss: abc123`), and the PostHog `phc_` keys are public write-only event-ingestion identifiers by design.

But this is still worth addressing because:
- **Cassettes leak metadata**: which AWS account, which Bedrock model, which day, which API surface. For an observability SDK that ships to enterprises, that's worth scrubbing.
- **One bad re-record away from a real secret**: if VCR isn't configured to anonymize, the next contributor recording a cassette with a real prod key against a different provider will accidentally land it. Unfiltered cassettes are a recurring source of real-world key leaks across Python OSS.

Recommended fix: configure VCR's `filter_headers`, `filter_query_parameters`, and `before_record_response` in the test base (probably in each package's `conftest.py` or a shared `tests/common/`):

```python
import vcr

vcr_config = vcr.VCR(
    filter_headers=[
        ('authorization', 'REDACTED'),
        ('x-api-key', 'REDACTED'),
    ],
    filter_query_parameters=[
        ('api_key', 'REDACTED'),
    ],
[...]

PR #4151, opened 2026-05-16
fix(traceloop-sdk): document jinja2 autoescape rationale (Refs #4150 part 2)

dparikh79:
Refs #4150 (the second of the two items the reporter raised).

The prompt registry's `Environment()` in `packages/traceloop-sdk/traceloop/sdk/prompts/client.py:44` is correct as written. Static analyzers (Semgrep `direct-use-of-jinja2`, Bandit `S701`) flag the call because Jinja2's default `autoescape=False` is unsafe when the rendered output is HTML. Here the rendered output goes to an LLM as a prompt, where enabling autoescape would actively damage the result by escaping `<`, `>`, and `&` that callers may want to send to the model verbatim.

This PR adds an inline rationale comment and a scoped `# noqa: S701` so the intent is documented for future contributors and so scanners stop re-flagging this on every run. Runtime behavior is unchanged.

Scope deliberately excludes the first item the reporter raised (VCR cassette anonymization). That one touches every package's `conftest.py` and warrants its own PR with maintainer input on whether the right place is per-package `conftest.py` or a shared `tests/common/`. Happy to follow up with a separate PR for that piece if you'd like the same approach.

## Validation

- File still parses: `python -c "import ast; ast.parse(open('packages/traceloop-sdk/traceloop/sdk/prompts/client.py').read())"` clean.
- The added `# noqa: S701` is the standard Bandit silencer for `B701: jinja2_autoescape_false`; the comment block above it explains the why.

## AI Assistance Disclosure

This PR was drafted with AI assistance (Claude Code). I read the reporter's scan write-up, traced the Environment() use site at `packages/traceloop-sdk/traceloop/sdk/prompts/client.py:44`, confirmed the intent (LLM prompt rendering, not HTML), and added the comment and Bandit suppression in line with the reporter's suggestion. Two-file repo audit would be needed to roll out the VCR change in part 1 of the issue, which is why that piece is intentionally left for a separate PR.

## Summary by CodeRabbit

* **Chores**
  * Improved inline documentation clarifying template rendering and autoescaping behavior for prompt handling. This update tightens guidance for maintainers, reduces potential confusion during future edits, and addresses code-quality checks. Small, low-risk documentation change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4151?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/prompts/client.py
+++ b/packages/traceloop-sdk/traceloop/sdk/prompts/client.py
@@ -41,6 +41,10 @@ class PromptRegistryClient:
         if not hasattr(cls, "instance"):
             obj = cls.instance = super(PromptRegistryClient, cls).__new__(cls)
             obj._registry = PromptRegistry()
+            # autoescape disabled intentionally: rendered output goes to an
+            # LLM as a prompt, not to HTML. Enabling autoescape here would
+            # escape characters like '<', '>', and '&' that callers may have
+            # meant to send to the model verbatim.
             obj._jinja_env = Environment()
 
         return cls.instance


dparikh79:
Thanks for the careful review. Addressed both points.

1. Switched the inline comment from RST-style double backticks to single quotes, matching conventional Python inline-comment formatting.
2. Dropped the `# noqa: S701` suppression. Confirmed `pyproject.toml`'s `[tool.ruff.lint] select` only enables `E`, `F`, `W`, so the suppression has no current effect; the rationale comment alone is enough until/unless S-rules are enabled in the future.

Issue #4157, opened 2026-05-19
🐛 Bug Report: GenerationSpanData missing response field causes _extract_response_attributes to never execute


heyanlong:
### Which component is this bug for?

OpenAI Instrumentation

### 📜 Description

GenerationSpanData missing response field causes _extract_response_attributes to never execute

```python
@dont_throw
    def on_span_end(self, span):
        """Called when a span ends - finish OpenTelemetry span."""
        from agents import GenerationSpanData

        if not span or not hasattr(span, "span_data"):
            return

        if span in self._otel_spans:
            otel_span = self._otel_spans[span]
            span_data = getattr(span, "span_data", None)
            trace_content = should_send_prompts()
            if span_data and (
                type(span_data).__name__ == "ResponseSpanData"
                or isinstance(span_data, GenerationSpanData)
            ):
                # Extract prompt data from input
                input_data = getattr(span_data, "input", [])
                _extract_prompt_attributes(otel_span, input_data, trace_content)

                # Add function/tool specifications to the request using OpenAI semantic conventions
                response = getattr(span_data, "response", None) # if span_data is GenerationSpanData, so _extract_response_attributes to never execute
                if response:
                    model_settings = _extract_response_attributes(otel_span, response, trace_content)
                    self._last_model_settings = model_settings
```

### 👟 Reproduction steps

```
from opentelemetry.instrumentation.openai_agents import OpenAIAgentsInstrumentor

trace.set_tracer_provider(trace_provider)

OpenAIAgentsInstrumentor(
    replace_existing_processors=True
).instrument(
    agent_name="TestAgent",
    tracer_provider=trace_provider,
)

agent = Agent(
            name=name,
            instructions=f"{RECOMMENDED_PROMPT_PREFIX}\n\n{mark_instructions}",
            model=LitellmModel(model=model),
            model_settings=ModelSettings(
                include_usage=True
            ))
Runner.run_streamed(agent, messages, context=context, **kwargs)

```

### 👍 Expected behavior

There should be gen_ai.output.xxx

### 👎 Actual Behavior with Screenshots

<img width="243" height="509" alt="Image" src="https://github.com/user-attachments/assets/46d5480b-17a8-4d85-ab14-9a8217b94379" />

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

RudraDudhat2509:
[...]

PR #4184, opened 2026-05-23
fix(openai-agents): emit gen_ai.* attributes for GenerationSpanData (issue #4157)

devteamaegis:
## Problem

`_end_generation_span` in `_hooks.py` always looked for `span_data.response` to populate OTel output/usage attributes:

```python
response = getattr(span_data, "response", None)   # line 960
...
if response:                                        # line 976
    _extract_response_attributes(otel_span, response, trace_content)
```

`GenerationSpanData` (the span type produced by LiteLLM and any other provider that does not go through the OpenAI Responses API) has **no `response` attribute**. Its fields are `input`, `output`, `model`, `model_config`, and `usage` — directly on the object. So `response` was always `None`, the `if response:` guard never fired, and **every `gen_ai.response.model`, `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, and `gen_ai.output.messages` attribute was silently dropped** for those spans.

Reported in issue #4157.

## Fix

Add a new `_extract_generation_span_data_attributes` helper that handles the `GenerationSpanData` shape: reads `model`, `model_config` (temperature, top_p, max_tokens), `output` (list of message dicts), and `usage` (dict of token counts) directly from `span_data`, then maps them to the same OTel attributes that `_extract_response_attributes` sets for `ResponseSpanData`.

`_end_generation_span` falls through to the new helper when `span_data` has no `response` attribute but does expose `output` or `usage` (i.e. when it is a `GenerationSpanData`).

## Tests

Three new tests in `TestEndGenerationSpan`:

| Test | Verifies |
|---|---|
| `test_generation_span_data_usage_attributes` | model, usage tokens, temperature, max_tokens are all set |
| `test_generation_span_data_output_messages` | output messages serialised when `trace_content=True` |
| `test_generation_span_data_output_suppressed_when_content_gated` | messages absent but usage still set when `trace_content=False` |

All 13 `TestEndGenerationSpan` tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced compatibility with non-OpenAI providers by ensuring model settings, token usage, and output messages are properly captured in telemetry traces.

* **Tests**
  * Added test coverage for generation span data handling with configurable message serialization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4184?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
+++ b/packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
@@ -383,6 +383,69 @@ def _extract_prompt_attributes(otel_span, input_data, trace_content: bool):
         )
 
 
+def _extract_generation_span_data_attributes(otel_span, span_data, trace_content: bool):
+    """Extract OTel gen_ai.* attributes from a GenerationSpanData object.
+
+    GenerationSpanData (used by LiteLLM and other providers that do not go through
+    the OpenAI Responses API) exposes ``model``, ``model_config``, ``output``, and
+    ``usage`` *directly* on the span_data object rather than via a nested
+    ``response`` attribute.  This function maps those fields to the same OTel
+    attributes that :func:`_extract_response_attributes` populates for
+    ResponseSpanData, fixing the silent data loss reported in issue #4157.
+    """
+    # Response model name
+    model = getattr(span_data, "model", None)
+    if model:
+        otel_span.set_attribute(GenAIAttributes.GEN_AI_RESPONSE_MODEL, model)
+
+    # Model config: temperature, top_p, max_tokens (stored as a plain dict)
+    model_config = getattr(span_data, "model_config", None) or {}
+    if isinstance(model_config, dict):
+        temperature = model_config.get("temperature")
+        if temperature is not None:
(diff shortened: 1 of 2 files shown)


packages/opentelemetry-instrumentation-openai-agents/opentelemetry/instrumentation/openai_agents/_hooks.py
```diff
@@ -383,6 +383,69 @@ def _extract_prompt_attributes(otel_span, input_data, trace_content: bool):
...
+    if output and trace_content:
+        output_messages = []
+        for item in output:
+            if isinstance(item, dict):
+                role = item.get("role", "assistant")
+                content = item.get("content", "")
+            else:
+                role = getattr(item, "role", "assistant")
+                content = getattr(item, "content", "")
+            if content is not None:
```
RichardoMrMu, line 425: [P1] Preserve tool-only GenerationSpanData outputs

OpenAI Agents writes GenerationSpanData.output as [message.model_dump()], and the supported OpenAI message model allows a tool-call response to have content=None while carrying the call in tool_calls. This loop only reads role/content and then skips the item when content is None, so a tool-only agent turn emits no gen_ai.output.messages at all—call ID, name, and arguments all disappear from the trace. The existing _convert_chat_message() above already maps tool_calls into OTel tool_call parts. Please use that conversion for GenerationSpanData output and add a regression with content=None plus tool_calls.

PR #4189, opened 2026-05-25
feat(mcp): respect TRACELOOP_TRACE_CONTENT=false in standard MCP instrumentor

### LINKED ISSUE #4188: feat: add support for TRACELOOP_TRACE_CONTENT=false in mcp instrumentor
# Feature description:
https://github.com/traceloop/openllmetry/blob/v0.60.0/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py does not respect `TRACELOOP_TRACE_CONTENT=false` (only the fast_mcp version does)

# Why is this feature needed?:
to make flows that can't store prompts/answers be able to use the sdk

## Note:
doing blank issue as can't see `mcp instrumentor` in list of components when creating issue

milkisbad:
The standard MCP instrumentor was unconditionally recording span input/output attributes, unlike the FastMCP instrumentor which already honored the TRACELOOP_TRACE_CONTENT env var. Guards all content attributes (TRACELOOP_ENTITY_INPUT, TRACELOOP_ENTITY_OUTPUT, MCP_RESPONSE_VALUE) behind the same _should_send_prompts() check, enabling workflows that cannot store prompts/responses to opt out. Closes #4188.

# screens

<details>
- disabled confidential information error pass:  
image
image
- diabled confidential information happy path:  
image
image
- enabled confidential information error pass:  
image
image
- enabled confidential information happy path:  
image
image

## Summary by CodeRabbit

* **New Features**
  * Centralized gating for prompt/content tracing: when enabled, tool and MCP spans include input/output and response value details (using cleaned output when configured).
  * When disabled, sensitive content (including tool error text and response/error descriptions) is suppressed, while operational metadata and span identity still appear.
* **Tests**
  * Added span-inspection coverage for `TRACELOOP_TRACE_CONTENT` across tool spans, MCP method spans, and error paths (ensuring no redacted sentinel text leaks).
* **Documentation**
  * Updated README with a “what is and isn’t gated” breakdown of always-on versus content-gated span fields.


YuxiangJiangCT:
Thanks for taking this — the gating on `TRACELOOP_ENTITY_INPUT` / `TRACELOOP_ENTITY_OUTPUT` / `MCP_RESPONSE_VALUE` reads cleanly.

One thing that might be in scope given the motivation ("workflows that cannot store prompts/responses"): two error-path `set_status` calls still write response content text into the span status, and they sit outside the `should_send_prompts()` guard:

- `_execute_and_handle_result`: `Status(StatusCode.ERROR, f"{result.content[0].text}")`
- `InstrumentedStreamWriter.send` (the `ResponseStreamWriter` span): `Status(StatusCode.ERROR, f"{request.result['content'][0]['text']}")`

So with `TRACELOOP_TRACE_CONTENT=false`, an error response's text still ends up in the span status description. Might be worth gating those too, or setting a generic message when content is disabled. The `Status(..., str(e))` ones look fine — those are exception messages, not response content.

Since `_execute_and_handle_result` is shared by both the tool and mcp-method paths, gating it there covers both. Happy to open a small follow-up if you'd rather keep this PR focused.

milkisbad:
Changed. It was a bit more complicated than what you wrote. I've identified 2 cases 
- 1st is tool errors via [Tool Execution Errors](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling) - my understanding is that it is not guarded against confidential information
- 2nd is there is error on Protocol Errors. i.e. in fastmcp providing wrong argument for tool results in pydantic model the error would look like (i've put the confidential info in **input_value='ber'**:   

```py
1 validation error for Model
foo
  value is not "bar", got "ber" [type=not_a_bar, **input_value='ber'**, input_type=str]
```

decided to fix both, but it required touching a lot of places.

milkisbad:
providing updated screenshots:

<details>
- disabled confidential information error pass:  
image
image
- diabled confidential information happy path:  
image
image
- enabled confidential information error pass:  
image
image
- enabled confidential information happy path:  
image
image

milkisbad:
@YuxiangJiangCT making sure you got the notification

YuxiangJiangCT:
@milkisbad nice, this is more thorough than what I flagged. Both isError branches are behind `should_send_prompts()` now, and good call extending it to the exception path. I'd written off `str(e)` as safe, but your pydantic `input_value=...` example is exactly why it isn't.

One asymmetry I noticed: `_handle_tool_call` and `_handle_mcp_method` both set `record_exception=False` / `set_status_on_exception=False`, but the `ResponseStreamWriter` span doesn't. So if the wrapped `send` raises in there, the SDK would still auto-record the exception and set status from `str(exc)` with no `should_send_prompts()` gate. Intentional (that path can't surface content-bearing exceptions), or worth the same treatment?

milkisbad:
> One asymmetry I noticed: `_handle_tool_call` and `_handle_mcp_method` both set `record_exception=False` / `set_status_on_exception=False`, but the `ResponseStreamWriter` span doesn't. So if the wrapped `send` raises in there, the SDK would still auto-record the exception and set status from `str(exc)` with no `should_send_prompts()` gate. Intentional (that path can't surface content-bearing exceptions), or worth the same treatment?

@YuxiangJiangCT Hmm could you help me out with this one - not too confident on the stream paths. Any concrete path you have in mind?  
We do handle expected errors correctly. So the only tracebacks would be like stream serialization specific? can't think of probable scenario where that would leak data and I do not want to mask too much right now

YuxiangJiangCT:
Yeah, I can't give you a concrete, probable one, so I'd leave it as is. With content off, both lines in `send` that touch the payload (`serialize(request.result)` and the `content[0]['text']` status) are already short-circuited by `should_send_prompts()`, so the only thing still handling the response inside that span is the wrapped `send` itself, i.e. your serialization case. I can't construct a realistic traceback there that carries the prompt/response text, and anything else that escapes would be structural or transport-level with a generic message rather than payload. Agreed it's not worth masking real errors for a hypothetical; if a concrete one ever shows up it can be its own follow-up. The tool/method gating is where it matters, and that looks solid.

milkisbad:
@YuxiangJiangCT oke how do follow from now on?

YuxiangJiangCT:
Nothing more from me, I'm good with it as-is. The next step is really out of my hands though: it needs a maintainer with write access to approve and kick off CI, and mine doesn't count for that. Tagging one on the PR is probably the fastest nudge. Glad I could help on this one.

Necmttn:
This is the right privacy split for MCP instrumentation: always trace metadata, gate content. I would make the metadata/content boundary explicit in docs and tests: method/tool name, server id, duration, status, and error class are always safe; prompt, response value, and serialized payload require TRACELOOP_TRACE_CONTENT=true. Local raw retention can stay a separate opt-in.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

milkisbad:
> This is the right privacy split for MCP instrumentation: always trace metadata, gate content. I would make the metadata/content boundary explicit in docs and tests: method/tool name, server id, duration, status, and error class are always safe; prompt, response value, and serialized payload require TRACELOOP_TRACE_CONTENT=true. Local raw retention can stay a separate opt-in.

Hi @Necmttn agreed, docs and tests updated. As far as I understand the current logic satisfies the split deifinition.

PR #4195, opened 2026-05-27
feat(sdk): expose trace_content via Traceloop.init()

### LINKED ISSUE #137: 🚀 Feature: allow disabling prompt sending as an argument to Traceloop.init()

Traceloop SDK

### 🔖 Feature description

Add a setting called `trace_content` (which will default to `true`) and will control whether all the instrumentations are sending sensitive content (like prompts and completions). This should override the env var behavior (similar to other flags we have).

More control on this sensitive feature

-

### 🔄️ Additional Information

_No response_

None

yoheidemachi:
## Summary

Adds a `trace_content: Optional[bool] = None` argument to `Traceloop.init()` that mirrors the existing `TRACELOOP_TRACE_CONTENT` environment variable, addressing #137 (open since 2023).

- `trace_content=None` (default): the `TRACELOOP_TRACE_CONTENT` env var stays authoritative — no behavior change for existing deployments.
- `trace_content=True` / `False`: the argument explicitly overrides the env var so applications can toggle prompt/completion capture without relying on process-level configuration.

The implementation sets `os.environ["TRACELOOP_TRACE_CONTENT"]` before any instrumentor initializes, so every bundled instrumentation that reads the env var (bedrock, watsonx, alephalpha, writer, …) picks up the value transparently without further changes.

## Design notes

- An explicit arg wins over the env var (consistent with `use_attributes` / `use_legacy_attributes`, which also use `Optional[bool] = None` to signal "not explicitly set").

### Behavior to be aware of

- **Process-wide side effect.** Because the implementation writes `os.environ["TRACELOOP_TRACE_CONTENT"]`, child processes that inherit the env and any other code reading that variable will see the override too. This is the same surface area the existing env var already has — the arg just adds a programmatic entry point.
- **The override is sticky.** Once `trace_content=True`/`False` has been passed, a later `Traceloop.init(trace_content=None)` will not restore the previous env-driven value — the most recent explicit setting keeps winning until the env var is changed externally. Fine for the typical "init once at startup" flow; documented in the docstring.
- **Per-span override still works.** Setting `trace_content=False` does not block the existing `override_enable_content_tracing` OTel context value — instrumentations treat env and context as OR, so individual spans can opt back in while the global default stays off. This preserves the current semantics intentionally.

## Test plan

- [x] `test_trace_content_false_disables_content_tracing` — explicit False overrides env=true
- [x] `test_trace_content_true_overrides_env` — explicit True overrides env=false
- [x] `test_trace_content_none_honors_env` — omitting the arg preserves env-driven behavior
- [x] Existing `test_sdk_initialization.py` + `test_privacy_no_prompts.py` (32 tests) still pass
- [x] `nx run traceloop-sdk:lint` passes

Closes #137

## Release Notes

* **New Features**
  * Introduced `trace_content` parameter to SDK initialization for controlling sensitive data capture in distributed traces. Once configured, the setting persists across subsequent initialization calls, ensuring consistent tracing behavior throughout the application lifecycle. Individual spans can still override the global configuration as needed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4195?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/traceloop-sdk/tests/test_sdk_initialization.py
+++ b/packages/traceloop-sdk/tests/test_sdk_initialization.py
@@ -417,3 +417,26 @@ def test_trace_content_none_honors_env(
 
     assert is_content_tracing_enabled() is False
     assert os.environ["TRACELOOP_TRACE_CONTENT"] == "false"
+
+
+def test_trace_content_override_is_sticky_across_inits(
+    isolated_tracer_wrapper, isolated_trace_content_env
+):
+    """An explicit trace_content setting must persist across subsequent init()
+    calls that omit the arg. Once the env has been overwritten, trace_content=None
+    cannot restore the pre-init env value — the most recent explicit setting wins
+    until something external resets the env. This pins the documented behavior so
+    a future "save/restore env on init exit" refactor would force a docs update."""
+    os.environ["TRACELOOP_TRACE_CONTENT"] = "false"
+
+    Traceloop.init(
+        exporter=InMemorySpanExporter(),
+        disable_batch=True,
+        trace_content=True,
+    )
+    assert os.environ["TRACELOOP_TRACE_CONTENT"] == "true"
+
+    Traceloop.init(exporter=InMemorySpanExporter(), disable_batch=True)
+
+    assert is_content_tracing_enabled() is True
+    assert os.environ["TRACELOOP_TRACE_CONTENT"] == "true"
--- a/packages/traceloop-sdk/tests/test_sdk_initialization.py
+++ b/packages/traceloop-sdk/tests/test_sdk_initialization.py
@@ -364,8 +364,12 @@ def test_both_exporter_and_processor_warns():
 
 @pytest.fixture
 def isolated_trace_content_env():
-    """Save/restore TRACELOOP_TRACE_CONTENT so trace_content tests don't leak."""
-    saved = os.environ.get("TRACELOOP_TRACE_CONTENT")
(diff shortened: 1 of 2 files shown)


GioiaZheng:
Independent privacy review against `376f59e0`:

- All five `TRACELOOP_TRACE_CONTENT × trace_content` matrix rows behaved as documented.
- With `env=true, trace_content=False`, a mock OpenAI span contained neither the synthetic prompt nor completion.
- Input/output token counts, model, duration, and span status were still retained.
- `tests/test_sdk_initialization.py tests/test_privacy_no_prompts.py`: **33 passed**.
- I also confirmed the documented process-global semantics: an explicit `False` writes env=false, a later `init(trace_content=None)` remains false, and concurrent/repeated init calls are last-writer-wins. That seems acceptable only if `init()` is treated as startup-only configuration; the current docstring makes the sticky behavior explicit.

One remaining privacy gap is outside the normal prompt/completion attributes. Using a mock provider 400 response whose error message echoed the synthetic prompt, `trace_content=False` correctly removed `gen_ai.input.messages`, but the prompt still appeared in:

- `exception.message`
- `exception.stacktrace`
- span `status.description`

The span was otherwise correctly marked `ERROR` and retained the model. This looks pre-existing rather than introduced by this PR, so I would not necessarily block #137 on it if the flag is intentionally scoped only to GenAI content fields. However, for a privacy control, it is worth either sanitizing provider exception text when content tracing is disabled or tracking/documenting this boundary explicitly.

PR #3500, opened 2025-12-06
fix(qdrant): support all versions of qdrant package

elinacse:
- [x] I have added tests that cover my changes.
- [ ] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change.
- [x] PR name follows conventional commits format: `feat(instrumentation): ...` or `fix(instrumentation): ...`.
- [ ] (If applicable) I have updated the documentation accordingly.

----

> [!IMPORTANT]
> Enhance Qdrant instrumentation by adding `query_points` and `query_batch_points` methods, updating tests, and adding a sample app.
> 
>   - **Instrumentation Enhancements**:
>     - Add `query_points` and `query_batch_points` methods to `qdrant_client_methods.json`.
>     - Update `_instrument()` in `__init__.py` to handle new methods and log errors if instrumentation fails.
>     - Modify `_wrap()` in `wrapper.py` to support new methods and map them for backward compatibility.
>   - **Testing**:
>     - Update `test_qdrant_instrumentation.py` to test `query_points` and `query_batch_points` methods.
>   - **Sample App**:
>     - Add `qdrant_app.py` to demonstrate usage of `query_points` with a sample dataset.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=traceloop%2Fopenllmetry&utm_source=github&utm_medium=referral)<sup> for d0960215954d7c43d042b477c63dce56cfc8869b. You can [customize](https://app.ellipsis.dev/traceloop/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry instrumentation added for Qdrant query_points and query_batch_points methods
  * New sample app demonstrating Qdrant integration with embeddings

* **Bug Fixes**
  * Improved instrumentation reliability with additional validation and guarded error handling

* **Breaking Changes**
  * Public API: search/search_batch and related request fields renamed to query_points/query_batch_points and QueryRequest/query respectively; update integrations accordingly

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #3492, opened 2025-12-04
🐛 Bug Report: `opentelemetry-instrumentation-qdrant` is incompatible with `qdrant-client` version `1.16.1`


vitalii-diachkov-sumup:
I have just upgraded `qdrant-client` to `v1.16.1` and with `opentelemetry-insturmentation-qdrant` of `v0.48.1` I am getting attribute error. Traceback:

```
chatbot      |   File "/usr/src/chatbot/app/libs/otel.py", line 144, in init_otel_instrumentation
chatbot      |     QdrantInstrumentor().instrument(tracer_provider=otel_provider)
chatbot      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
chatbot      |   File "/usr/src/chatbot/.venv/lib/python3.13/site-packages/opentelemetry/instrumentation/instrumentor.py", line 125, in instrument
chatbot      |     result = self._instrument(  # pylint: disable=assignment-from-no-return
chatbot      |         **kwargs
chatbot      |     )
chatbot      |   File "/usr/src/chatbot/.venv/lib/python3.13/site-packages/opentelemetry/instrumentation/qdrant/__init__.py", line 55, in _instrument
chatbot      |     wrap_function_wrapper(
chatbot      |     ~~~~~~~~~~~~~~~~~~~~~^
chatbot      |         MODULE,
chatbot      |         ^^^^^^^
chatbot      |         f"{wrap_object}.{wrap_method}",
chatbot      |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
chatbot      |         _wrap(tracer, wrapped_method),
chatbot      |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
chatbot      |     )
chatbot      |     ^
chatbot      |   File "/usr/src/chatbot/.venv/lib/python3.13/site-packages/wrapt/patches.py", line 114, in wrap_function_wrapper
chatbot      |     return wrap_object(module, name, FunctionWrapper, (wrapper,))
chatbot      |   File "/usr/src/chatbot/.venv/lib/python3.13/site-packages/wrapt/patches.py", line 60, in wrap_object
chatbot      |     (parent, attribute, original) = resolve_path(module, name)
chatbot      |                                     ~~~~~~~~~~~~^^^^^^^^^^^^^^
[...]

PR #4202, opened 2026-05-30
fix(qdrant): cover `AsyncQdrantClient.query_points` and `query_batch_points`

NishchayMahor:
## Summary

Follow-up to #3492 / #3500 picking up the async-parity gap @kngyeol flagged in the issue thread.

The original crash from `upload_records` is gone since #3500, but the `query_points` / `query_batch_points` mismatch between the sync and async method configs is still present on `main`. Both methods exist on `AsyncQdrantClient` in current `qdrant-client` releases, the sync JSON already lists them, the async JSON does not, so async users get no spans for the modern query APIs.

## Changes

- `async_qdrant_client_methods.json`: add `AsyncQdrantClient.query_points` and `AsyncQdrantClient.query_batch_points` next to the existing `query` / `query_batch` entries, mirroring the sync layout.
- `tests/test_qdrant_instrumentation.py`: add two async tests that drive `query_points` and `query_batch_points` through the in-memory `AsyncQdrantClient`, run via `asyncio.run` (no new test deps), and assert the spans show up with the expected attributes. Used `upsert` for setup since `AsyncQdrantClient.upload_collection` is sync internally and returns `None`.

## Test plan

- [x] `uv run pytest tests/` reports `6 passed` (4 existing sync tests + 2 new async ones).
- [x] `uv run ruff check .` reports `All checks passed!`.
- [x] Verified locally that both methods exist on `AsyncQdrantClient` in current `qdrant-client` so the wrap will not skip them at instrumentation time.

Refs #3492.

## Summary by CodeRabbit

* **New Features**
  * Added OpenTelemetry spans for two asynchronous Qdrant query operations: qdrant.query_points and qdrant.query_batch_points, providing clearer tracing and query-specific attributes.

* **Tests**
  * Added tests that run async queries and verify emitted span names and key attributes (collection, top-k, batch count) to ensure telemetry correctness.

### DIFF
--- a/packages/opentelemetry-instrumentation-qdrant/opentelemetry/instrumentation/qdrant/async_qdrant_client_methods.json
+++ b/packages/opentelemetry-instrumentation-qdrant/opentelemetry/instrumentation/qdrant/async_qdrant_client_methods.json
@@ -46,8 +46,18 @@
     },
     {
         "object": "AsyncQdrantClient",
+        "method": "query_points",
+        "span_name": "qdrant.query_points"
+    },
+    {
+        "object": "AsyncQdrantClient",
         "method": "query_batch",
         "span_name": "qdrant.query_batch"
+    },
+    {
+        "object": "AsyncQdrantClient",
+        "method": "query_batch_points",
+        "span_name": "qdrant.query_batch_points"
     },
     {
         "object": "AsyncQdrantClient",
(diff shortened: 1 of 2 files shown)


NishchayMahor:
Thanks @coderabbitai :), fair callout. The `clear_exporter` autouse fixture in `conftest.py` already wipes the in-memory exporter between tests, so the failure mode you're describing does not actually trigger with the current setup, but you're right that the explicit clear inside each async test makes the contract obvious from the test body alone and survives any future fixture change.

Pushed in `ef41cfc`:

- `exporter.clear()` at the top of both `test_qdrant_async_query_points` and `test_qdrant_async_query_batch_points`, right before the `asyncio.run(...)` block, so the `next(span for span ...)` lookup only sees spans produced by this test.

Re-ran locally, still 6/6 passing. Appreciate the second pair of eyes :)

PR #4203, opened 2026-05-30
fix: mark tracer disabled when tracing is disabled via env var

### LINKED ISSUE #4191: 🐛 Bug Report: TRACELOOP_TRACING_ENABLED does not match Traceloop.init(enabled) parameter behavior

Traceloop SDK

### 📜 Description

`Traceloop.init(enabled=False)` works as a no-op setting. Any subsequent call on instrumented methods are silently skipped.

TRACELOOP_TRACING_ENABLED=false does not work in the same way. E.g calling methods that are decorated with `@workflow()` yields error logs.

### 👟 Reproduction steps

1. set TRACELOOP_TRACING_ENABLED=false
2. run `Traceloop.init()`
3. run:

```python

@workflow(name="hello-world")
def hello_world():
    return "hello world!"

hello_world()
```

### 👍 Expected behavior

2. printed "Traceloop instrumentation is disabled via init flag"

3. nothing

2. printed "Tracing is disabled"

3. printed "Warning: Traceloop not initialized, make sure you call Traceloop.init()"

### 🤖 Python Version

3.12

Probably related to the two different ways of handling disabled in [these lines](https://github.com/traceloop/openllmetry/blob/main/packages/traceloop-sdk/traceloop/sdk/__init__.py#L111C1-L126C9).

koriyoshi2041:
Fixes #4191.

There are two ways to disable tracing, and they diverge:

- `Traceloop.init(enabled=False)` calls `TracerWrapper.set_disabled(True)` and returns, so `@workflow`/`@task` decorators become clean no-ops (`verify_initialized()` returns `False` early when disabled).
- `TRACELOOP_TRACING_ENABLED=false` just prints `"Tracing is disabled"` and returns **without** `set_disabled(True)`.

So with the env var, `TracerWrapper` is never initialized *and* never marked disabled. Decorated functions then hit the uninitialized branch and spam `Warning: Traceloop not initialized, make sure you call Traceloop.init()`.

The fix calls `TracerWrapper.set_disabled(True)` in the env-var path too, mirroring the `enabled=False` branch.

Added `test_tracing_disabled_via_env_var_is_silent_noop` (in `test_sdk_initialization.py`): with `TRACELOOP_TRACING_ENABLED=false`, after `init()` a `@workflow`-decorated call must not print the "not initialized" warning. It fails before the fix and passes after.

(The issue title says `TRACELOOP_TRACKING_ENABLED`, but the actual env var is `TRACELOOP_TRACING_ENABLED` — the behavior is the same regardless.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SDK to operate silently when tracing is disabled via configuration, preventing unnecessary warnings from decorated functions.

* **Tests**
  * Added test to verify that disabled tracing state allows decorated functions to execute as silent no-ops without initialization warnings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/traceloop/openllmetry/pull/4203?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/traceloop-sdk/tests/test_sdk_initialization.py
+++ b/packages/traceloop-sdk/tests/test_sdk_initialization.py
@@ -358,3 +358,25 @@ def test_both_exporter_and_processor_warns():
             del TracerWrapper.instance
         if saved_instance is not None:
             TracerWrapper.instance = saved_instance
+
+
+def test_tracing_disabled_via_env_var_is_silent_noop(
+    isolated_tracer_wrapper, monkeypatch, capsys
+):
+    """When tracing is disabled via TRACELOOP_TRACING_ENABLED=false, init() must
+    mark the tracer disabled so @workflow becomes a silent no-op rather than
+    warning that Traceloop was never initialized (issue #4191)."""
+    monkeypatch.setenv("TRACELOOP_TRACING_ENABLED", "false")
+    try:
+        Traceloop.init(app_name="test")
+        assert TracerWrapper.verify_initialized() is False
+
+        @workflow(name="w")
+        def f():
+            return 42
+
+        capsys.readouterr()  # discard init() output
+        assert f() == 42
+        assert "Traceloop not initialized" not in capsys.readouterr().out
+    finally:
+        TracerWrapper.set_disabled(False)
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
I have read the CLA Document and I hereby sign the CLA

Issue #4230, opened 2026-06-11
🐛 Bug Report: Anthropic event emitter leaks `NOT_GIVEN` sentinel into log Body, breaking OTLP log export


marians:
### Which component is this bug for?

Anthropic Instrumentation

### 📜 Description

When the Anthropic instrumentation is run in **event-logging mode** (`use_legacy_attributes=False`, with an `event_logger_provider` configured), a `messages.create()` call that **omits `tools`** causes the emitted GenAI input event to contain the Anthropic SDK sentinel `anthropic.NOT_GIVEN` in its log record Body. The OTLP log exporter cannot serialize `anthropic.NotGiven`, so the **entire log batch fails to export** with:

```
opentelemetry.sdk._shared_internal - ERROR - Exception while exporting Log.
...
Exception: Invalid type <class 'anthropic.NotGiven'> of value NOT_GIVEN
```

### Root cause

In `opentelemetry/instrumentation/anthropic/event_emitter.py`, `emit_input_events()`:

```python
    if kwargs.get("system"):                       # truthiness — NOT_GIVEN is falsy, correctly skipped
        emit_event(MessageEvent(content=kwargs.get("system"), role="system"), event_logger)
    for message in kwargs.get("messages"):
        emit_event(MessageEvent(content=message.get("content"), role=message.get("role")), event_logger)
    if kwargs.get("tools") is not None:            # BUG: `is not None` lets NOT_GIVEN through
        emit_event(
            MessageEvent(content={"tools": kwargs.get("tools")}, role="user"),   # NOT_GIVEN placed in Body
            event_logger,
        )
```

When a caller omits `tools`, the Anthropic SDK passes `tools=anthropic.NOT_GIVEN`. `NOT_GIVEN is not None` evaluates to **`True`**, so the guard passes and the sentinel object is put verbatim into the `MessageEvent` body
(`{"tools": NOT_GIVEN}`). It then flows into the OTel `LogRecord` body, and the OTLP protobuf encoder
(`opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder`) raises on the unsupported type.

Note the inconsistency: the `system` branch immediately above uses a truthiness check (`if kwargs.get("system"):`), which **correctly** drops `NOT_GIVEN` (`bool(NOT_GIVEN)` is `False`). Only the `tools` branch uses `is not None`, so
only `tools` leaks the sentinel.

### 👟 Reproduction steps

[...]

PR #4248, opened 2026-06-15
fix(anthropic): guard tools event emission with truthiness check

QuentinBisson:
## What

`event_emitter.py` checks `if kwargs.get("tools") is not None:` before emitting
tool `MessageEvent`s. `anthropic.NOT_GIVEN` (the sentinel used for omitted
parameters) is not `None`, so on tool-free requests the sentinel leaks into the
OTLP log body.

## Fix

Change the guard to `if kwargs.get("tools"):`, matching the `system` truthiness
check two lines above.

Severity: observability-only — produces noisy/wrong log records, no functional
impact on the instrumented application.

## Bug Fixes
* Corrected event emission behavior to prevent unnecessary tool-related events when tool definitions are not provided.

## Tests
* Added test coverage for edge cases in tool parameter handling.


QuentinBisson:
Closing in favour of #4236, which makes the same change and was opened first.

Note that #4230 was closed as completed, but the guard on `main` is still `if kwargs.get("tools") is not None:`, so the `NOT_GIVEN` sentinel still leaks into the OTLP log body on tool-free requests. #4236 is green and mergeable.

PR #4261, opened 2026-06-16
fix(langchain): emit cache creation tokens

dvirski:
## Summary by CodeRabbit

* **New Features**
  * Enhanced GenAI cache metrics tracking by reporting both cache read and cache creation token usage (when available).
* **Bug Fixes**
  * Improved cache token metric handling to avoid setting usage attributes with invalid or missing token values, ensuring metrics are recorded only when the underlying data is numeric and meaningful.

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
@@ -393,8 +393,8 @@ def set_chat_response_usage(
     input_tokens = 0
     output_tokens = 0
     total_tokens = 0
-    cache_read_tokens = 0
-    cache_creation_tokens = 0
+    cache_read_tokens = None
+    cache_creation_tokens = None
 
     # Early return if no generations to avoid potential issues
     if not response.generations:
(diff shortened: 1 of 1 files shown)


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
```diff
@@ -424,6 +425,9 @@ def set_chat_response_usage(
                             "input_token_details", {}
                         )
                         cache_read_tokens += input_token_details.get("cache_read", 0)
+                        cache_creation_tokens += input_token_details.get(
+                            "cache_creation", 0
+                        )
```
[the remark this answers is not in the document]
dvirski: The LangChain docs confirm cache_read and cache_creation are both plain int fields — no dict variant exists.
So this is over defensive.

keeping only int/float guards:

  raw_cache_read = input_token_details.get("cache_read", 0)
  if isinstance(raw_cache_read, (int, float)):
      cache_read_tokens += raw_cache_read

  raw_cache_creation = input_token_details.get("cache_creation", 0)
  if isinstance(raw_cache_creation, (int, float)):
      cache_creation_tokens += raw_cache_creation

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
```diff
@@ -423,7 +424,12 @@ def set_chat_response_usage(
                         input_token_details = generation.message.usage_metadata.get(
                             "input_token_details", {}
                         )
-                        cache_read_tokens += input_token_details.get("cache_read", 0)
+                        raw_cache_read = input_token_details.get("cache_read")
+                        if isinstance(raw_cache_read, (int, float)):
+                            cache_read_tokens = (cache_read_tokens or 0) + raw_cache_read
+                        raw_cache_creation = input_token_details.get("cache_creation")
+                        if isinstance(raw_cache_creation, (int, float)):
+                            cache_creation_tokens = (cache_creation_tokens or 0) + raw_cache_creation
```
nina-kollman, line 432: Why not init the 'cache_creation_tokens' to 0 ? same for cache_read_tokens

nina-kollman (approved):

packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/span_utils.py
dvirski: cache_creation_tokens and cache_read_tokens might not be provided at all. so we want to distinguish between not provided and 0.

If not provided we omit them, but we do want to log them if they have value of 0.

PR #4288, opened 2026-06-18
fix(sdk): @agent decorator no longer overwrites enclosing @workflow name

dvirski:
The @agent decorator unconditionally called set_workflow_name(entity_name),
  clobbering the workflow_name attached to the OTel context by an enclosing
  @workflow. Child LLM spans then read the agent name as their workflow name,
  making (agent_name, workflow_name) groupings collapse: the same agent
  running under two different workflows became indistinguishable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved observability trace naming so workflow names are applied only to the appropriate span types, preventing agent spans from overwriting workflow context.

* **Tests**
  * Added regression coverage verifying workflow/agent name behavior for nested and standalone agents, including distinct workflow names across multiple workflows.

### DIFF
--- a/packages/traceloop-sdk/traceloop/sdk/decorators/base.py
+++ b/packages/traceloop-sdk/traceloop/sdk/decorators/base.py
@@ -137,14 +137,12 @@ def _is_async_method(fn):
 
 def _setup_span(entity_name, tlp_span_kind, version):
     """Sets up the OpenTelemetry span and context"""
-    if tlp_span_kind in [
-        TraceloopSpanKindValues.WORKFLOW,
-        TraceloopSpanKindValues.AGENT,
-    ]:
+    if tlp_span_kind == TraceloopSpanKindValues.WORKFLOW:
         set_workflow_name(entity_name)
-
-    if tlp_span_kind == TraceloopSpanKindValues.AGENT:
+    elif tlp_span_kind == TraceloopSpanKindValues.AGENT:
         set_agent_name(entity_name)
+        if context_api.get_value("workflow_name") is None:
+            set_workflow_name(entity_name)
 
     span_name = f"{entity_name}.{tlp_span_kind.value}"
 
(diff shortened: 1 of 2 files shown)


packages/traceloop-sdk/traceloop/sdk/decorators/base.py
```diff
@@ -137,14 +137,12 @@ def _is_async_method(fn):
...
-        TraceloopSpanKindValues.AGENT,
-    ]:
+    if tlp_span_kind == TraceloopSpanKindValues.WORKFLOW:
         set_workflow_name(entity_name)
-
-    if tlp_span_kind == TraceloopSpanKindValues.AGENT:
+    elif tlp_span_kind == TraceloopSpanKindValues.AGENT:
         set_agent_name(entity_name)
+        if context_api.get_value("workflow_name") is None:
+            set_workflow_name(entity_name)
```
max-deygin-traceloop: so if "workflow_name" is none we set entity_name as workflow name.

But if it's not None? It's never set?

packages/traceloop-sdk/traceloop/sdk/decorators/base.py
```diff
@@ -137,14 +137,12 @@ def _is_async_method(fn):
...
-        TraceloopSpanKindValues.WORKFLOW,
-        TraceloopSpanKindValues.AGENT,
-    ]:
+    if tlp_span_kind == TraceloopSpanKindValues.WORKFLOW:
         set_workflow_name(entity_name)
-
-    if tlp_span_kind == TraceloopSpanKindValues.AGENT:
+    elif tlp_span_kind == TraceloopSpanKindValues.AGENT:
         set_agent_name(entity_name)
+        if context_api.get_value("workflow_name") is None:
```
galzilber: const it please no magic strings

dvirski: this is not not relevant since it's removed

packages/traceloop-sdk/traceloop/sdk/decorators/base.py
dvirski: This was to keep the previous behavior in which agent spans that had no workflow parent were populating workflow name with the agent name itself, instead of leaving it empty.

But that's actually a wrong behavior.

After checking the web app and making sure it won't break anything. I removed this block.
It make senes that an agent span that's not part of a workflow will have empty workflow name.

nina-kollman (approved):

PR #4334, opened 2026-06-30
fix: add missing [] defaults to 5 .get() calls in Bedrock span_utils

truecallerabreham:
Five .get() calls in span_utils.py iterate over the result without providing a default, so when the response body doesn't include the expected key you get a TypeError instead of gracefully skipping.

The pattern already exists on line 220 (response_body.get("generations", [])) but the other five sites using "generations", "completions", and "results" were missed. This hits Cohere, AI21, Meta/Llama, and Amazon Titan models in the non-streaming path.

Fix is adding [], nothing else changes.

**Before / After:**

terminal_screenshot


truecallerabreham:
CLA was signed but status shows pending — could someone trigger a recheck? Thanks.

PR #4349, opened 2026-07-04
fix(qdrant): skip missing client methods

1cbyc:
## What Problem This Solves

versions that no longer expose  were still being instrumented, which can raise an  during .

## Why This Change Was Made
The instrumentor now uses a static attribute lookup before wrapping methods, so dynamic class lookup cannot falsely report removed methods as present. I also added a regression test with a fake Qdrant client that only exposes  through metaclass lookup.
## User Impact
Users on newer  releases can instrument Qdrant without the startup crash. Existing spans for supported methods still work.
## Evidence
- ============================= test session starts ==============================
platform linux -- Python 3.10.19, pytest-8.4.2, pluggy-1.6.0
rootdir: /tmp/openllmetry-3492/packages/opentelemetry-instrumentation-qdrant
configfile: pyproject.toml
plugins: anyio-4.12.1, sugar-1.0.0
collected 6 items

tests/test_missing_methods.py ..                                         [ 33%]
tests/test_qdrant_instrumentation.py ....                                [100%]

============================== 6 passed in 0.05s ===============================\n- 
All checks passed!\n\nFixes traceloop/openllmetry#3492

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Qdrant instrumentation to only wrap/unwrap methods that are statically present on the client.
  * Avoided potential issues caused by dynamically resolved attributes during instrumentation and cleanup.
  * Increased reliability for both sync and async Qdrant client integrations.
* **Tests**
  * Added coverage to verify methods that are dynamically provided are skipped for both instrumenting and uninstrumenting.


1cbyc:
Added companion coverage for the static-attribute check in the `_uninstrument` path.

Verification run locally from `packages/opentelemetry-instrumentation-qdrant`:
- `uv run --group test pytest tests/test_missing_methods.py` -> 3 passed
- `uv run ruff check tests/test_missing_methods.py` -> passed

PR #4367, opened 2026-07-17
feat(traceloop-sdk): add trace_content parameter to Traceloop.init()

### LINKED ISSUE #137: 🚀 Feature: allow disabling prompt sending as an argument to Traceloop.init()

smz202000:
## Summary

Adds a `trace_content` parameter to `Traceloop.init()` that allows users to control whether sensitive content (prompts and completions) is sent in traces.

Fixes #137

## Changes

- Modified `is_content_tracing_enabled()` in `traceloop/sdk/config/__init__.py` to accept an optional `trace_content` parameter. When explicitly provided, it overrides the `TRACELOOP_TRACE_CONTENT` environment variable. When not provided (None), it falls back to the environment variable (backward compatible).
- Added `trace_content: Optional[bool] = None` parameter to `Traceloop.init()` in `traceloop/sdk/__init__.py`.
- Added docstring documentation for the new parameter.

## Usage

```python
# Disable content tracing programmatically (overrides env var)
Traceloop.init(app_name="my-app", trace_content=False)

# Enable content tracing explicitly (overrides env var)
Traceloop.init(app_name="my-app", trace_content=True)

# Default behavior: uses TRACELOOP_TRACE_CONTENT env var
Traceloop.init(app_name="my-app")
```

## Backward Compatibility

This change is fully backward compatible. When `trace_content` is not passed (the default), the existing `TRACELOOP_TRACE_CONTENT` environment variable behavior is preserved.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to control whether prompt and completion content is included in traces.
  * Added configuration options for enabling or disabling metrics and logging.
  * Content tracing now supports explicit configuration while retaining environment-based defaults.


GioiaZheng:
Independent privacy review against `0da96148`: the SDK-level override does not currently reach the actual instrumentation gates.

I ran the same matrix against `TracerWrapper.enable_content_tracing`, OpenAI's `should_send_prompts()`, and the SDK decorator gate:

| env | arg | SDK | OpenAI/decorator |
|---|---:|---:|---:|
| unset | None | true | true |
| true | False | false | **true** |
| false | True | true | **false** |
| false | None | false | false |
| true | None | true | true |

Minimal regression check:

```python
monkeypatch.setenv("TRACELOOP_TRACE_CONTENT", "true")
Traceloop.init(
    exporter=InMemorySpanExporter(),
    disable_batch=True,
    trace_content=False,
)
assert TracerWrapper.enable_content_tracing is False
assert should_send_prompts() is False  # currently True
```

I also exercised an OpenAI chat span with an in-memory exporter and a mock HTTP transport (no provider call). With `env=true, trace_content=False`, the emitted `openai.chat` span still contained both the synthetic prompt and completion, including `gen_ai.input.messages`, although the SDK flag was false. The inverse conflict (`env=false, trace_content=True`) also leaves the instrumentation disabled.

The existing SDK privacy tests pass (29/29) because they disable content through the environment variable and therefore do not cover this split-brain case. I think this needs an end-to-end regression test at the instrumentation/span boundary before merge; changing only `is_content_tracing_enabled()` is not sufficient while provider instrumentations and decorators independently read `TRACELOOP_TRACE_CONTENT`.

Issue #4340, opened 2026-07-01
Mapping external AI-task receipt IDs into LLM trace/span metadata


0xddneto:
I am exploring how [AI Proof of Us](https://github.com/0xddneto/AI-Proof-of-Us) receipts should map into LLM traces.

AIPOU creates signed MCP task receipts with hashes, nonce/replay checks, local collector signatures, and explicit evidence boundaries. It does not replace traces, and trace systems should not be responsible for validating AIPOU rewards.

For observability, the useful part may be a small external reference:

```json
{
  "aipou.receipt_id": "0x...",
  "aipou.validation_status": "local | validated | claimed | rejected",
  "aipou.evidence_boundary": "https://github.com/0xddneto/AI-Proof-of-Us/blob/main/docs/evidence-boundaries.md"
}
```

Would this belong as span attributes, session metadata, a linked event, or an external reference? I am trying to keep raw prompts/outputs private and avoid noisy or misleading metadata.

Interoperability note:

https://github.com/0xddneto/AI-Proof-of-Us/blob/main/docs/receiptid-interoperability.md

0xddneto:
Follow-up for observability reviewers:

AIPOU now documents tokenomics and launch transparency here: https://github.com/0xddneto/AI-Proof-of-Us/blob/main/docs/tokenomics.md

The intended trace relationship is intentionally narrow: traces should not validate rewards or token claims. They can optionally carry a receipt reference such as `aipou.receipt_id`, while AIPOU keeps validation and settlement separate.

Human users benefit because approved AI-work receipts can later claim rewards. Agents and platforms benefit because the receipt is portable work evidence that can be referenced without exposing raw prompts/outputs.

The simplest non-payment test path is Local Receipt Mode: https://github.com/0xddneto/AI-Proof-of-Us/tree/main/examples/local-receipt-mode

0xddneto:
Small AIPOU update for the trace/span metadata question:

AIPOU now has `settle_all_ai_rewards`, so explicit user settlement can process all currently eligible pending AI-work receipts from the shared `AIPOU_DATA_DIR` in bounded batches. The receipt remains external to tracing: hashes, provider/model metadata, wallet authorization, nonce/replay checks, collector signature, and optional validator-approved claims.

Token display metadata is now available:
- transparent image: https://raw.githubusercontent.com/0xddneto/AI-Proof-of-Us/main/assets/token/aipou.png
[...]

PR #4373, opened 2026-07-22
feat(semconv): add AIPOU work-receipt reference span attributes

Diyaaa-12:
Follow-up to discussion in #4340.

Adds four opt-in SpanAttributes constants for referencing an external
AIPOU work receipt:
- AIPOU_WORK_RECEIPT_ID
- AIPOU_EVIDENCE_CLASS
- AIPOU_SCHEME
- AIPOU_VALIDATION_STATUS

Scope is intentionally minimal per @0xddneto's proposed boundary:
constants + tests only, no helper function yet — want to confirm the
naming/shape here before adding one.

Tests added in `_testing.py` (shared compliance suite, reused by both
`test_semconv_compliance.py` and `test_span_attributes.py`).

Note: 30 unrelated pre-existing failures in this suite come from a
missing `opentelemetry-semantic-conventions` dependency in my local
env — confirmed via `git stash` that they fail identically without
this change.

Ref: traceloop/openllmetry#4340

## Summary by CodeRabbit

* **Enhancements**
  * Added four new deprecated AI observability span attributes for work receipt ID, evidence class, scheme, and validation status.

* **Tests**
  * Added coverage to verify these new attributes map to the expected standardized `aipou.*` values.


0xddneto:
Thank you, @Diyaaa-12. This is the minimal integration surface we discussed, and keeping it to constants plus conformance tests is a good first step. The four fields are correlation-only; the authoritative receipt and validation lifecycle remain external to OpenLLMetry.

I found one important placement issue before maintainer review: the new constants currently sit immediately below `# Deprecated`, so CodeRabbit is describing all four AIPOU attributes as deprecated. Could you move them into a dedicated section such as `# External work receipt references (opt-in)` outside that block? The trailing newline flagged by CodeRabbit should also be added.

With that correction, the scope matches the agreed privacy and trust boundary: meaningful work-unit spans may carry the compact reference, while OpenLLMetry does not collect private receipt payloads, derive validation status, or become claim authority. No helper is needed in this PR.

0xddneto:
I applied the complementary AIPOU-side reference projection in https://github.com/0xddneto/AI-Proof-of-Us/commit/90551cf.

`buildReceiptSpanAttributes` now emits exactly the four strings proposed here:

- `aipou.work_receipt_id`
- `aipou.evidence_class`
- `aipou.scheme`
- `aipou.validation_status`

The conformance test also asserts that wallet, reward, prompt, output, and claim authority are absent. The MCP suite passes 38/38, the production dependency audit reports zero vulnerabilities, and the package dry-run succeeds. This gives the PR an executable producer-side reference without requiring OpenLLMetry to add a helper or validate AIPOU receipts.

I also saw that the trailing-newline fix landed immediately, thank you. The only review item still visible in the current diff is moving the constants out from under the existing `# Deprecated` heading so their status is unambiguous.

Diyaaa-12:
Thank You @0xddneto ! Fixed  moved the AIPOU constants out from under the `# Deprecated` heading into their own `# External work receipt references (opt-in)` section, right after the Watson block. Pushed in cce055924's follow up commit. Ready for another look whenever convenient.

0xddneto:
Reviewed commit `301044a`. The deprecation concern is resolved, the four names match the AIPOU producer-side projection, and the privacy/trust boundary is correct. Thank you for the quick fix.

One small source-organization cleanup remains: the AIPOU section currently sits between `GEN_AI_WATSONX_DECODING_METHOD` and the other Watson attributes (`RANDOM_SEED`, `MAX_NEW_TOKENS`, and so on), so it splits the Watson block. Could you move the complete AIPOU section after all Watson attributes, or immediately before the `# Watson/genai LLM` heading? This does not change behavior; it just keeps the semantic-convention groups coherent. From the AIPOU side, the naming and boundary are otherwise ready for maintainer review.

Diyaaa-12:
Hi @0xddneto ! Fixed  moved the complete AIPOU section to right after `GEN_AI_WATSONX_REPETITION_PENALTY`, so the Watson block is no longer split. Pushed in the latest commit. Thanks for catching that.

Diyaaa-12:
Hi @0xddneto ! waiting for review

0xddneto:
Sorry for the slow loop closure here, and thank you for the patience.

The honest update from the AIPOU side is that my preferred direction has shifted since this PR was opened. I still think your draft is coherent on its own terms, but the direction I would recommend upstream today is the newer **generic external evidence** shape rather than vendor-specific ipou.* keys.

In other words:

- this PR correctly keeps the boundary correlation-only;
- but the longer-lived semantic-conventions fit looks more like gen_ai.external_evidence.* fields that any receipt or audit artifact can use;
- AIPOU should then be one scheme carried by that generic surface, not four dedicated constants in core semconv.

So from my side I would not ask you to keep pushing this AIPOU-specific PR forward unless the maintainers explicitly want a vendor-specific example. If the project wants the more general path instead, I am happy to help with fixture wording and boundary notes there.

Diyaaa-12:
Thanks for the context, @0xddneto  that makes sense as the more durable design.

Before I put time into reworking this PR: would it be worth confirming with the maintainers first whether they'd want a generic `gen_ai.external_evidence.*` surface at all, versus keeping this as a vendor-specific example? I don't want to rewrite the constants/tests to that shape and then have it stall for the same reason on the generic side.

Happy to do either  just want to make sure there's actual appetite before I put in the rework. Let me know what you (or the maintainers) think is the better next step.

0xddneto:
That sounds right to me. I would not want you to spend time rewriting this into generic `gen_ai.external_evidence.*` fields unless maintainers first confirm they actually want that surface upstream.

From the AIPOU side, my current preference is:
- keep this PR coherent as the vendor-specific example it already is, or
- pause and ask maintainers whether there is appetite for the generic external-evidence shape before doing the rewrite.

So I agree with your instinct here: confirm maintainer appetite first, then rework only if there is a real path for it.

Issue #2338, opened 2024-11-26
🐛 Bug Report: CohereInstrumentor do not instrument anything


bbaabemhp:
### Which component is this bug for?

Cohere Instrumentation

### 📜 Description

I use the CohereInstrumentor, but it does not create any spans for my trace.

### 👟 Reproduction steps

```python
import logging

from azure.monitor.opentelemetry import configure_azure_monitor
from opentelemetry import trace
from opentelemetry.instrumentation.cohere import CohereInstrumentor
from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
from opentelemetry.instrumentation.langchain import LangchainInstrumentor
from opentelemetry.instrumentation.logging import LoggingInstrumentor
from opentelemetry.instrumentation.openai import OpenAIInstrumentor
from opentelemetry.instrumentation.psycopg import PsycopgInstrumentor
from opentelemetry.instrumentation.requests import RequestsInstrumentor
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
from dataproduct.settings import settings
from opentelemetry.instrumentation.asyncio import AsyncioInstrumentor


def configure_tracing(app=None):
    if settings.APPLICATIONINSIGHTS_CONNECTION_STRING:
        configure_azure_monitor(
            connection_string=settings.APPLICATIONINSIGHTS_CONNECTION_STRING,
            enable_live_metrics=True,
            logger_name="dataproduct",
        )
        print("Config done")
    else:
        provider = TracerProvider()
        console_exporter = ConsoleSpanExporter()
        provider.add_span_processor(BatchSpanProcessor(console_exporter))
        trace.set_tracer_provider(provider)
    RequestsInstrumentor().instrument()
    HTTPXClientInstrumentor().instrument()
    AsyncioInstrumentor().instrument()
    CohereInstrumentor().instrument()
    OpenAIInstrumentor().instrument()
    # LangchainInstrumentor().instrument()
    PsycopgInstrumentor().instrument(enable_commenter=False, skip_dep_check=True)
    LoggingInstrumentor().instrument(
        set_logging_format=True,
        log_level=logging.DEBUG,
    )
    if app:
        FastAPIInstrumentor.instrument_app(app)
    return trace.get_tracer(
        __name__,
    )

_ = configure_tracing()

[...]

PR #4375, opened 2026-07-24
fix(cohere): avoid double-ending async error spans

euisuh:
## Summary

Fixes Cohere async instrumentation error handling so spans are not ended twice when an async Cohere call raises.

`_awrap()` uses `tracer.start_as_current_span(...)`, which already ends the span when the context manager exits. The exception handler also called `span.end()`, causing OpenTelemetry to emit:

```text
Calling end() on an ended span.
```

This removes the redundant manual end call while preserving ERROR status, exception recording, and re-raising behavior.

## Related

Related to #2338, which tracks Cohere async instrumentation behavior.

## Test Plan

- [x] `uv run pytest tests/test_async_error_handling.py tests/test_rerank.py::test_cohere_v2_rerank_legacy_async tests/test_chat.py::test_cohere_chat_legacy_async -q`
- [x] `uv run pytest tests/ -q`
- [x] `uv run ruff check .`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenTelemetry error handling for asynchronous Cohere chat requests.
  * Exceptions are still recorded with `ERROR` status, and spans now finish cleanly without duplicate-end warnings.
* **Tests**
  * Added an async test to verify the exception is raised once, the span is reported once with the correct error details, and includes an `exception` event.

### DIFF
--- a/packages/opentelemetry-instrumentation-cohere/opentelemetry/instrumentation/cohere/__init__.py
+++ b/packages/opentelemetry-instrumentation-cohere/opentelemetry/instrumentation/cohere/__init__.py
@@ -287,7 +287,6 @@ async def _awrap(
             if span.is_recording():
                 span.set_status(Status(StatusCode.ERROR, str(e)))
                 span.record_exception(e)
-                span.end()
             raise
 
         set_span_response_attributes(span, response)
--- a/packages/opentelemetry-instrumentation-cohere/tests/test_async_error_handling.py
+++ b/packages/opentelemetry-instrumentation-cohere/tests/test_async_error_handling.py
@@ -32,4 +32,11 @@ async def test_async_cohere_exception_records_error_once(
     assert span.name == "cohere.chat"
     assert span.status.status_code == StatusCode.ERROR
     assert "cohere async failure" in span.status.description
-    assert not any("Calling end() on an ended span" in record.message for record in caplog.records)
+    assert any(
+        event.name == "exception"
+        and "cohere async failure" in event.attributes.get("exception.message", "")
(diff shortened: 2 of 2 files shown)


packages/opentelemetry-instrumentation-cohere/opentelemetry/instrumentation/cohere/__init__.py
```diff
@@ -287,7 +287,6 @@ async def _awrap(
             if span.is_recording():
                 span.set_status(Status(StatusCode.ERROR, str(e)))
                 span.record_exception(e)
```
gnanirahulnutakki, line 289: The outer `start_as_current_span(...)` context manager already records any exception that escapes this block (its default is `record_exception=True`), so this explicit call still creates duplicate exception telemetry even though the duplicate `end()` is gone. I reproduced this on `9c303fd`: changing the new test to `assert len(exception_events) == 1` fails with `2 == 1`.

Please give one layer sole ownership of exception recording—e.g. remove this manual `try/except` and let the context manager set ERROR/record the exception, or disable its automatic exception handling—and make the regression assert exact cardinality. With the manual handler removed, the strengthened test plus `test_cohere_v2_rerank_legacy_async` and `test_cohere_chat_legacy_async` all pass locally.

gnanirahulnutakki (commented):
I reproduced one telemetry-correctness issue on the submitted head. The inline comment includes the failing exact-cardinality assertion and a locally verified minimal fix; otherwise the change is focused and the stated async tests pass.

PR #4377, opened 2026-07-26
fix(anthropic): stop double-counting output tokens on streamed messages

arthi-arumugam-git:
## Problem

`_process_response_item` treats `message_delta.usage.output_tokens` as an increment and adds it to the count already stored from `message_start`:

```python
item_output_tokens = dict(item.usage).get("output_tokens", 0)
existing_output_tokens = complete_response["usage"].get("output_tokens", 0)
complete_response["usage"]["output_tokens"] = (
    item_output_tokens + existing_output_tokens
)
```

Anthropic sends `message_delta.usage.output_tokens` as the running total for the whole message. The value in `message_start` is a partial count that is already contained in it. Adding them inflates output tokens, and therefore total tokens and any cost computed downstream, on every streamed Anthropic span. Nothing raises, so the wrong number just propagates.

The vendor SDK assigns rather than adds, in `anthropic/lib/streaming/_messages.py`:

```python
current_snapshot.usage.output_tokens = event.usage.output_tokens
```

This repo's own Bedrock streaming wrapper also overwrites rather than accumulates (`bedrock/streaming_wrapper.py`, `usage.update(...)`).

## Reproduction

Replaying the SSE events from this repo's existing cassette, `tests/cassettes/test_messages/test_anthropic_message_streaming_legacy.yaml`, through both accumulators:

```
message_start.usage : {'input_tokens': 17, 'output_tokens': 3}
message_delta.usage : {'output_tokens': 171}

openllmetry  output_tokens : 174
anthropic SDK output_tokens: 171   <- ground truth
```

A 1.75% overcount on this recording. The gap is whatever `message_start` reported, so it grows with prompt-cache and tool-use shapes where that initial count is larger.

## Fix

Overwrite with the delta's value instead of adding, and leave the stored count alone when the delta omits `output_tokens`.

## Test

`test_anthropic_message_streaming_legacy` already asserted `input_tokens == 17` and the `input + output == total` identity, but never asserted the output value itself, so the bug passed CI. This PR adds that one assertion. It fails with `assert 174 == 171` on `main` and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected streamed output token reporting to use the provider’s cumulative totals, preventing inflated usage counts.
  * Ensured usage metrics remain accurate when token totals are omitted from individual streaming updates.


arthi-arumugam-git:
Gentle ping: CodeRabbit, GitGuardian, and the CLA are all green, and the branch is current as of yesterday. The test workflows are still waiting on the fork workflow approval click; would a maintainer be able to approve the run so CI can report on this change?

roy-tong:
Adding a piece of authoritative evidence to this thread: the official Anthropic SDK documents `message_delta.usage` as **cumulative** — [anthropic-sdk-python `_messages.py:520-524`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/_messages.py#L520-L524) implements accumulate as *"cumulative totals... overwrite rather than add"*. So summing `output_tokens` across `message_delta` events multiplies the final value by the event count (a recorded stream with deltas 2 then 6 reports 8 instead of 6 — reproducible with the cassette in this repo). The overwrite-style accumulation the SDK ships is the shape the fix here needs. Same class we've confirmed in the wild in a few downstream tools (e.g. one-api-pro#13, +50.7% billed input with no cache).

PR #4386, opened 2026-07-29
fix(langchain): handle early LangGraph stream closure

### LINKED ISSUE #4385: 🐛 Bug Report: LangGraph instrumentation incorrectly marks successful astream() executions as ERROR due to GeneratorExit

Langchain Instrumentation

### 📜 Description

While using the TraceLoop LangGraph instrumentation (opentelemetry-instrumentation-langchain), successful LangGraph executions are incorrectly reported as ERROR.

The issue occurs when consuming graph.astream() with an async for loop and returning once the desired result is produced. During normal async generator shutdown, Python raises GeneratorExit to close the underlying generator. This is expected control flow and does not indicate an application error.

However, the instrumentation catches GeneratorExit as a BaseException and records it as a span failure. As a result, spans such as invoke_agent and, in our environment, the associated workflow span are marked with `StatusCode.ERROR `even though the workflow completes successfully.

This leads to misleading traces in observability backends like Langfuse, where successful executions appear as failures and can negatively impact debugging, monitoring, and error-rate metrics.

### 👟 Reproduction steps

1. Instrument a LangGraph application using the TraceLoop SDK.
2. Execute a graph using graph.astream().
3. Consume the stream using an async for loop.
4. Return from the loop once the desired output has been produced (instead of consuming the generator to exhaustion).
Example:
async for event in graph.astream(inputs):
    if is_final_response(event):
        return response
5. Inspect the generated trace in Langfuse (or another OpenTelemetry backend).

### 👍 Expected behavior

Successful LangGraph executions should produce successful spans.
Normal async generator shutdown (GeneratorExit) should be treated as expected Python control flow and should not mark spans as ERROR.

The instrumentation catches GeneratorExit while wrapping graph.astream() and records it as an application failure.

As a result:
1. invoke_agent spans are marked as StatusCode.ERROR.
2. Workflow spans are also marked as StatusCode.ERROR.
3. The application itself completes successfully and returns the expected response.

### 🤖 Python Version

Python 3.12

The current instrumentation wraps stream()/astream() using except BaseException, which also catches GeneratorExit.
Since GeneratorExit inherits from BaseException (not Exception), normal async generator shutdown is incorrectly treated as an application error.
A possible fix would be to either:
handle GeneratorExit separately and immediately re-raise it without modifying the span, or
catch Exception instead of BaseException, aligning with OpenTelemetry's handling of control-flow exceptions.
In our testing, adding a dedicated GeneratorExit branch before the generic exception handling prevents successful LangGraph executions from being reported as failed while preserving normal error reporting for genuine exceptions.

- [ ] I checked and didn't find similar issue

dk3yyyy:
- [x] (If applicable) I have updated the documentation accordingly. *(Not applicable: this bug fix does not change documented user-facing behavior.)*

## Summary

- treat `GeneratorExit` from early `LangGraph.astream()` termination as normal control flow
- finish the workflow and graph spans without error status or exception events
- preserve error recording and propagation for genuine application exceptions
- safely detach context tokens finalized from a copied async context without emitting misleading traceback logs

## Tests

- `npx nx run opentelemetry-instrumentation-langchain:lint`
- `npx nx run opentelemetry-instrumentation-langchain:test`
- focused early-return and genuine-error regressions on Python 3.11 and Python 3.12

Closes #4385

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tracing for asynchronous streaming workflows that end early, preventing normal termination from being reported as an error.
  - Ensured genuine errors during streaming are correctly marked as failures and include exception details.
  - Improved context cleanup to avoid misleading detach-related errors during asynchronous operations.
- **Tests**
  - Added coverage for early stream termination and error recording in asynchronous workflows.


karthik2oo4-git:
Could you please take a look at this PR? It looks correct to me and seems to address the issue I reported. It’s been open for a few days, so it would be great if someone could review it. Thanks!

karthik2oo4-git:
any update on this PR?

PR #4415, opened 2026-08-12
fix(vertexai): make dont_throw work on async functions

### LINKED ISSUE #4414: 🐛 Bug Report: dont_throw has no effect on async functions in VertexAI instrumentation

VertexAI Instrumentation

### 📜 Description

`dont_throw` in `opentelemetry/instrumentation/vertexai/utils.py` defines only a synchronous wrapper:

```python
def dont_throw(func):
    logger = logging.getLogger(func.__module__)

    def wrapper(*args, **kwargs):
        try:
            return func(*args, **kwargs)
        except Exception as e:
            logger.debug(...)

    return wrapper
```

Calling an `async def` returns a coroutine immediately, so the `try` block exits before the function body runs and the caller awaits the coroutine outside the guard. The decorator has no effect on async functions.

It is applied to async functions in this package: `span_utils.set_input_attributes` and `_handle_request` in `__init__.py`. `_handle_request` is awaited with no surrounding try, and before the wrapped model call, so an instrumentation error in async prompt capture propagates into the user's application and the LLM call never happens.

The Anthropic, OpenAI and google-generativeai packages all branch on `asyncio.iscoroutinefunction` and return an async wrapper. VertexAI and Cohere do not.

### 👟 Reproduction steps

```python
from opentelemetry.instrumentation.vertexai.utils import dont_throw
import asyncio

@dont_throw
async def boom():
    raise RuntimeError("instrumentation failed")

asyncio.run(boom())
```

### 👍 Expected behavior

The exception is logged and swallowed, as it is for synchronous functions.

```
RuntimeError: instrumentation failed
```

The exception reaches the caller. Applied to `_handle_request`, that means the user's Vertex AI call is never made.

### 🤖 Python Version

3.12

Cohere has the same gap, but its async streaming entry points are async generators and need a third branch, so it is worth handling separately.

chiruu12:
`dont_throw` in the VertexAI instrumentation defines only a synchronous wrapper. Calling an `async def` returns a coroutine immediately, so the `try` block exits before the function body runs and the caller awaits the coroutine outside the guard. The decorator has no effect on async functions.

It is applied to async functions in this package: `span_utils.set_input_attributes` and `_handle_request`. `_handle_request` is awaited at `__init__.py` with no surrounding try, and before the wrapped model call, so an instrumentation error in async prompt capture propagates into the user's application and the LLM call never happens.

Before:

```
await async_boom() -> RuntimeError: instrumentation failed
```

After:

```
await async_boom() -> None
await async_ok()   -> ok
```

The Anthropic, OpenAI and google-generativeai packages already branch on `asyncio.iscoroutinefunction` and return an async wrapper. This copies that shape into VertexAI.

Cohere has the same gap, but its async streaming entry points are async generators, which need a third branch. Left out to keep this focused.

Tests added in `tests/test_dont_throw.py` cover both wrappers, the exception path and the return-value path. They fail on main for the async cases.

Fixes #4414

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved exception handling for asynchronous operations so errors are suppressed and logged consistently with synchronous operations.
  - Preserved return values for both synchronous and asynchronous operations.
  - Ensured errors from optional error reporting do not interrupt application behavior.

- **Tests**
  - Added coverage for asynchronous exception handling, return values, wrapper selection, and resilient error reporting across supported operation types.


chiruu12:
@dvirski could you approve the workflow runs here when you get a chance? Only GitGuardian has run, so the required build, lint and test checks have never started and the PR sits blocked with nothing to show.

The diff is small: `dont_throw` wraps the function in a sync wrapper, so on an async function it returns the coroutine unawaited and the exception escapes anyway. This adds an async wrapper and a test for both paths.

chiruu12:
Still green and still mergeable here, along with #4418 and #4420. No conflicts on any of the three.

Is there anything you want changed before these can be looked at? Happy to split them up or rebase if that helps.

PR #4418, opened 2026-08-13
fix(google-generativeai): emit choice events from the last chunk

### LINKED ISSUE #4417: 🐛 Bug Report: streaming choice events use the drained generator instead of the last chunk

Google Generative AI Instrumentation

### 📜 Description

Both streaming builders in `opentelemetry/instrumentation/google_generativeai/__init__.py` pass the generator to `emit_choice_events` instead of the chunk:

```python
complete_response = "".join(text_parts)

if emit_events:
    emit_choice_events(response, event_logger)
else:
    if last_chunk is not None and getattr(last_chunk, "candidates", None):
        set_response_attributes(span, last_chunk, llm_model)
```

By that point `response` has been fully drained by the `for item in response` loop above. `emit_choice_events` does `for index, candidate in enumerate(response.candidates)`, and a generator has no `candidates`. The non-events branch immediately below already uses `last_chunk`, which is the object that carries them.

This only fires when events are enabled, so `use_legacy_attributes=False`.

Two things make it worse than a dropped span. `_build_from_streaming_response` is not decorated `@dont_throw`, unlike the other handlers in the file, and it has no `try/finally` around `span.end()`. So the `AttributeError` is raised out of the generator on the final `next()`, into the calling application, and the span is never ended or exported.

### 👟 Reproduction steps

Instrument with events enabled and stream a response:

```python
GoogleGenerativeAiInstrumentor(use_legacy_attributes=False).instrument()

for chunk in client.models.generate_content_stream(model="gemini-2.0-flash", contents="hi"):
    print(chunk.text)
```

### 👍 Expected behavior

Choice events are emitted from the final chunk, the stream is consumed cleanly, and the span is exported.

```
AttributeError: 'generator' object has no attribute 'candidates'
```

Raised into the caller on the last iteration. Spans exported: 0. With `use_legacy_attributes=True` the same call streams cleanly and exports one span.

### 🤖 Python Version

3.12

An empty stream needs a guard too, since `last_chunk` stays `None` there.

Separately, the missing `@dont_throw` and `span.end()` in a `finally` on these builders is worth addressing, but I have left that out to keep the change focused.

chiruu12:
Both streaming builders passed the generator to `emit_choice_events` after the `for item in response` loop had already drained it. `emit_choice_events` reads `response.candidates`, which a generator does not have, so with events enabled every streamed Gemini call raised `AttributeError` out of the generator into the caller, and the span was never ended.

The non-events branch three lines below already used `last_chunk`. This makes the events branch do the same, and skips emitting when the stream produced no chunks at all.

Neither builder is decorated `@dont_throw` and neither ends the span in a `finally`, which is why this surfaced in user code rather than a log line. Worth fixing, but out of scope here.

Tests in `tests/test_streaming_events.py` cover the sync builder, the async builder and the empty stream. All three fail on main.

Fixes #4417

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronous and asynchronous streaming event handling by using the final response chunk.
  * Prevented choice events from being emitted when streaming responses contain no chunks.

* **Tests**
  * Added coverage for final-chunk event handling in synchronous and asynchronous streams.
  * Added validation for empty streaming responses.


chiruu12:
@dvirski a workflow approval would help here. GitGuardian is the only check that has run, so build, lint and test have never started and the PR is blocked on checks that do not exist yet.

The change itself is contained: the streaming builders handed the already drained generator to `emit_choice_events`, which reads `response.candidates`, so with events enabled every streamed Gemini call raised `AttributeError` into the caller and the span never ended. It now uses `last_chunk`, the way the branch three lines below already did.

PR #4420, opened 2026-08-13
fix(groq): record token and duration metrics for streaming calls

### LINKED ISSUE #4419: 🐛 Bug Report: Groq streaming calls record no token or duration metrics

Groq Instrumentation

### 📜 Description

```python
elif response:
    if duration_histogram:
```

### 👟 Reproduction steps

```python

    stream=True,
)
for _ in response:
    pass

```

### 👍 Expected behavior

### 🤖 Python Version

3.12

chiruu12:
Fixes #4419

Streaming calls returned into the stream processor before the metric block, and neither `_create_stream_processor` nor `_create_async_stream_processor` took a metric instrument. So `gen_ai.client.token.usage` and `gen_ai.client.operation.duration` were recorded for non-streaming calls only, with no warning, since the span attributes were still set correctly.

Both processors now take `token_histogram`, `duration_histogram` and `start_time`, and record on completion. The usage was already being accumulated off `chunk.x_groq.usage`, it just never reached the histogram. The response model is captured from the chunks, since a streaming response has no single object to read it from.

Two small refactors to avoid duplicating the recording logic:

- `record_token_usage_metrics()` in `span_utils.py`, now used by both paths so they emit the same metric under the same attributes.
- `streaming_metrics_attributes(model)` in `utils.py`. `shared_metrics_attributes(response)` now calls it, so behaviour there is unchanged.

### Tests

New `tests/metrics/`, laid out like the openai package, running against recorded cassettes.

This required moving `tests/traces/conftest.py` to `tests/conftest.py` so `tests/metrics/` inherits the fixtures. The file contents are unchanged, and openai already uses that layout.

On main, without the source change:

```
tests/metrics/test_groq_metrics.py::test_chat_metrics            PASSED
tests/metrics/test_groq_metrics.py::test_chat_streaming_metrics  FAILED
    AssertionError: streaming calls record no token usage metric
    assert 'gen_ai.client.token.usage' in {}
```

With the change, both pass.

Full package suite:

```
pytest tests
127 passed
```

flake8 clean.

## Summary by CodeRabbit

- **New Features**
  - Added token usage metrics for streamed chat responses, including input and output tokens when available.
  - Added streaming operation duration metrics with response-model details.
  - Added consistent model and provider attributes across streaming metrics.
  - Metrics are supported for both synchronous and asynchronous streaming responses.

- **Bug Fixes**
  - Improved token metric handling to safely ignore unavailable or invalid token counts.


chiruu12:
@dvirski could you approve the workflow runs on this one? GitGuardian has run and nothing else has, so the required checks have not started and there is no test signal to review against.

Streaming Groq calls never recorded token or duration metrics because the metric emission sat on the non-streaming path only. This moves it so both paths report.

Issue #4436, opened 2026-08-20
🐛 Bug Report: list-valued association properties crash metrics, and silently drop gen_ai.output.messages


leitneratselerity:
### Which component is this bug for?

OpenAI Instrumentation

### Description

When association properties (baggage set via `set_association_properties()` / OTel context, e.g. by the LangChain/LangGraph instrumentation's `langgraph_triggers`) contain a **list** value, `metrics_common_attributes()` in `traceloop/sdk/tracing/tracing.py` copies that value verbatim into the metrics attribute dict:

```python
association_properties = get_value("association_properties")
if association_properties is not None:
    for key, value in association_properties.items():
        common_attributes[f"{SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES}.{key}"] = value
```

This dict is registered as `get_common_metrics_attributes` for the OpenAI (and Anthropic/LiteLLM) instrumentor and merged into every `Counter.record(...)` call's attributes in `_set_chat_metrics` → `_set_token_counter_metrics` (`opentelemetry-instrumentation-openai`). OTel's metrics SDK requires attribute values to be scalar/hashable, since it uses `frozenset(attributes.items())` to key aggregation. A list value causes an unhandled `TypeError: unhashable type: 'list'`.

Because this happens inside `_handle_response` in `opentelemetry/instrumentation/openai/shared/chat_wrappers.py`, which is wrapped by a single `@dont_throw`, the exception is silently swallowed (only a DEBUG log is emitted) and **every response-side span attribute set after that point is lost** — `gen_ai.response.model`, `gen_ai.usage.*`, and, most importantly, `gen_ai.output.messages`. Only `gen_ai.input.messages` (set earlier, before the response arrives) survives. Users see prompts/inputs in their traces but no completions/outputs, with no visible error unless they specifically enable DEBUG logging for `opentelemetry.instrumentation.openai`.

### 👟 Reproduction steps

1. Install `traceloop-sdk` with LangGraph instrumentation enabled (`opentelemetry-instrumentation-langchain` picks this up automatically) and `opentelemetry-instrumentation-openai`.
2. Run a LangGraph graph with more than one edge/trigger into a node so that LangGraph's callback sets `langgraph_triggers` (a `list[str]`) as an association property/baggage value for that node's span — e.g. a node reachable via `["branch:to:some_node"]`.
[...]

PR #4437, opened 2026-08-20
fix(traceloop-sdk): list-valued association properties crash metrics

leitneratselerity:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metric attribute handling for list- and dictionary-valued association properties by encoding them safely.
  * Ensured equivalent dictionaries produce consistent metric attributes regardless of insertion order.
  * Prevented metric aggregation errors caused by unhashable attribute values.
  * Preserved existing behavior for scalar association properties.

* **Tests**
  * Added coverage for scalar, list, and dictionary association-property values.


leitneratselerity:
Fix for issue https://github.com/traceloop/openllmetry/issues/4436

doronkopit5:
@leitneratselerity can you fix comment https://github.com/traceloop/openllmetry/pull/4437#discussion_r3823372674

leitneratselerity:
@doronkopit5 : done

packages/traceloop-sdk/traceloop/sdk/tracing/tracing.py
```diff
@@ -1221,6 +1222,8 @@ def metrics_common_attributes():
     association_properties = get_value("association_properties")
     if association_properties is not None:
         for key, value in association_properties.items():
+            if isinstance(value, (list, dict)):
+                value = json.dumps(value, sort_keys=True)
```
doronkopit5: Thanks for the fix. I reproduced the crash on OTel SDK 1.43 (`histogram.record` raises `TypeError: unhashable type: 'list'`, and the same for dicts), and fixing it here in the SDK covers every instrumentation that uses `get_common_metrics_attributes`, so the placement looks right to me.

A few suggestions:

1. **`json.dumps` can still raise, so the crash just moves here.** A list or dict holding a value JSON can't encode, like a `UUID`, `datetime`, `bytes` or a nested `set`, raises `TypeError` in the same path, e.g. `Traceloop.set_association_properties({"ids": [uuid4()]})`. `sort_keys=True` also raises on a dict with mixed key types (`{1: "a", "b": 2}`). I'd skip the key in that case instead of using `default=str`, because `str()` on an arbitrary object can render its constructor state, including credentials (the issue #4467 fixed on the LangChain side).
2. **`set` is also unhashable** and still crashes the metrics SDK. With the `try`, adding it to the `isinstance` check is enough: `json.dumps` rejects it and the key is skipped.
3. *(Optional)* **Spans and metrics store the value differently.** The span keeps the list as an OTel array attribute, but the metric gets a JSON string. `tuple(value)` for flat lists would be hashable and would match the span. I'm fine with keeping one JSON rule for simplicity, since a tuple of nested lists would still be unhashable.

```suggestion
            if isinstance(value, (list, set, dict)):
                try:
                    value = json.dumps(value, sort_keys=True)
                except TypeError:
                    # Skip values JSON can't encode (UUID, datetime, bytes, ...)
                    # rather than raising out of the caller's LLM call.
                    continue
```

doronkopit5:
The existing tests look good: they're focused, need no cassettes, and pass locally along with the full SDK suite. It would help to add a few cases that cover the edges from the inline comment:

1. **A value JSON can't encode is skipped, not raised.** This one would have caught the `json.dumps` gap.
2. **A `set` value doesn't crash** (it's unhashable, like `list`).
3. **A dict with mixed key types doesn't crash** (`sort_keys=True` raises on `{1: "a", "b": 2}`).
4. **End-to-end through a real histogram.** `hash()` on one value is a proxy. Recording against an in-memory `MeterProvider` checks the actual failure mode (`frozenset(attributes.items())` in the SDK's aggregation).
5. *(Nit)* Detach what the tests attach. I checked that nothing leaks into later tests today, but a token and `detach` keeps the tests independent of how pytest handles context.

Something like:

```python
import uuid

import pytest
from opentelemetry.context import attach, detach, set_value
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import InMemoryMetricReader

PREFIX = SpanAttributes.TRACELOOP_ASSOCIATION_PROPERTIES


@pytest.fixture
def association_properties():
    tokens = []

    def _set(props):
        tokens.append(attach(set_value("association_properties", props)))

    yield _set
    for token in reversed(tokens):
        detach(token)


def test_unencodable_value_is_skipped(association_properties):
    association_properties({"ids": [uuid.uuid4()], "user_id": 1})

    attributes = metrics_common_attributes()

    assert f"{PREFIX}.ids" not in attributes
    assert attributes[f"{PREFIX}.user_id"] == 1


def test_set_value_does_not_crash(association_properties):
    association_properties({"tags": {"a"}})

    attributes = metrics_common_attributes()

    assert f"{PREFIX}.tags" not in attributes


def test_dict_with_mixed_key_types_does_not_crash(association_properties):
    association_properties({"metadata": {1: "a", "b": 2}})

    attributes = metrics_common_attributes()

    assert f"{PREFIX}.metadata" not in attributes


def test_attributes_can_be_recorded_on_a_histogram(association_properties):
    association_properties(
        {"langgraph_triggers": ["branch:to:x"], "metadata": {"a": 1}, "user_id": 1}
    )
    histogram = MeterProvider(metric_readers=[reader]).get_meter("test").create_histogram("h")

    histogram.record(1, metrics_common_attributes())  # raised TypeError before the fix

    assert reader.get_metrics_data() is not None
```

PR #4439, opened 2026-08-22
 record token usage and duration metrics for streaming responses

### LINKED ISSUE #4419: 🐛 Bug Report: Groq streaming calls record no token or duration metrics

chengwudi1:
## What                                                                                                               
  Fixes #4419 — streaming Groq calls recorded no `gen_ai.client.token.usage` or `gen_ai.client.operation.duration`      
  metrics. The metric block in `_wrap`/`_awrap` was only reached for non-streaming responses.                           
                                                                                                                        
  ## How                                                                                                                
  - `_create_stream_processor` / `_create_async_stream_processor` now receive `token_histogram`, `duration_histogram`,  
  `start_time`, and `llm_model`.                                                                                        
  - A shared `_record_streaming_metrics` helper records duration (always) and token usage (input/output, from           
  `chunk.x_groq.usage` on the final chunk) once the stream is fully drained.                                            
  - Added `streaming_metrics_attributes` in `utils.py` for streaming metric attributes (no full response object exists  
  in the streaming path).                                                                                               
  - Moved `tests/traces/conftest.py` up to `tests/conftest.py` so the new `tests/metrics/` suite can reuse the existing 
  fixtures (as suggested in the issue).                                                                                 
                                                                                                                        
  ## Tests                                                                                                              
  Added `tests/metrics/test_groq_streaming_metrics.py` (5 tests, mock-based — no network/cassettes):                    
  - sync streaming records token + duration metrics                                                                     
  - duration recorded even when no usage chunk                                                                          
  - metrics disabled (`TRACELOOP_METRICS_ENABLED=false`) skips recording                                                
  - async streaming records token + duration metrics                                                                    
  - `streaming_metrics_attributes` unit test                                                                            
                                                                                                                        
  `130 passed` in the groq package; `ruff check` and `ruff format` clean.                                               
                                                                                                                        
  - [ ] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability     
  platform showing the change.                                                                                          

## Summary by CodeRabbit

* **New Features**
  * Added token-usage metrics for streaming Groq chat completions, including input and output token counts.
  * Added duration metrics for completed synchronous and asynchronous streaming requests.
  * Metrics include the associated model and token type for clearer observability.
  * Duration is recorded even when token-usage data is unavailable or a stream encounters an error.

* **Tests**
  * Added coverage for synchronous and asynchronous streaming metrics, disabled metrics, missing usage data, and stream errors.


chengwudi1:
Hi maintainers! Just a friendly ping — the CI workflows on this PR have been
  awaiting approval to run (first-time contributor). Could you approve them when
  you get a chance? Happy to make any changes. Thanks!

packages/opentelemetry-instrumentation-groq/opentelemetry/instrumentation/groq/__init__.py
```diff
@@ -218,13 +265,22 @@ def _create_stream_processor(response, span, event_logger):
         _handle_streaming_response(
             span, accumulated_content, tool_calls, accumulated_finish_reasons, usage, event_logger
         )
+        _record_streaming_metrics(usage, token_histogram, duration_histogram, start_time, llm_model)
```
RichardoMrMu, line 273: **[P2] Record duration when stream iteration fails.** This call is only reached from the try block's `else`, so if Groq yields one or more chunks and then raises, the exception path marks the span as failed and re-raises but emits no operation-duration data point. Streaming failures therefore disappear from the duration metric, unlike failures before stream creation in `_wrap` / `_awrap`, which record duration with `error_metrics_attributes(e)`. A regression using a generator that yields once and then raises `RuntimeError` currently leaves the duration metric absent. Please record duration with error attributes in both the sync and async exception paths before re-raising.

chengwudi1:
Hi @RichardoMrMu! Friendly follow-up on this one. Your [P2] comment from Aug 29 (record duration when stream iteration fails) is addressed in `2f712bd`: when Groq yields chunks and then raises, the exception path now emits the operation-duration data point with `error_metrics_attributes(e)` for both sync and async paths — matching the pre-stream failure behavior, with 3 new regression tests (133 passed, ruff clean). Also, if the CI workflows on this PR still need approval to run, approving would help. Thanks!

PR #4445, opened 2026-08-29
fix(groq): record token and duration metrics for streaming completions (#4419)

### LINKED ISSUE #4419: 🐛 Bug Report: Groq streaming calls record no token or duration metrics

vaibhavmashal:
### Description
Fixes #4419

Streaming chat completions in \opentelemetry-instrumentation-groq\ were creating trace spans and capturing token usage on spans, but they were not recording OpenTelemetry metric histograms (\gen_ai.client.token.usage\ and \gen_ai.client.operation.duration\).

### Changes
- Updated \_create_stream_processor\ and \_create_async_stream_processor\ to accept \start_time\, \duration_histogram\, \	oken_histogram\, and \llm_model\ (with chunk model extraction).
- Added duration histogram metric recording on stream completion and on stream error.
- Updated \set_model_streaming_response_attributes\ to record input and output token usage on \	oken_histogram\ when \chunk.x_groq.usage\ is provided.
- Added comprehensive unit tests in \TestCreateStreamProcessor\, \TestCreateAsyncStreamProcessor\, and \TestSetModelStreamingResponseAttributes\ covering sync and async stream completion, error duration recording, and token metrics.

### Testing
- [x] Verified all 121 unit tests pass in \packages/opentelemetry-instrumentation-groq\.

## Summary by CodeRabbit

* **New Features**
  * Streaming responses now report operation duration and input/output token metrics.
  * Telemetry distinguishes the requested model from the model used in streamed responses.
  * Successful and failed streaming operations include relevant duration, token, model, and status details.
  * Token metrics are recorded even when spans are not actively recording.

* **Bug Fixes**
  * Improved handling of optional usage data during streaming.
  * Corrected cache-read token attribute reporting.
  * Improved error handling when streaming operations fail.


packages/opentelemetry-instrumentation-groq/opentelemetry/instrumentation/groq/span_utils.py
```diff
@@ -202,6 +218,30 @@ def set_model_streaming_response_attributes(span, usage, finish_reasons=None):
                     cached_tokens,
                 )
 
+        if isinstance(prompt_tokens, int) and prompt_tokens >= 0 and token_histogram is not None:
```
RichardoMrMu, line 200: [P1] Decouple token metrics from span sampling

set_model_streaming_response_attributes returns before either token_histogram.record call whenever span.is_recording() is false. With a DROP/ParentBased sampler, the processor still records the duration histogram, but every successful streaming call silently loses both input- and output-token metrics. I reproduced this on head f76c27a49cd627c1bf76869a3611b6f6652594f5 with a non-recording span and 10/20-token usage: expected 2 histogram records, observed 0. Metrics should remain independent of the trace sampling decision; please move the histogram work outside the span-recording guard (or split it into a metrics-only helper) and add a non-recording-span regression for both sync and async streaming paths.

vaibhavmashal: Good catch @RichardoMrMu! I've updated \set_model_streaming_response_attributes\ so token histogram recording occurs before the \span.is_recording()\ check, ensuring input and output token metrics are recorded regardless of whether the span is sampled/recording. I also added non-recording span regression tests for both sync and async streaming paths in \	est_init.py\.

Pushed in commit \88d6c91\ — all 88 unit tests passing cleanly.

PR #4455, opened 2026-09-02
fix(crewai): wrap native provider LLM classes missed by the LLM factory

### LINKED ISSUE #4453: 🐛 Bug Report: CrewAI: LLM.call wrap never fires on crewai >= 1.x native providers, so no LLM span is emitted

All Packages

### 📜 Description

Component: **CrewAI Instrumentation** — not in the dropdown, so "All Packages."

`_instrument` wraps `crewai.llm` `LLM.call` ([instrumentation.py#L93-L94](https://github.com/traceloop/openllmetry/blob/62e24c2ffde6c1ee04dc290e52d8d5dbda054cff/packages/opentelemetry-instrumentation-crewai/opentelemetry/instrumentation/crewai/instrumentation.py#L93-L94)), but on crewai 1.15.18 that wrap can never be reached on the native-provider path, so `wrap_llm_call` never runs — no `{model}.llm` span, and the duration histogram in `_record_duration` never records.

`crewai.llm.LLM.__new__` is a factory that routes to native provider classes rather than returning an `LLM`:

    def __new__(cls, model: str, is_litellm: bool = False, **kwargs) -> LLM:
        """Factory method that routes to native SDK or falls back to LiteLLM."""

The resulting object is not in `LLM`'s hierarchy at all:

    >>> a = Agent(role='r', goal='g', backstory='b')
    >>> type(a.llm)
    crewai.llms.providers.openai.completion.OpenAICompletion
    >>> isinstance(a.llm, crewai.llm.LLM)
    False
    >>> [c.__name__ for c in OpenAICompletion.__mro__]
    ['OpenAICompletion', 'BaseLLM', 'BaseModel', 'ABC', 'object']

So patching `crewai.llm.LLM.call` patches a class the agent never instantiates.

### 👟 Reproduction steps

1. `pip install crewai==1.15.18 traceloop-sdk==0.62.3 opentelemetry-instrumentation-crewai==0.62.3`
2. `Traceloop.init(...)`, build a crew with a default `Agent` (no explicit LLM), run `Crew.kickoff()`.
3. Inventory the exported span names.

### 👍 Expected behavior

A `{model}.llm` span per LLM call, as `wrap_llm_call` is written to produce, plus the `gen_ai.client.operation.duration` recording.

No `{model}.llm` span. Full inventory from a real run:

```
   3  openai.chat
   1  Support Responder.agent
   1  <task description>.task
   1  crewai.workflow
```

### 🤖 Python Version

3.12.3

Scope: I verified the native OpenAI provider route only. `LLM.__new__` also has a LiteLLM fallback (`is_litellm=True`), and I haven't checked whether that path yields a real `crewai.llm.LLM` instance where the existing wrap would still fire. So this may be "dead for native providers" rather than "dead everywhere" — worth confirming on your side.

On the fix: `BaseLLM` is the common ancestor and does define `call`, but `OpenAICompletion` overrides it, so wrapping `BaseLLM.call` wouldn't intercept either. The right seam is probably the provider classes, or whatever single funnel exists upstream of them — you'll know that better than I do. Happy to send a PR once you've picked one.

Related: #4452  covers the missing tool-execution span in the same package. (The one I previously opened)

Huang-404-Q:
## Summary

On crewai >= 1.x, the CrewAI instrumentation produces **no LLM spans and no duration metrics** for any model on a native provider path (OpenAI, Azure, Anthropic, Gemini, Bedrock) — the wrap target is a class that is never instantiated there.

### Root cause

`_instrument` wraps `crewai.llm.LLM.call` (`instrumentation.py:93-94` on main). But crewai 1.x's `LLM.__new__` is a factory that routes to native provider classes:

```python
def __new__(cls, model: str, is_litellm: bool = False, **kwargs: Any) -> LLM:
    """Factory method that routes to native SDK or falls back to LiteLLM."""
```

The resulting object is not an `LLM` at all:

```text
>>> from crewai.llms.providers.openai.completion import OpenAICompletion
>>> [c.__name__ for c in OpenAICompletion.__mro__]
['OpenAICompletion', 'BaseLLM', 'BaseModel', 'ABC', 'object']
>>> isinstance(OpenAICompletion(model="gpt-4o-mini", api_key="sk-test"), crewai.llm.LLM)
False
```

Wrapping the shared base `BaseLLM.call` does not work either: every native provider overrides `call`, and `wrap_function_wrapper` patches the class attribute slot the override shadows.

### Fix

Wrap each native provider class's `call` alongside `LLM.call`:

- `crewai.llms.providers.openai.completion.OpenAICompletion` (also covers `OpenAICompatibleCompletion`, which inherits it without overriding)
- `crewai.llms.providers.azure.completion.AzureCompletion`
- `crewai.llms.providers.anthropic.completion.AnthropicCompletion`
- `crewai.llms.providers.gemini.completion.GeminiCompletion`
- `crewai.llms.providers.bedrock.completion.BedrockCompletion`

`wrap_llm_call` works unchanged for all of them: `model` is a `BaseLLM` field, `messages` is the first positional argument of every `call` signature, and `last_token_usage` access is already `hasattr`-guarded. Provider modules import their SDK lazily and may be absent, so each wrap is best-effort (`ImportError`/`AttributeError` → debug log); `_uninstrument` mirrors the same list.

## Verification

- Structural regression tests: `OpenAICompletion.call` carries the wrapt wrapper after `instrument()` and is restored after `uninstrument()`.
- End-to-end against a real `OpenAICompletion` with a mocked OpenAI client — before this change: zero spans; after:

```text
spans: ['gpt-4o-mini.llm']
LLM span: gpt-4o-mini.llm status: OK
gen_ai.request.model: gpt-4o-mini
gen_ai.provider.name: openai
```

- Full package suite: 175 passed.

### Test infrastructure note

The new `tests/conftest.py` handles a trap that likely let this bug persist: `BaseInstrumentor` is a singleton (its `__new__` returns a shared instance), so the existing `mock_instrumentor` fixture leaves `instrument`/`uninstrument` mocked on that shared instance for every later test. The real-instrumentation fixture now removes those mocks and re-syncs the instrumented flag before use.

Fixes #4453

## Summary by CodeRabbit

* **New Features**
  * CrewAI LLM telemetry now reports explicit provider identities and model details.
  * LLM operation duration metrics include the associated provider, with model-based detection retained as a fallback.

* **Bug Fixes**
  * Uninstrumentation now reliably restores native provider integrations and existing CrewAI LLM instrumentation.

* **Tests**
  * Added coverage for provider identification, telemetry attributes, metrics, and instrumentation restoration.


Huang-404-Q:
Review addressed in d37a056: native provider wraps now carry their authoritative OTel provider value at wrap time — `AzureCompletion` spans report `gen_ai.system = azure` (normalized to `GenAiSystemValues.AZURE_AI_OPENAI`) even for `gpt-4` model names, and the duration metric attributes resolve from the same value. The LiteLLM fallback (`crewai.llm.LLM.call`) keeps model-name inference, including the never-guess behavior for unknown models. Regression test added for inference retention on the fallback path.

Huang-404-Q:
Round-3 addressed in cb8879d: the test credential now comes from `OPENAI_API_KEY` (no API-key-shaped literal), and the duration data point assertion checks `gen_ai.provider.name == 'openai'` so provider-label regressions fail the test.

PR #4465, opened 2026-09-07
fix(crewai): stop the __dict__ walk leaking credentials onto spans

IdoGol24:
`CrewAISpanAttributes `stringified every attribute of the Crew, Agent and Task objects it was handed, so anything those objects held -- an LLM client, an embedder config -- reached the span through its repr, api keys included.

Read a fixed set of fields instead. `Agents `already had a hand-written allowlist in _extract_agent_data; the standalone-agent path just wasn't using it. Tasks reuse the dict _parse_tasks already built. Only Crew needed a new list. Tools go through the existing `_serialize_tools `on every path rather than being repr'd, and the dead `_populate_llm_attributes` is gone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry safety by preventing credentials and other sensitive configuration values from appearing in span attributes.
  * Standardized captured task and agent details, including task identifiers and supported fields.
  * Represented tool information as structured JSON arrays instead of re-encoded strings or Python representations.
  * Continued emitting supported Crew, task, and agent identifiers while excluding unsupported configuration data.
  * Improved consistency of telemetry attributes across CrewAI components.

### DIFF
--- a/packages/opentelemetry-instrumentation-crewai/opentelemetry/instrumentation/crewai/crewai_span_attributes.py
+++ b/packages/opentelemetry-instrumentation-crewai/opentelemetry/instrumentation/crewai/crewai_span_attributes.py
@@ -102,6 +102,7 @@ class CrewAISpanAttributes:
                 self._set_attribute(SpanAttributes.GEN_AI_REQUEST_MAX_COMPLETION_TOKENS, value)
 
     def _populate_crew_attributes(self):
+        """Collect the allowlisted Crew fields, plus its tasks and agents."""
         for key in CREW_FIELDS:
             value = getattr(self.instance, key, None)
             if value is not None:
@@ -110,13 +111,16 @@ class CrewAISpanAttributes:
         self._parse_agents(self.instance.agents or [])
 
     def _populate_agent_attributes(self):
+        """Collect the allowlisted Agent fields for a standalone agent span."""
         return self._stringify(self._extract_agent_data(self.instance))
 
     def _populate_task_attributes(self):
+        """Collect the allowlisted Task fields for a standalone task span."""
         return self._stringify(self._extract_task_data(self.instance))
 
     @staticmethod
     def _stringify(data):
+        """Render a field dict as span-attribute values, dropping the unset ones."""
         return {key: str(value) for key, value in data.items() if value is not None}
 
     def _parse_agents(self, agents):
@@ -125,9 +129,11 @@ class CrewAISpanAttributes:
         ]
 
     def _parse_tasks(self, tasks):
+        """Attach the crew's tasks, each reduced to its allowlisted fields."""
         self.crew["tasks"] = [self._extract_task_data(task) for task in tasks if task is not None]
 
     def _extract_task_data(self, task):
+        """Return the allowlisted Task fields, with the agent named by its role."""
         return {
             "id": str(task.id),
             "agent": task.agent.role if task.agent else None,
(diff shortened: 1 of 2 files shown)


packages/opentelemetry-instrumentation-crewai/opentelemetry/instrumentation/crewai/crewai_span_attributes.py
```diff
@@ -151,24 +159,12 @@ def _extract_agent_data(self, agent):
             "goal": agent.goal,
             "backstory": agent.backstory,
             "cache": agent.cache,
-            "config": agent.config,
             "verbose": agent.verbose,
             "allow_delegation": agent.allow_delegation,
-            "tools": agent.tools,
+            "tools": self._serialize_tools(agent.tools or []),
```
doronkopit5: `_serialize_tools` returns a JSON *string*. That's correct for the standalone Agent/Task spans, where `_stringify` stringifies every field anyway — but the same dicts also feed the Crew path via `_parse_agents`/`_parse_tasks`, and `_process_crew` then runs `json.dumps(value, default=str)` over the whole list, re-escaping the already-encoded `tools`.

Running both revisions against real `crewai` objects:

```
main:  json.loads(attrs["crewai.crew.agents"])[0]["tools"]  ->  list
here:  json.loads(attrs["crewai.crew.agents"])[0]["tools"]  ->  str   '[{"name": "echo", ...}]'
```

So inside `crewai.crew.agents` / `crewai.crew.tasks`, `tools` becomes the one field that needs a second `json.loads()`, while every sibling stays plain. On `main` that nested value was a real array (of full object reprs — which this PR is right to sanitize), so it's a shape change introduced here rather than a pre-existing wart.

Suggestion: let the extractors hand back native lists and encode once at the span boundary. Not a one-click suggestion because it spans three spots, but each edit is small:

```python
# 1. _serialize_tools -> return plain dicts
    def _tool_dicts(self, tools):
        """Return the allowlisted fields of each tool, as plain dicts."""
        return [
            {k: v for k, v in vars(tool).items() if v is not None and k in ["name", "description"]}
            for tool in tools
        ]

# 2. both extractors use it
            "tools": self._tool_dicts(task.tools or []),
            "tools": self._tool_dicts(agent.tools or []),

# 3. _stringify encodes structured values as JSON instead of repr-ing them
    @staticmethod
    def _stringify(data):
        """Render a field dict as span-attribute values, dropping the unset ones."""
        return {
            key: json.dumps(value) if isinstance(value, (list, dict)) else str(value)
            for key, value in data.items()
            if value is not None
        }
```

Step 3 matters: without it `_stringify` would `str()` the list and emit a Python repr (single quotes) instead of JSON.

I verified this patch against real `Agent`/`Task`/`Crew` objects — nested crew `tools` comes back as a real array, `crewai.agent.tools` and `crewai.task.tools` keep exactly the JSON-array string they produce today, and the new allowlist test still passes with no sentinel leaking.

IdoGol24:
Hi @doronkopit5 thank you for your reply!
I have Fixed in 11b7368c.
` _serialize_tools `→ `_tool_dicts` returning plain dicts, both extractors updated, and `_stringify `now JSON-encodes lists/dicts at the span boundary. Added two regression tests: nested `crewai.crew.agents`/`tasks` tools decode in one pass (these fail on the previous revision), and `crewai.agent.tools`/`crewai.task.tools` pass on both revisions, confirming their shape is unchanged.

doronkopit5 (approved):

PR #4466, opened 2026-09-09
fix(mcp): honor TRACELOOP_TRACE_CONTENT on the client path

IdoGol24:
The package documents TRACELOOP_TRACE_CONTENT as the switch that turns off content logging, but only the FastMCP server-side wrapper consulted it. The MCP client path recorded content regardless: tools/call arguments and results via _extract_clean_input/_extract_clean_output, whole request and response bodies via serialize() in _handle_mcp_method and _execute_and_handle_result, and the response value in InstrumentedStreamWriter.send. An operator who set the variable to false still got request and response payloads on their spans.

Move should_send_prompts() into utils so one gate serves both wrappers, and apply it to every content-bearing attribute on the client path. Span names, entity names, span kind, request ids and error status are unaffected; only content is withheld.

Tests drive the real client against a FastMCP server and assert a marker value is absent from every span attribute when the switch is off and still present when it is on, so neither the gate nor the capture can regress unnoticed.

## Summary by CodeRabbit

- **New Features**
  - Added centralized control for capturing MCP request, response, and error content in telemetry.
  - Content capture is enabled by default and can be disabled with the `TRACELOOP_TRACE_CONTENT` setting.

- **Bug Fixes**
  - MCP client and server instrumentation now consistently respect the content-capture setting.
  - When disabled, sensitive messages and source details are omitted while error types, statuses, and sanitized stack frames remain available.
  - Error reporting is now consistent across MCP operations, including session teardown and send failures.

### DIFF
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py
@@ -16,6 +16,7 @@ class FastMCPInstrumentor:
     """Handles FastMCP-specific instrumentation logic."""
 
     def __init__(self):
+        """Create the instrumentor with no tracer or server name bound yet."""
         self._tracer = None
         self._server_name = None
 
@@ -50,6 +51,7 @@ class FastMCPInstrumentor:
         @dont_throw
         def traced_method(wrapped, instance, args, kwargs):
             # Call the original __init__ first
+            """Record the server name from FastMCP's constructor arguments."""
             result = wrapped(*args, **kwargs)
 
             if args and len(args) > 0:
@@ -63,6 +65,7 @@ class FastMCPInstrumentor:
     def _fastmcp_tool_wrapper(self):
         """Create wrapper for FastMCP tool execution."""
         async def traced_method(wrapped, instance, args, kwargs):
+            """Wrap a FastMCP tool call in server and tool spans."""
             if not self._tracer:
                 return await wrapped(*args, **kwargs)
 
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py
@@ -9,7 +9,7 @@ from opentelemetry.semconv_ai import SpanAttributes, TraceloopSpanKindValues
 from opentelemetry.semconv.attributes.error_attributes import ERROR_TYPE
 from wrapt import register_post_import_hook, wrap_function_wrapper
 
-from .utils import dont_throw
+from .utils import dont_throw, should_send_prompts
 
 
 class FastMCPInstrumentor:
@@ -155,9 +155,7 @@ class FastMCPInstrumentor:
 
     def _should_send_prompts(self):
         """Check if content tracing is enabled (matches traceloop SDK)"""
-        return (
-            os.getenv("TRACELOOP_TRACE_CONTENT") or "true"
-        ).lower() == "true"
+        return should_send_prompts()
 
     def _get_json_encoder(self):
         """Get JSON encoder class (simplified - traceloop SDK uses custom JSONEncoder)"""
--- a/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
+++ b/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
@@ -28,15 +28,19 @@ _instruments = ("mcp >= 1.6.0",)
 
 
 class McpInstrumentor(BaseInstrumentor):
+    """Instrument the MCP client, server and transports with OpenTelemetry spans."""
     def __init__(self, exception_logger=None):
+        """Store the exception logger and build the FastMCP sub-instrumentor."""
         super().__init__()
         Config.exception_logger = exception_logger
         self._fastmcp_instrumentor = FastMCPInstrumentor()
 
     def instrumentation_dependencies(self) -> Collection[str]:
+        """Return the package versions this instrumentation supports."""
         return _instruments
 
     def _instrument(self, **kwargs):
+        """Wrap the MCP client, server sessions and every supported transport."""
         tracer_provider = kwargs.get("tracer_provider")
         tracer = get_tracer(__name__, __version__, tracer_provider)
 
@@ -118,11 +122,13 @@ class McpInstrumentor(BaseInstrumentor):
         )
 
     def _uninstrument(self, **kwargs):
+        """Unwrap the transports this instrumentation replaced."""
         unwrap("mcp.client.stdio", "stdio_client")
         unwrap("mcp.server.stdio", "stdio_server")
         self._fastmcp_instrumentor.uninstrument()
 
     def _transport_wrapper(self, tracer):
+        """Wrap a transport so its read and write streams are instrumented."""
         @asynccontextmanager
         async def traced_method(
             wrapped: Callable[..., Any], instance: Any, args: Any, kwargs: Any
@@ -133,6 +139,7 @@ class McpInstrumentor(BaseInstrumentor):
             ],
             None,
         ]:
+            """Yield the transport's streams wrapped in instrumented proxies."""
             async with wrapped(*args, **kwargs) as result:
                 try:
                     read_stream, write_stream = result
@@ -161,9 +168,11 @@ class McpInstrumentor(BaseInstrumentor):
         return traced_method
 
     def _base_session_init_wrapper(self, tracer):
+        """Wrap a server session's incoming message streams to carry trace context."""
         def traced_method(
             wrapped: Callable[..., None], instance: Any, args: Any, kwargs: Any
         ) -> None:
+            """Replace the session's incoming stream pair with context-propagating proxies."""
             wrapped(*args, **kwargs)
             reader = getattr(instance, "_incoming_message_stream_reader", None)
             writer = getattr(instance, "_incoming_message_stream_writer", None)
@@ -182,8 +191,10 @@ class McpInstrumentor(BaseInstrumentor):
         return traced_method
 
     def patch_mcp_client(self, tracer: Tracer):
+        """Wrap BaseSession.send_request so each MCP request becomes a span."""
         @dont_throw
         async def traced_method(wrapped, instance, args, kwargs):
+            """Start a span for the outgoing request and inject trace context into its meta."""
             meta = None
             method = None
             params = None
@@ -220,6 +231,7 @@ class McpInstrumentor(BaseInstrumentor):
         @dont_throw
         async def traced_method(wrapped, instance, args, kwargs):
             # Start a root span for the MCP client session and make it current
+            """Wrap a FastMCP client session enter to open a session span."""
             span_context_manager = tracer.start_as_current_span("mcp.client.session")
             span = span_context_manager.__enter__()
             span.set_attribute(SpanAttributes.TRACELOOP_SPAN_KIND, "session")
@@ -247,6 +259,7 @@ class McpInstrumentor(BaseInstrumentor):
 
         @dont_throw
         async def traced_method(wrapped, instance, args, kwargs):
+            """Close the session span when the FastMCP client exits."""
             try:
                 # Call the original method first
                 result = await wrapped(*args, **kwargs)
@@ -472,6 +485,7 @@ def serialize(request, depth=0, max_depth=4):
     depth += 1
 
     def is_serializable(request):
+        """Return whether a value can be JSON-encoded without a fallback."""
         try:
             json.dumps(request)
             return True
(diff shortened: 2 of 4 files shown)


packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py
```diff
@@ -9,13 +9,14 @@
...
 
-from .utils import dont_throw
+from .utils import dont_throw, should_send_prompts
 
 
 class FastMCPInstrumentor:
     """Handles FastMCP-specific instrumentation logic."""
 
     def __init__(self):
+        """Create the instrumentor with no tracer or server name bound yet."""
```
doronkopit5: remove all these unrelated docstrings

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -321,8 +346,11 @@ async def _execute_and_handle_result(
         """Execute the wrapped function and handle the result"""
         try:
             result = await wrapped(*args, **kwargs)
-            # Add output
-            if clean_output:
+            # Add output. The response body is content, so it is recorded only
+            # when content capture is enabled.
+            if not should_send_prompts():
```
doronkopit5, line 357: The gate stops here, but ~20 lines down we still do `span.set_status(Status(StatusCode.ERROR, f"{result.content[0].text}"))` for `isError` results, and that text is the tool's response body.

I tried it: with `TRACELOOP_TRACE_CONTENT=false`, a tool raising `ToolError(f"failure involving {token}")` gave a `boom.tool` span whose status description was `failure involving <my marker>` — the caller's own argument round-tripping back out. Status descriptions are exported like attributes, so every failed tool call still carries content with the switch off.

Either gate that one too, or set a fixed description (`"tool error"`) when capture is off and keep the error code.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -565,9 +604,13 @@ async def send(self, item: Any) -> Any:
 
         with self._tracer.start_as_current_span("ResponseStreamWriter") as span:
             if hasattr(request, "result"):
-                span.set_attribute(
-                    SpanAttributes.MCP_RESPONSE_VALUE, f"{serialize(request.result)}"
-                )
+                # The response body is content; the error status below is not,
```
doronkopit5: This assumption doesn't hold: the status *code* isn't content, but the *description* is. Right below we set `Status(StatusCode.ERROR, f"{request.result['content'][0]['text']}")` — the same payload text withheld two lines up.

This is the proxy used for stdio/SSE, and the new tests go through the in-memory FastMCP client, so nothing covers this path.

packages/opentelemetry-instrumentation-mcp/tests/test_content_capture_gate.py
```diff
@@ -0,0 +1,124 @@
...
+from fastmcp import Client, FastMCP
+
+MARKER = "content-capture-marker-9f3a"
+
+
+def _all_attribute_text(span_exporter) -> str:
+    """Every attribute value across every exported span, as one string."""
+    chunks = []
+    for span in span_exporter.get_finished_spans():
+        for value in (span.attributes or {}).values():
```
doronkopit5: This only walks `span.attributes` — never `span.status.description` or `span.events`. The module docstring says the test fails "if either the gate or the capture itself regresses", but the leaks I flagged in `instrumentation.py` all pass it.

If you fold status descriptions and event attributes in here, these tests fail as they stand today, which is sort of the point.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
doronkopit5, line 1: (File-level since the line is outside the diff.) One more place the gate doesn't reach: `span.record_exception(e)` in `_execute_and_handle_result` writes `exception.message` and the full `exception.stacktrace` as event attributes. On the client path the `McpError` message comes from the server, so it is content. `fastmcp_instrumentation.py` around line 114 does the same.

With the switch off I still got `exception.message='failure involving <marker>'` and the marker inside the stacktrace, on both `boom.tool` and `mcp.server`.

Fine by me if exception text is deliberately out of scope, but then let's note it in the PR description — as written the claim is that content is withheld.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -216,7 +221,11 @@ def _fastmcp_client_enter_wrapper(self, tracer):
         @dont_throw
         async def traced_method(wrapped, instance, args, kwargs):
             # Start a root span for the MCP client session and make it current
-            span_context_manager = tracer.start_as_current_span("mcp.client.session")
+            span_context_manager = tracer.start_as_current_span(
+                "mcp.client.session",
+                record_exception=False,
```
doronkopit5, line 227: The reasoning in the comment at 295-297 holds for the other spans, but not for this one: nothing reports a failure on `mcp.client.session`. The exit wrapper at 268-274 only forwards to `context_manager.__exit__(type(e), e, e.__traceback__)` and re-raises into `@dont_throw`, which swallows it and logs at DEBUG. With these two flags off that `__exit__` is a no-op, so a teardown failure now leaves nothing behind.

Base vs. head, with a `Client.__aexit__` that raises:

```
base: mcp.client.session status=ERROR desc='RuntimeError: teardown exploded' events=[exception]
head: mcp.client.session status=UNSET desc=None      events=[]
```

This happens with `TRACELOOP_TRACE_CONTENT` at its default, so it isn't a content-gating tradeoff. Stdio/SSE transports raising on teardown (`Attempted to exit cancel scope in a different task`, `BrokenResourceError`, `ClosedResourceError`) is the case you'd most want visible on the session span.

Either keep the defaults on this span, or give the exit wrapper's `except` the same three lines the enter wrapper got at 243-246. Nothing in the suite covers this path either way.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
```diff
@@ -343,16 +371,18 @@ async def _execute_and_handle_result(
...
             return result
         except Exception as e:
             span.set_attribute(ERROR_TYPE, type(e).__name__)
-            span.record_exception(e)
-            span.set_status(Status(StatusCode.ERROR, str(e)))
+            # record_exception writes the message and the full stacktrace as
+            # event attributes, and on the client path that text comes from the
+            # server, so it is content.
+            if should_send_prompts():
+                span.record_exception(e)
```
doronkopit5: The message is content, agreed — but `exception.type` and `exception.stacktrace` aren't, and dropping the event takes them along. With the switch off there's now no stack frame recorded for any MCP failure. `error.type` keeps the class name, so it's not total, but the stacktrace is the part you actually debug from.

An `else` that adds the event with just the type and the formatted traceback would withhold `exception.message` and nothing more. Same at `fastmcp_instrumentation.py:127-128`.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
IdoGol24: Confirmed, I took the second option rather than restoring the defaults, so this span follows the same rule as the other five instead of being the one exception.

The exit wrapper couldn't call record_error as-is: it only had the context manager, not the span. The enter wrapper now also stores the span on the instance (`_tracing_session_span`), and the exit wrapper's except records on it before `__exit__ `ends it.

You were right that nothing covered this path. `test_session_teardown_failure_is_recorded `drives the enter and exit wrappers directly - `Client.__aexit__` is already wrapped, so monkeypatching it would replace the wrapper under test. Parameterized off/on: status is ERROR either way, the message only when content capture is on. It fails against the previous head.

Side effect worth noting: `error.type` now appears on this span and on `ResponseStreamWriter`, neither of which carried it on main.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
IdoGol24: Agreed!
I've fixed via a shared record_error(span, exc) that every error path now calls, including `fastmcp_instrumentation.py.` Type and frames recorded either way, message withheld when the switch is off.

One correction to the suggestion: 
Formatted traceback does not withhold the message and nothing more. It reappears in two places.

format_exc         leaks message: True   # last line is "Type: message", plus chained causes
format_exception   leaks message: True
format_tb          leaks message: True   # not the last line — the source line of the raise frame

format_tb drops the final line but still renders each frame's source line, so a tool raising ToolError("...") prints its own message there. So the frames are rendered by hand (file, line, function, no source line). Runtime data was never at risk either way.

That second part took a specific test to catch: my first version raised `ValueError`, which fastmcp re-wraps, so the original frame never survives and the test passed under both implementations. `test_source_literals_do_not_leak_through_the_stacktrace `uses `ToolError`, which the tool manager doesn't re-wrap, and fails under format_tb.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
IdoGol24: Fixed. Took the "gate it" option over a fixed "tool error" string: 
`error.type` already names the failure class, so a constant description gives a consumer nothing extra to key on while looking like real text. StatusCode.ERROR is preserved.

Also worth flagging since it came out of this: the sweep in the test helper only walked span.attributes, so none of these leaks could have been caught. It now folds in status descriptions and event attributes, which is what turned your three reports into failing tests

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py
IdoGol24: It ran deeper than the two explicit record_exception calls. `start_as_current_span `defaults to `record_exception=True`, `set_status_on_exception=True,` so as the exception left each with block OTel re-recorded the message and stacktrace ungated and overwrote the gated status, which is also why every failure was carrying two identical exception events. 

Both flags are now off at the five spans whose code reports failures itself.

Your follow-up on this thread is in too: dropping the event entirely also cost `exception.type` and `exception.stacktrace`, so record_error now keeps both and withholds only the message. Your exact case is what `test_error_text_suppressed_when_content_capture_off ` asserts against, and it fails without the fix.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/utils.py
```diff
@@ -2,13 +2,71 @@
...
+    span.set_attribute(ERROR_TYPE, type(exc).__name__)
+    if should_send_prompts():
+        span.record_exception(exc)
+    else:
+        span.add_event(
+            "exception",
+            {
+                "exception.type": f"{type(exc).__module__}.{type(exc).__qualname__}",
+                # File, line and function, but not the frame's source line: a
+                # raise site like ToolError("...") reproduces its own message
```
doronkopit5, line 57: This doesn't match what the SDK's `record_exception` writes, so the event looks different depending on the switch. For a `ValueError`: on gives `exception.type="ValueError"`, off gives `"builtins.ValueError"` (the SDK drops the module for builtins). `exception.escaped` is also a bool here but the string `"False"` from the SDK. Anything grouping on `exception.type` splits in two.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/utils.py
```diff
@@ -2,13 +2,71 @@
...
 
+def should_send_prompts() -> bool:
+    """Whether request/response content may be recorded on spans.
+
+    Mirrors the traceloop SDK's ``TRACELOOP_TRACE_CONTENT`` switch: content
+    capture is on unless an operator explicitly turns it off. Shared by the
+    FastMCP server wrapper and the MCP client path so a single environment
+    variable governs both, which is what the package README documents.
+    """
+    return (os.getenv("TRACELOOP_TRACE_CONTENT") or "true").lower() == "true"
```
doronkopit5, line 24: The other instrumentations also check `context_api.get_value("override_enable_content_tracing")` here (see anthropic/openai `should_send_prompts`). The SDK sets that from the content allow list, so without it MCP ignores per-association content enablement. Can we match them?

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/utils.py
IdoGol24: Fixed in 0add16305. Both branches now go through `span.record_exception`, and the content-off branch only overrides `exception.message` (to `""`) and `exception.stacktrace` (frames only). The SDK writes `exception.type` and `exception.escaped`, so they come out the same either way. The one difference: with content off, `exception.message` is present but empty, because `record_exception` lets a caller replace a key but not drop it.

Added `test_exception_event_shape_is_the_same_either_way`, which raises the same error with the switch on and off and compares type and escaped. It fails on the previous commit.

packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/utils.py
IdoGol24: Yes, matched them in 83d84c4d7. `should_send_prompts()` now reads `(env or "true") == "true" or context_api.get_value("override_enable_content_tracing")`, the same as anthropic and openai, so the override can only turn capture on. Every gate on this branch goes through that one function.

Added `test_content_allow_list_override_enables_capture`: switch off, override attached, and the tool arguments show up on the spans.

Issue #2537, opened 2025-01-21
🐛 Bug Report: Invalid type NoneType for attribute 'traceloop.association.properties.ls_temperature' value.


dinmukhamedm:
### Which component is this bug for?

Langchain Instrumentation

### 📜 Description

With fairly recent version of LangChain, default temperature is not `0.7` anymore (thanks LangChain for realizing), and so users get a warning
about invalid attribute value. Full warning message:

```
WARNING:opentelemetry.attributes:Invalid type NoneType for attribute 'traceloop.association.properties.ls_temperature' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

### 👟 Reproduction steps

```
from langchain_openai import ChatOpenAI
from langchain.schema import HumanMessage
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from traceloop.sdk import Traceloop

Traceloop.init()

def main():
    llm = ChatOpenAI(
        model="gpt-4o-mini",
    )
    a = llm.invoke(
        [HumanMessage("What is the capital of Australia?")]
    )
    print(a)

if __name__ == "__main__":
    main()

```

```pyproject.toml
dependencies = [
    "langchain>=0.3.15",
    "langchain-openai>=0.3.1",
    "traceloop-sdk>=0.36.1",
]

```

With some older versions of Langchain, I was not able to reproduce the bug.

### 👍 Expected behavior

None attributes should not be attempted to be set

### 👎 Actual Behavior with Screenshots

Auto-instrumentation must not attempt setting invalid attributes causing warnings

### 🤖 Python Version

3.12

### 📃 Provide any additional context for the Bug.

Not exactly sure where this happens, but I am suspecting this line

https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py#L315

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

nirga:
Thanks @dinmukhamedm! Yes I suspect some of the metadata properties can't be put directly on an attribute since they aren't string / int / bool. We should filter them out

PR #4467, opened 2026-09-09
fix(langchain): keep objects out of association properties

IdoGol24:
`_sanitize_metadata_value` passed any non-primitive through `str()`, so an object's
repr became the span attribute value. A model, client or config object renders its
constructor state, which routinely includes an API key. Association properties are
copied onto every descendant span, so one such value spread across the whole trace —
and this path is not gated by `TRACELOOP_TRACE_CONTENT`, so turning content capture
Forward plain data only, through a single `_metadata_scalar` predicate that every
branch routes through:

- **primitives** (`bool`, `str`, `bytes`, `int`, `float`) pass through unchanged;
- **stdlib scalars** — `UUID`, `datetime`, `Decimal`, `Enum`, `Path` — are
  stringified. Their `str()` renders the value itself rather than constructor state,
  so they carry none of the credential risk. A `session_id` passed as a UUID is the
  common case here, and dropping it would have reinstated the #2537 symptom that
  bfb761f set out to fix;
- **lists** keep their scalar elements and lose the rest, preserving the
  homogeneous-sequence shape cc2bf14 established;
- **mappings** are kept as JSON with only the keys JSON can carry, so one
  object-valued or `bytes` key no longer discards its siblings;
- **anything else** sanitizes to `None` and the key is dropped rather than recorded
  as a stringified object.

Documented usage is unaffected: `user_id` and `session_id` still populate association
properties, whether they arrive as strings, numbers or UUIDs.

Also tick two boxes that are now true: I have added tests (tests/test_metadata_sanitization.py, 10 cases) and PR name follows conventional commits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LangChain telemetry metadata handling.
  * Preserved supported primitive values and converted common scalar types, such as dates, UUIDs, decimals, paths, and enums, to safe text values.
  * Removed unsupported or unserializable objects from recorded metadata, including nested lists, tuples, and mappings.
  * Prevented empty or invalid metadata from appearing in span attributes or association properties.

* **Tests**
  * Added coverage for supported values, structured metadata, invalid objects, serialization, and end-to-end telemetry behavior.

### DIFF
--- a/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
+++ b/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
@@ -106,16 +106,38 @@ def _extract_class_name_from_serialized(serialized: Optional[dict[str, Any]]) ->
         return ""
 
 
+_METADATA_PRIMITIVES = (bool, str, bytes, int, float)
+
+
 def _sanitize_metadata_value(value: Any) -> Any:
-    """Convert metadata values to OpenTelemetry-compatible types."""
+    """Convert metadata values to OpenTelemetry-compatible types.
+
+    Only plain data is forwarded. An arbitrary object is dropped rather than
+    stringified: ``str()`` on a model, client or config object renders its
+    constructor state, which routinely includes credentials the caller never
+    meant to export. Association properties are also copied onto every
+    descendant span, so a single such value spreads across the whole trace.
+
+    Returns ``None`` for anything that is not plain data; callers drop those
+    keys rather than recording a placeholder.
+    """
     if value is None:
         return None
-    if isinstance(value, (bool, str, bytes, int, float)):
+    if isinstance(value, _METADATA_PRIMITIVES):
         return value
     if isinstance(value, (list, tuple)):
-        return [str(_sanitize_metadata_value(v)) for v in value]
-    # Convert other types to strings
-    return str(value)
+        # Keep primitive elements, drop object elements, preserving the
+        # existing "sequence of strings" attribute shape.
+        return [str(v) for v in value if isinstance(v, _METADATA_PRIMITIVES)]
+    if isinstance(value, dict):
+        # A mapping of plain data is legitimate metadata, so keep it as JSON.
+        # json.dumps has no ``default``, so a mapping holding a non-serializable
+        # object raises and the key is dropped instead of being stringified.
+        try:
+            return json.dumps(value)
+        except (TypeError, ValueError):
+            return None
+    return None
 
 
 def valid_role(role: str) -> bool:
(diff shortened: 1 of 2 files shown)


packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -106,16 +106,38 @@ def _extract_class_name_from_serialized(serialized: Optional[dict[str, Any]]) ->
...
+        return [str(v) for v in value if isinstance(v, _METADATA_PRIMITIVES)]
+    if isinstance(value, dict):
+        # A mapping of plain data is legitimate metadata, so keep it as JSON.
+        # json.dumps has no ``default``, so a mapping holding a non-serializable
+        # object raises and the key is dropped instead of being stringified.
+        try:
+            return json.dumps(value)
+        except (TypeError, ValueError):
+            return None
+    return None
```
doronkopit5, line 162: This catch-all is a silent regression for metadata types that carry no credential risk. Probed against this branch:

| value | before | after |
|---|---|---|
| `uuid.uuid4()` | `"3f2b…"` | dropped |
| `datetime.now()` | `"2026-09-16T…"` | dropped |
| `Enum.A` | `"E.A"` | dropped |
| `Decimal("3.14")` | `"3.14"` | dropped |

It also reaches the dict branch above: `json.dumps` has no `default`, so `{"tenant": "acme", "sid": uuid4()}` raises and the **whole mapping** is discarded, losing `tenant` because of an unrelated key. The list branch right above keeps its good elements, so the two are asymmetric.

Worth noting the history here: the `str()` fallback this replaces was added deliberately in bfb761fa7 (#2608) to fix #2537, where non-primitive metadata was silently missing from spans — OTel's validator rejected the raw value and logged `Invalid type ... for attribute`. cc2bf14f2 (#2665) then tightened the list branch for OTel's homogeneous-sequence rule. Dropping instead of stringifying reinstates the original #2537 symptom, now without the warning that used to make it visible.

https://github.com/traceloop/openllmetry/blob/5a254e4da236059441c964141e2c29b31d09fae6/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py#L132-L140

The leak this PR targets is real and worth closing. A narrower rule would close it without the regression: stringify known-safe stdlib scalars (`UUID`, `datetime`, `Decimal`, `Enum`, `Path`) and drop only genuinely arbitrary objects. That also fixes the dict case, since those values would then serialize.

Also note `session_id` in the PR description is only a primitive when it is a string; as a UUID, which is common, it is now dropped.

IdoGol24: Right. I've attended all three.
`bfb761f `added that str() deliberately, so a blanket drop re-opens #2537. Fixed in `6c0f9fd9`: one `_metadata_scalar `predicate that every branch routes through, primitives pass through, UUID/datetime/Decimal/Enum/Path stringify, arbitrary objects drop. The dict branch now keeps its serializable keys and loses only the object-valued ones, so {"tenant": "acme", "sid": uuid4()} survives intact and {"tenant": "acme", "client": <obj>} keeps tenant. The session_id line in the description was wrong and I've corrected the behaviour rather than the wording.

doronkopit5 (approved):

PR #4475, opened 2026-09-16
fix(bedrock): don't count a missing guardrailAction as an activation

### LINKED ISSUE #4471: 🐛 Bug Report: gen_ai.bedrock.guardrail.activation counts every Bedrock call
**Component:** Bedrock Instrumentation

### 📜 Description

`is_guardrail_activated()` in `guardrail.py` ends with:

```python
return response.get("amazon-bedrock-guardrailAction") != "NONE"
```

When no guardrail is configured, Bedrock omits that key, so `.get()` returns `None` and `None != "NONE"` is `True`. Every ordinary response is treated as a guardrail activation, and `gen_ai.bedrock.guardrail.activation` is incremented on every Bedrock call.

Defaulting the key fixes it: `response.get("amazon-bedrock-guardrailAction", "NONE") != "NONE"`.

### 👟 Reproduction steps

```python
from opentelemetry.instrumentation.bedrock.guardrail import is_guardrail_activated

print(is_guardrail_activated({"stopReason": "end_turn"}))  # True, expected False
```

### 👍 Expected behavior

A response with no `amazon-bedrock-guardrailAction` key is not an activation.

### 👎 Actual Behavior

It counts as one, so the metric reports activations for users with no guardrails configured.

### 🤖 Python Version

3.12

### 📃 Additional context

The same line crashes when metrics are disabled, since `_instrument()` sets the counters to `None` but `guardrail.py` dereferences them unconditionally:

```
guardrail.py", line 179, in guardrail_converse
    metric_params.guardrail_activation.add(1, attrs)
AttributeError: 'NoneType' object has no attribute 'add'
```

Present from 0.39.0 through 0.62.3.

### 👀 Duplicate check

Checked, no existing issue.

AnoshanJ:
Fixes #4471.

`is_guardrail_activated()` ended with:

```python
return response.get("amazon-bedrock-guardrailAction") != "NONE"
```

Bedrock omits `amazon-bedrock-guardrailAction` when no guardrail is configured, so `.get()` returns `None` and `None != "NONE"` is `True`. Every ordinary response was counted as a guardrail activation, incrementing `gen_ai.bedrock.guardrail.activation` on every Bedrock call.

Defaulting the key to `"NONE"` restores the intended meaning. The two checks above it already cover the real activation paths (`CONTENT_FILTERED` in `results`, and `stopReason == "guardrail_intervened"`).

Added unit tests covering a response with no guardrail key, a configured guardrail that did not fire, and the three activation paths. The first case fails without this change.

The existing guardrail cassettes are unaffected: all eight signal activation explicitly, either via `guardrail_intervened` or `amazon-bedrock-guardrailAction: INTERVENED`, so none of them relied on the fallthrough.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected guardrail activation detection when Bedrock responses omit the guardrail action.
  - Responses without an activation signal are no longer incorrectly reported as guardrail activations.
  - Guardrail interventions continue to be detected through supported action, stop-reason, and completion-reason signals.

- **Tests**
  - Added coverage for missing actions, inactive guardrails, and confirmed intervention scenarios.


Santoshkumarpuppala:
This turns `test_guardrail_converse_stream` red: converse_stream no longer records activation when the stream reports `guardrail_intervened` on messageStop.

Ran at `6f9be7a`: `pytest tests/traces/test_guardrails.py` gives 1 failed, 3 passed. The KeyError on `gen_ai.bedrock.guardrail.input_filter` (`test_guardrails.py:299`) is because span attributes are only set when activation is true (`guardrail.py:178-180`). Reverting only `guardrail.py:34` gives 4 passed. A one-off script replaying the metrics cassettes shows `guardrail.activation` per converse_stream call is 0 here, 1 on main; converse is 1 on both. CI hasn't run (`action_required`).

The description says none of the eight cassettes relied on the fallthrough. The two converse_stream ones did. The regression is `guardrail.py:34`, but the stream path depended on the fallthrough because `packages/opentelemetry-instrumentation-bedrock/opentelemetry/instrumentation/bedrock/__init__.py:798` and `:886` pass only `event["metadata"]` (read). In both cassettes that frame has no `stopReason` or `amazon-bedrock-guardrailAction`; `stopReason` is on messageStop. Called on the metadata frame, `guardrail_converse` gives 0 here, 1 on main, and 1 here once `stopReason` is added. Main also counted a stream whose metadata had no guardrail trace.

Fix: carry `stopReason` from messageStop to the metadata event (a per-event local today; nonlocal or span_state works) and pass it in. That restores activation for both stream cassettes. Happy to push it with a stream test, or leave it to you.

linhongyu510:
This correctly fixes the false-positive activation. One thing worth folding in while you're here: issue #4471 also reports a second, related crash that this PR doesn't cover yet.

When metrics are disabled (`TRACELOOP_METRICS_ENABLED=false`), `_instrument()` sets `guardrail_activation` (and the other counters) to `None`, but both call sites dereference it unconditionally on a *genuine* activation:

```python
# guardrail.py, guardrail_converse (~L179) and guardrail_handling (~L208)
if is_guardrail_activated(response):
    metric_params.guardrail_activation.add(1, attrs)   # AttributeError: 'NoneType' ...
    set_guardrail_attributes(span, input_filters, output_filters)
```

Reproducible offline:

```python
mp = MagicMock(); mp.guardrail_activation = None
guardrail_converse(span, {"stopReason": "guardrail_intervened"}, "aws", "m", mp)
# -> AttributeError: 'NoneType' object has no attribute 'add'
```

A minimal guard at both sites (keeping `set_guardrail_attributes` unconditional, since span attributes are independent of the metrics toggle) closes it:

```python
if metric_params.guardrail_activation is not None:
    metric_params.guardrail_activation.add(1, attrs)
```

I had a separate PR (#4487) covering both halves, but yours predates it on the activation fix, so I'm closing mine in favour of this one — just flagging the second half so #4471 can be fully closed here.

PR #4477, opened 2026-09-16
fix(openai): don't crash tracing responses.with_raw_response.create(stream=True)

### LINKED ISSUE #4476: Responses API streaming crashes with 'AsyncStream' object has no attribute 'id' when using .with_raw_response

OpenAI Instrumentation

### 📜 Description

`client.responses.with_raw_response.create(stream=True, ...)` crashes `async_responses_get_or_create_wrapper` with:

```
AttributeError: 'AsyncStream' object has no attribute 'id'
```

This is a variant of #4058 / PR #4078 that the fix didn't cover. #4078 taught `parse_response`/`async_parse_response`
to unwrap `APIResponse`/`AsyncAPIResponse` (used by `.with_streaming_response`), but `.with_raw_response.create()` on
`Responses` returns a `LegacyAPIResponse` (via `openai._legacy_response.async_to_raw_response_wrapper`), which was
already unwrapped before #4078. The problem isn't the unwrapping itself — it's that `LegacyAPIResponse.parse()`,
when the underlying request was `stream=True`, returns a `Stream`/`AsyncStream` instance, not a parsed `Response`.

The only `isinstance(response, (Stream, AsyncStream))` check in `async_responses_get_or_create_wrapper` runs
*before* parsing, on the raw, unparsed return of `wrapped()`:

https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py#L666

When the call goes through `.with_raw_response`, that raw return is a `LegacyAPIResponse` wrapper, not a
`Stream`/`AsyncStream`, so the check misses it. Execution falls through, `async_parse_response()` calls
`.parse()` on it, which (per the `self._stream` branch in `openai._legacy_response.LegacyAPIResponse._parse`)
returns an `AsyncStream`. There's no post-parse recheck for `Stream`/`AsyncStream`, so it crashes reading `.id`:

https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/v1/responses_wrappers.py#L746

This combination (`.with_raw_response` + `stream=True`) is exactly what `agent_framework_openai` (Microsoft Agent
Framework) uses as of >=1.6.0, to read the `x-ms-served-model` response header — so every agentic streaming turn
instrumented with this package crashes.

### 👟 Reproduction steps

Self-contained (mocks the HTTP transport, no API key needed):

```python
import asyncio
import httpx
from opentelemetry.instrumentation.openai import OpenAIInstrumentor

trace.set_tracer_provider(TracerProvider())
OpenAIInstrumentor().instrument()

from openai import AsyncOpenAI  # noqa: E402

SSE_BODY = (
    b'event: response.created\n'
    b'data: {"type":"response.created","response":{"id":"resp_123","object":"response",'
    b'"created_at":0,"status":"in_progress","model":"gpt-4o-mini","output":[]}}\n\n'
    b'event: response.completed\n'
    b'data: {"type":"response.completed","response":{"id":"resp_123","object":"response",'
    b'"created_at":0,"status":"completed","model":"gpt-4o-mini","output":[],'
    b'"usage":{"input_tokens":1,"output_tokens":1,"total_tokens":2}}}\n\n'
)

def handler(request: httpx.Request) -> httpx.Response:
    return httpx.Response(
        200,
        headers={"content-type": "text/event-stream", "x-ms-served-model": "gpt-4o-mini"},
        content=SSE_BODY,
    )

async def main():
    client = AsyncOpenAI(api_key="sk-test", http_client=httpx.AsyncClient(transport=httpx.MockTransport(handler)))
    raw = await client.responses.with_raw_response.create(
        model="gpt-4o-mini",
        input="hello",
        stream=True,
    )
    stream = raw.parse()
    async for event in stream:
        pass
    print("no crash")

asyncio.run(main())
```

### 👍 Expected behavior

No crash — either trace the stream normally (matching the non-`with_raw_response` path) or skip tracing gracefully.

### 👎 Actual Behavior

Reproduced on both:
- `opentelemetry-instrumentation-openai==0.62.3` + `openai==2.50.0`
- `opentelemetry-instrumentation-openai==0.62.3` + `openai==3.14.1` (latest)

```
File ".../opentelemetry/instrumentation/openai/v1/responses_wrappers.py", line 746, in async_responses_get_or_create_wrapper
    existing_data = responses.get(parsed_response.id)
                                  ^^^^^^^^^^^^^^^^^^
AttributeError: 'AsyncStream' object has no attribute 'id'
```

### 🤖 Python Version

3.12

### 📃 Additional context

We're currently working around this in our own codebase by unwrapping just the Responses API span instrumentation
after `OpenAIInstrumentor().instrument()` runs.

Yes I am willing to submit a PR if someone can confirm my assessment of the issue.

markuspalme:
Fixes #4476

## Problem

`client.responses.with_raw_response.create(stream=True, ...)` - what `agent-framework-openai` (Microsoft Agent Framework) >=1.6.0 uses to read the `x-ms-served-model` response header before streaming - crashes `async_responses_get_or_create_wrapper` (and its sync counterpart) with:

```
AttributeError: 'AsyncStream' object has no attribute 'id'
```

This is a variant of #4058 / #4078 that fix didn't cover. #4078 taught `parse_response`/`async_parse_response` to unwrap `LegacyAPIResponse`/`APIResponse`/`AsyncAPIResponse` so downstream code can read `.id` off whatever comes back. But when the underlying call is *streaming*, `LegacyAPIResponse.parse()` (per its `self._stream` branch) returns the `Stream`/`AsyncStream` itself, not a parsed `Response` - so there's still no `.id` to read, just via a different path than #4058 hit.

The only `isinstance(response, (Stream, AsyncStream))` check in these wrappers runs *before* parsing, on the raw, unparsed return of `wrapped()`. That only catches the direct `.create(stream=True)` case, where `wrapped()` itself returns the `Stream`. When the call goes through `.with_raw_response`, the unparsed return is a `LegacyAPIResponse` wrapper, so that check misses it, and there's no recheck after parsing.

## Fix

Add that recheck: after `parse_response()`/`async_parse_response()`, if the result is a `Stream`/`AsyncStream`, return the raw response unmodified instead of trying to read `.id` off it. This matches the existing fallback behavior a few lines down (`except Exception: return response`) when trace assembly fails for other reasons - the call goes untraced instead of crashing the host application. Fixed in both `responses_get_or_create_wrapper` and `async_responses_get_or_create_wrapper`.

I considered actually tracing this path (wrapping the resulting stream the same way the direct `.create(stream=True)` case does via `ResponseStream`), but `.with_raw_response` callers depend on the returned object still behaving like a raw response (`.headers`, `.parse()`, etc. - `agent-framework-openai` reads `.headers` before calling `.parse()`), and monkeypatching that raw wrapper's `.parse()` to inject a `ResponseStream` felt like the wrong tradeoff for a bugfix PR. Happy to take a different direction if you'd rather trace this path than skip it - opted for the smaller, safer fix.

## Testing

Added a regression test for each of the sync/async wrappers, using a mocked httpx transport that returns a real (minimal) Responses-API SSE stream, run through the actual `OpenAIInstrumentor()` instrumentation path end-to-end (not just unit-testing the parse helpers) - both fail with the reported `AttributeError` against `main`, and pass with the fix. Confirmed the crash independently against the currently-published `opentelemetry-instrumentation-openai==0.62.3` with both `openai==2.50.0` and `openai==3.14.1` before opening #4476.

```
uv run pytest tests/traces/test_responses.py -q
# 27 passed
uv run pytest -q
# 270 passed, 1 skipped
uv run ruff check opentelemetry/instrumentation/openai/v1/responses_wrappers.py tests/traces/test_responses.py
# All checks passed!
```

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Type of change

- [x] Bugfix

## Bug Fixes

- Improved synchronous and asynchronous OpenAI response streaming.
- Prevented errors caused by attempting to read unavailable metadata from raw streams.
- Ensured raw streaming responses pass through correctly without creating invalid trace spans.
- Preserved access to response headers before streamed events are consumed.

## Tests

- Added coverage for synchronous and asynchronous raw streaming, including event consumption, header access, and trace validation.


markuspalme:
Good catch, added. Asserted `raw.headers["x-ms-served-model"]` before `raw.parse()` in both tests, and confirmed it actually catches the regression you described: temporarily changing the fix to `return parsed_response` instead of `return response` fails the new assertion with `AttributeError: 'AsyncStream' object has no attribute 'headers'` (rather than silently passing).

doronkopit5 (approved):

PR #4478, opened 2026-09-17
fix(mistralai): record chunked assistant content instead of dropping it

HarianthK:
## What

Mistral's reasoning models (`magistral-*`) return the assistant message content as a list of chunks instead of a string: a `thinking` chunk (a list of text chunks) followed by a `text` chunk with the answer (`AssistantMessageContent = Union[str, List[ContentChunk]]` in the SDK). In `_set_response_attributes` the non-string branch calls `json.dumps(choice.message.content)`, but the chunks are SDK models, so it raises `TypeError: Object of type ThinkChunk is not JSON serializable`. `@dont_throw` swallows that, and the span ends with no `gen_ai.completion.0.content`, no `.role` and no `.finish_reason` for that choice. The model name and token counts survive because they are set in a separate function.

## Fix

Chunks are dumped to plain data with `model_dump(mode="json", exclude_none=True)` before `json.dumps`, so the completion is recorded as a JSON string, which is what the list branch was already trying to do. String content takes the same path as before.

## Tests

Two tests in `tests/test_chunked_content.py` build the magistral response with the SDK's own `ChatCompletionResponse.model_validate` and call the attribute setter directly, so no cassette or API key is needed: one for the chunked shape (fails on `main` with "the completion content was dropped"), one pinning plain string content as unchanged. `uv run pytest tests/` passes, 23 tests; `ruff check` is clean.

## Streaming

Both streaming accumulators did `message.content += choice.delta.content`, which raises `TypeError` on a chunk-list delta before `_handle_response` and `span.end()` run. A `_merge_delta_content` helper now joins text deltas as before, collects chunk lists as plain data, and turns a text delta that follows chunks into a text chunk, so the accumulated response serialises the same way as the non-streaming one. Two more tests drive the sync and async accumulators with a streamed magistral answer built from the SDK's `CompletionEvent` models; both fail on `main` with `can only concatenate str (not "list") to str`.

- [ ] If adding a new instrumentation or changing an existing one, I've added screenshots from some observability platform showing the change. (No platform at hand; the test prints the recorded attribute, and I can add a screenshot if you want one.)
- [x] (If applicable) I have updated the documentation accordingly. (Nothing to change.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Mistral AI instrumentation support for reasoning-model responses with structured content.
  * Fixed synchronous and asynchronous streaming responses so chunked thinking and text content accumulate correctly without errors.
  * Preserved structured content and finish reasons in recorded completion data.
  * Ensured structured content is consistently serialized in recorded completion attributes and events.
  * Plain-text responses continue to be recorded unchanged.


HarianthK:
Fair point, that was the follow-up I had mentioned; it is in now. Both accumulators merge chunk-list deltas through one helper, and two tests drive the sync and async paths with a streamed magistral answer built from the SDK's event models.

packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py
```diff
@@ -217,6 +228,26 @@ def _set_model_response_attributes(span, llm_request_type, response):
...
+        return (current or "") + delta
+    chunks = [
+        chunk.model_dump(mode="json", exclude_none=True)
+        if hasattr(chunk, "model_dump")
+        else chunk
+        for chunk in delta
+    ]
+    if isinstance(current, str) and current:
+        return [{"type": "text", "text": current}] + chunks
+    return (current if isinstance(current, list) else []) + chunks
```
doronkopit5: This appends a new entry per delta instead of extending the trailing chunk, so token-by-token streams fragment. A 6-token reasoning + 2-token answer accumulates to 8 entries (`thinking` x6, `text` x2) where the non-streaming response for the same content is 2.

Nothing is lost or misordered, but streaming and non-streaming spans end up structurally different for the same logical response, and the attribute bloats on long reasoning traces. `anthropic` and `openai` consolidate instead, e.g. `streaming.py::_process_response_item` grows one block per index: `event["text"] = event.get("text", "") + item.delta.thinking`.

packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py
```diff
@@ -149,11 +164,7 @@ def _set_response_attributes(span, llm_request_type, response):
             _set_span_attribute(
                 span,
                 f"{prefix}.content",
-                (
-                    choice.message.content
-                    if isinstance(choice.message.content, str)
-                    else json.dumps(choice.message.content)
-                ),
+                _content_as_str(choice.message.content),
```
doronkopit5, line 167: This fix isn't applied to the events path. When `use_legacy_attributes=False`, `_handle_response` calls `_emit_choice_events` *instead of* this function, and that one still passes content raw:

https://github.com/traceloop/openllmetry/blob/9022ef518fb7a6e6029c7c5dcfaf7d0ddd4c865a/packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py#L397-L399

So chunked content stays as raw `ThinkChunk`/`TextChunk` there. `asdict()` in `event_emitter.py` deep-copies rather than serializes them, and the exporter drops the event with `TypeError: Object of type ThinkChunk is not JSON serializable` — the same silent drop this PR fixes for spans.

Worth calling `_content_as_str` at L398 too. Non-streaming only; streaming is fine since `_merge_delta_content` already normalizes to plain dicts by then.

packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py
HarianthK: Agreed, that was fragmenting. Deltas now extend the trailing chunk when the type matches (text onto text, thinking text onto the last thinking text), so the streamed test case ends as one thinking chunk and one text chunk, the same two the non-streamed response has.

packages/opentelemetry-instrumentation-mistralai/opentelemetry/instrumentation/mistralai/__init__.py
HarianthK: Good catch, I had only looked at the attributes path. `_emit_choice_events` now runs the content through `_content_as_str` as well, and there is a test that emits the choice event for the magistral response and reads the chunks back out of the log record; it fails on the previous commit with the event dropped.

doronkopit5 (approved):

PR #4482, opened 2026-09-18
fix(openai-responses): remove completed entries from the global responses dict

### LINKED ISSUE #4473: Responses API: completed responses are never removed from the global `responses` dict
## What happens

`responses_wrappers.py` keeps a module-global dict:

```python
responses: dict[str, TracedData] = {}   # line 199
```

Entries go in at three places (632 sync, 800 async, 1067 streaming). They come out at two (832 and 864), and both of those live inside the *cancel* wrappers. So any response that finishes normally stays in the dict for the life of the process.

We measured about 17 KiB retained per turn in a chat workload, roughly 166 MiB per 10k turns. Our pods were restarting on OOM. The dict survives `gc.collect()`, since the module global is a live reference.

Same code in 0.62.1, in 0.62.3, and on `main` today. I diffed all three, `responses_wrappers.py` is byte identical.

## Why the entries are dead

The sync path writes unconditionally, then emits the span only for `completed`:

```python
        responses[parsed_response.id] = traced_data     # 632
    except Exception:
        return response

    if parsed_response.status == "completed":           # 636
        ...
        span.end()                                      # 647, entry still in the dict
```

After `span.end()` nothing reads that entry again. The merge reads at 579 and 746 exist so a later `retrieve` on a still running response can recover `start_time`, `input`, `tools` and `trace_context` from the original `create`, which is the background polling flow. Once the span is out, there is nothing left to merge into.

Streaming is the clearest case. Line 1067 writes to the dict and the next three lines finish the span:

```python
                    responses[parsed_response.id] = self._traced_data   # 1067

                set_data_attributes(self._traced_data, self._span)
                self._span.set_status(StatusCode.OK)
                self._span.end()
```

Nothing ever reads that entry. It is garbage from the moment it is written.

Two things that might look like they need the retention, and do not:

- `previous_response_id` never appears in the file, so multi turn chaining never looks anything up. Turn N+1 gets a fresh id and misses the dict.
- `background` never appears either. There is no special handling, though the background flow (`create` returns `queued`, `retrieve` polls until `completed`) is exactly what the merge is for.

## A second bug in the same block

While tracing the reads I hit this: nothing guards against emitting a span twice. `create`, `retrieve` and `parse` all bind to `responses_get_or_create_wrapper` (`v1/__init__.py:305-318`), the gate at 636 and 804 only checks `status == "completed"`, and `TracedData` carries no "already emitted" field. Calling `retrieve` on a response that already completed emits another span for the same call.

This matters for the fix. Delete the entry at `span.end()` and nothing else, and that duplicate span gets worse instead of disappearing: `start_time` falls back to the retrieve's own timestamp instead of the create time, `input`/`instructions`/`tools` come back empty because a bare `retrieve` carries none, and `trace_context` becomes the current context instead of the originating one.

## Proposed fix

Two changes that belong together:

1. Delete the entry where the span is emitted (after 647 and 813), and drop the dead write at 1067 entirely.
2. Track emitted response ids in a bounded structure (LRU or TTL) and skip the emission block at 636/804 when the id is already there.

Change 1 alone leaves the degraded duplicate described above, which is why I would not land it by itself. Change 2 removes duplicate spans, which is a behavior change and the part worth your call before I open a PR.

The guard holds bare id strings, not `TracedData` with its `input`, `output_blocks` and `tools`, so bounding it costs little. Dropping an id from the guard just means a duplicate span again, which is today's behavior. Dropping an entry from the main dict would lose `start_time` and `trace_context` and degrade a real span, so the dict itself should stay exact rather than capacity bound.

I considered swapping the plain dict for a TTL or LRU cache, one line at 199. It caps growth instead of removing it, every completed response still sits there until eviction, the size is a guess about someone else's workload, evicting too early breaks an in flight background poll, and the duplicate spans stay. Happy to go that way if you prefer the smaller diff.

## Tests

A regression test needs no cassette and no network. The leak shows up in module state, and the file already has unit tests in that style, for example `test_responses_trace_context_propagation_unit` (tests/traces/test_responses.py:562), which imports `TracedData` from the wrapper directly and asserts on a local `InMemorySpanExporter`. Driving the completed path N times and asserting `len(responses) == 0` fits the same mold.

Worth noting for anyone reproducing: one cassette per test function, and no test sets `allow_playback_repeats`, so the VCR route would need a hand edited cassette. The unit style avoids that.

## Unrelated, noticed in passing

The exception path at 514 and 685 reads `non_sentinel_kwargs.get("response_id")`, but `response_id` is positional in the SDK (`openai/resources/responses/responses.py:1315`, before the `*`). For the normal `client.responses.retrieve("resp_abc")` that lookup always misses. Separate from this issue, say the word and I will open another.

---

Glad to send the PR if the two part fix works for you.

ZxlDragonDoctor:
### Problem

esponses_wrappers.py keeps a module-global 
esponses dict. Entries are inserted on create/get/streaming paths but only removed in cancel wrappers, so any response that finishes normally stays for the life of the process (#4473 reports OOM from retained TracedData). A later 
etrieve/parse on an already completed response could also emit a degraded duplicate span.

### Solution
1. After a successful completion span is emitted, pop the dict entry on sync/async/stream success paths. Cancel-path cleanup is unchanged.
2. Gate streaming cleanup on 
esponse_status == "completed" so in-progress entries remain available for Responses.retrieve polling (CodeRabbit review on this PR).
3. Track emitted response ids in a bounded LRU and skip a second completion span for an id already emitted, preserving create-time start_time/input/	ools/	race_context (#4473 part 2).

### Testing
- Command: python packages/opentelemetry-instrumentation-openai/tests/offline_review_4482.py
- Result: 3 passed, 0 failed
- Cases: completed stream pop; in-progress stream retained; emitted-id helper first/second mark; source contracts for sync/async pops and status-gated stream cleanup
- Added unit tests in 	ests/traces/test_responses.py for completed pop, in-progress retention, and duplicate-emission skip

### Agent dimension
streaming-state (observability memory retention on agent/response streams)

### Core value
Completed agent Responses traces no longer pin TracedData for the process lifetime, and retrieve-after-complete no longer emits a degraded duplicate span.

### Issue alignment
Addresses #4473:
- success paths remove completed entries (as reported)
- bounded emitted-id guard for the duplicate-span second bug called out in the issue body
- CodeRabbit review on this PR: stream cleanup restricted to completed responses
Fixes #4473

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate telemetry spans for the same completed OpenAI Responses API response.
  * Removes completed responses from temporary tracking after synchronous, asynchronous, and streamed requests.
  * Keeps in-progress streaming responses available for later retrieval.
  * Ensures canceled responses are cleaned up correctly.
  * Bounds duplicate-completion tracking to prevent unbounded memory growth.


ZxlDragonDoctor:
Pushed review follow-up on this PR:

1. **Stream cleanup gated on completed status** - ResponseStream._process_complete_response now pops only when 
esponse_status == "completed" and a 
esponse_id exists, so in-progress entries stay available for Responses.retrieve polling.
2. **Bounded emitted-id guard** - completion spans are emitted once per response id (LRU, 2048). A later 
etrieve/parse on an already completed response no longer emits a degraded duplicate span (#4473 part 2).
3. **Tests** - offline contract checks (3 passed) plus unit cases for completed pop, in-progress retention, and duplicate-emission skip.
4. **PR body** - cleaned and aligned with the final diff; Testing section reflects the current branch run.

Happy to adjust the LRU size or drop part 2 if maintainers prefer a smaller first landing.

ZxlDragonDoctor:
Pushed a follow-up for the remaining review notes on this PR:

1. **Thread-safe `_mark_response_emitted`** - module-global `threading.Lock` now covers lookup, LRU update, insert, and eviction so two completed wrappers cannot both observe first-emission.
2. **Check before ending the stream span** - `ResponseStream._process_complete_response` calls `_mark_response_emitted` *before* `span.end()`. When the helper returns `False` (another wrapper already emitted this `response_id` via retrieve/parse), the stream span is still closed but completion attributes are not re-exported; the completed entry is still popped from `responses`.

Offline contract tests: `python packages/opentelemetry-instrumentation-openai/tests/offline_review_4482.py` ? **3 passed**.

Branch: `fix/responses-completed-pop` @ `c4c3f68`.

ZxlDragonDoctor:
Pushed a follow-up for the latest review note on this PR:

**	hreading import in the duplicate-emission unit test**
- 	est_duplicate_completed_emission_is_skipped now does import threading before constructing 	hreading.Lock() (CodeRabbit F821 / NameError).
- Verified offline that every test function using 	hreading.Lock() imports 	hreading in the same function body (offline_test_4482_threading.py, 2 passed).

Branch ix/responses-completed-pop updated to 49b17b5.

ZxlDragonDoctor:
Friendly ping - review follow-ups are pushed (stream cleanup gating, atomic _mark_response_emitted, test 	hreading import). The branch is currently behind main and shows mergeable=false; I can rebase on request or once CI is green on the current head. Happy to adjust anything blocking merge. Thanks!

PR #4484, opened 2026-09-19
fix(groq): record metrics for streaming responses

### LINKED ISSUE #4419: 🐛 Bug Report: Groq streaming calls record no token or duration metrics

CJstate:
## Summary

Streaming (`stream=True`) Groq chat completions never recorded metrics: the stream processors only set span attributes, so `gen_ai.client.token.usage` and `gen_ai.client.operation.duration` were emitted for non-streaming calls only and streaming calls were invisible on metric dashboards.

This passes the token/duration histograms into the sync/async stream processors and records the accumulated usage (from `chunk.x_groq.usage`) plus the elapsed duration once the stream is drained.

- `_handle_streaming_response`: record token + duration histograms using the accumulated `usage`.
- `_create_stream_processor` / `_create_async_stream_processor`: accept the histograms and `start_time` (all optional, backward-compatible), record on stream completion.
- New regression test drives the processor with mocked chunks and asserts both metrics land.

Fixes #4419

## Checklist

- [x] PR name follows conventional commits format: `fix(groq): ...`.

## Summary by CodeRabbit

* **New Features**
  * Streaming Groq operations now report input and output token usage metrics.
  * Streaming operations now report duration metrics.
  * Metrics include provider, operation, and token type details for improved observability.

* **Bug Fixes**
  * Token usage is now recorded when final streaming responses contain usage data but no response choices.
  * Synchronous and asynchronous streaming metrics now consistently capture token and duration information.


CJstate:
Addressed the CodeRabbit finding (`_process_streaming_chunk` losing usage on empty-choice chunks) in 3d00c66.

```python
# Read usage before the empty-choices guard. Groq attaches x_groq.usage to a
# final chunk that may carry no choices at all; returning early would drop the
# only usage-bearing chunk and leave the token histograms empty.
usage = None
if hasattr(chunk, "x_groq") and chunk.x_groq and chunk.x_groq.usage:
    usage = chunk.x_groq.usage

if not chunk.choices:
    return None, [], [], usage
```

Regression coverage added for both the sync and async processors (`test_streaming_records_tokens_when_usage_chunk_has_no_choices`, `test_async_streaming_records_tokens_when_usage_chunk_has_no_choices`). They assert the recorded token sums rather than only the metric names — the earlier tests passed either way because `_chunk()` always carried a zeroed usage. Both fail without this commit (`prompt tokens not recorded, got {}`).

CJstate:
Pushed f08d810: the existing `test_empty_choices_returns_none_quad` in `tests/traces/test_init.py` asserted that usage is *always* `None` for empty-choice chunks — the exact behaviour this PR changes. It passed only because the bare `MagicMock` made the `chunk.x_groq.usage` truthiness check succeed.

That test now sets `x_groq = None` so it states the real contract, and a sibling case covers usage being preserved. Full package suite: 129 passed, `ruff check` and `ruff format --check` clean.

PR #4486, opened 2026-09-21
fix(anthropic): skip NotGiven/Omit sentinels in set_span_attribute

### LINKED ISSUE #4431: 🐛 Bug Report: NotGiven sentinel leak in Anthropic instrumentation

Anthropic Instrumentation

### 📜 Description

set_span_attribute() in \traceloop\openllmetry\tree\main\packages\opentelemetry-instrumentation-anthropic\opentelemetry\instrumentation\anthropic\utils.py filters None and empty strings but not the Anthropic SDK's NOT_GIVEN sentinel (type Omit / NotGiven). When optional parameters like temperature or top_p are not explicitly set, the sentinel leaks into span.set_attribute(), producing OTel type warnings:

```
Invalid type Omit for attribute 'gen_ai.request.temperature' value.
Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

Same pattern as #4230(log Body) and #3472(OpenAI Omit), but in Anthropic span attributes.

### 👟 Reproduction steps

1. Install opentelemetry-instrumentation-anthropic and anthropic SDK (0.52+)
2. Call client.messages.create(model="...", max_tokens=1024, messages=[...]) without setting temperature or top_p
3. Observe warning: Invalid type Omit for attribute 'gen_ai.request.temperature'

### 👍 Expected behavior

set_span_attribute() should silently skip NOT_GIVEN / Omit sentinel values, same as it skips None and "".

The sentinel passes through to span.set_attribute(), which rejects non-primitive types and logs a warning per attribute.

### 🤖 Python Version

3.11 / 3.13 (reproducible on both)

The OpenAI instrumentation already handles this in two places:
・_set_span_attribute checks value == openai.NOT_GIVEN
・_sanitize_sentinel_values() strips sentinels from kwargs(#3472)

linhongyu510:
## What

`set_span_attribute` in the Anthropic instrumentation filters `None` and `""`, but not the Anthropic SDK's `NotGiven` / `NOT_GIVEN` and `Omit` sentinels. When an optional request parameter such as `temperature` or `top_p` is left unset, the SDK passes the sentinel through, it is neither `None` nor `""`, so it reaches `span.set_attribute`, and OpenTelemetry rejects it:

```
Invalid type NotGiven for attribute 'gen_ai.request.temperature' value.
Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

One warning per unset parameter, on every call. Fixes #4431.

Before:

```python
def set_span_attribute(span, name, value):
    if value is not None:
        if value != "":
            span.set_attribute(name, value)
    return
```

## Fix

Discover the sentinel types at runtime and skip them — the same approach the OpenAI instrumentation already uses (`_get_openai_sentinel_types` in `openai/v1/responses_wrappers.py`), so this stays consistent with the existing codebase and degrades gracefully on older SDKs that lack one of the types.

```python
def set_span_attribute(span, name, value):
    if value is None:
        return
    if _ANTHROPIC_SENTINEL_TYPES and isinstance(value, _ANTHROPIC_SENTINEL_TYPES):
        return
    if value != "":
        span.set_attribute(name, value)
    return
```

## Reproduction (offline, no API key / network / collector)

I reproduced the defect against the real `set_span_attribute` with a real OTel span, capturing the `opentelemetry.attributes` logger:

```
temperature/top_p left unset  ->  kwargs['temperature'] == anthropic.NOT_GIVEN
NOT_GIVEN is None?  False
NOT_GIVEN == ''?    False
captured warning:  Invalid type NotGiven for attribute 'gen_ai.request.temperature' ...  (x2 temperature, x1 top_p)
```

After the fix, same scenario against the installed package: the sentinels produce no attribute and no warning, while `model`, `max_tokens=1024` and a real `temperature=0.7` still land on the span.

## Tests

Added `tests/test_set_span_attribute.py` (offline, no cassettes):

```
test_not_given_sentinel_is_not_set              PASSED
test_not_given_does_not_emit_otel_warning       PASSED
test_real_values_still_set                      PASSED
test_falsy_but_valid_values_still_set           PASSED   # 0 and False must survive
test_none_and_empty_string_still_skipped        PASSED
test_omit_sentinel_is_not_set                   PASSED
test_sentinel_types_discovered                  PASSED
```

The tests are load-bearing: reverting just the `set_span_attribute` body (keeping the sentinel discovery) makes `test_not_given_does_not_emit_otel_warning` fail with exactly the `Invalid type NotGiven` warning from the issue. `test_falsy_but_valid_values_still_set` guards the classic over-filtering mistake — `0` (a valid temperature) and `False` (a valid `is_streaming`) must not be dropped.

`ruff check` is clean on both files (repo config, line-length 120). No existing tests were modified; `test_json_encoder.py` still passes alongside the new file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Anthropic SDK sentinel values from being recorded as OpenTelemetry span attributes.
  * Continued supporting valid values such as `0` and `False`, while ignoring empty or unavailable values.
  * Eliminated related invalid-type warnings during instrumentation.


linhongyu510:
Closing #4486 as a duplicate — I opened it without checking for existing PRs first, and #4432 (2026-08-20) already fixes the same NotGiven/Omit sentinel leak with a test. It predates mine by a month, so it should be the one reviewed.

One small note for whoever reviews #4432, in case it's useful: it matches on `type(value).__name__ not in ("Omit", "NotGiven")`, i.e. the exact class name. That works today, but it misses a subclass of `NotGiven`/`Omit` and silently stops working if the SDK renames the class. An `isinstance` check against the actual types discovered at import time is a bit more robust:

```python
try:
    from anthropic import NotGiven, Omit
    _SENTINELS = (NotGiven, Omit)
except ImportError:
    _SENTINELS = ()
...
if _SENTINELS and isinstance(value, _SENTINELS):
    return
```

Either way #4432 is the right PR here; just flagging the option. Closing mine in its favour.

PR #4487, opened 2026-09-21
fix(bedrock): do not report a guardrail activation on every call

### LINKED ISSUE #4471: 🐛 Bug Report: gen_ai.bedrock.guardrail.activation counts every Bedrock call

linhongyu510:
## What

`is_guardrail_activated` ended with:

```python
return response.get("amazon-bedrock-guardrailAction") != "NONE"
```

When no guardrail is configured, Bedrock omits that key, so `.get()` returns `None`, and `None != "NONE"` is `True`. Every ordinary response — from users with **no guardrail configured at all** — was treated as a guardrail activation, and `gen_ai.bedrock.guardrail.activation` was incremented on every Bedrock call. Fixes #4471.

This bites the `converse` path in particular: `guardrail_converse` calls `is_guardrail_activated(response)` directly, and a normal converse response has no `amazon-bedrock-guardrailAction` key.

## Fix

Two changes in `guardrail.py`:

1. **Default the key to `"NONE"`** so an absent key is not an activation:
   ```python
   return response.get("amazon-bedrock-guardrailAction", "NONE") != "NONE"
   ```

2. **Guard the two `guardrail_activation.add()` call sites against `None`.** The issue also notes a crash: when metrics are disabled, `_instrument()` sets every counter to `None`, but the two activation call sites dereferenced it unconditionally, so a *genuine* activation raised `AttributeError: 'NoneType' object has no attribute 'add'`. The guard skips the counter while still setting the span attributes (span attributes are independent of the metrics toggle).

`+5/-3` in one source file.

## Reproduction (offline, no AWS)

Detection, before the fix:

```
is_guardrail_activated({"stopReason": "end_turn"})            -> True   (should be False)
is_guardrail_activated({})                                    -> True   (should be False)
is_guardrail_activated({"amazon-bedrock-guardrailAction":"NONE"}) -> False  (ok)
is_guardrail_activated({"stopReason":"guardrail_intervened"}) -> True   (ok, real activation)
```

Metrics-disabled crash, before the fix:

```
metric_params.guardrail_activation = None
guardrail_converse(span, {"stopReason": "guardrail_intervened"}, ...)
-> AttributeError: 'NoneType' object has no attribute 'add'
```

Both are resolved after the change, and real activations (`GUARDRAIL_INTERVENED`, `stopReason == "guardrail_intervened"`, `completionReason == "CONTENT_FILTERED"`) are still detected.

## Tests

Added `tests/test_guardrail_activation.py` (pure unit, no cassette / no AWS):

```
test_is_guardrail_activated[...]                                7 cases PASSED
test_converse_without_guardrail_does_not_record_activation      PASSED
test_converse_with_activation_records_once                      PASSED
test_activation_with_metrics_disabled_does_not_crash            PASSED
```

Load-bearing, verified by reverting each half:

- reverting the `"NONE"` default fails the "no key → False" cases and `test_converse_without_guardrail_does_not_record_activation`;
- reverting the `None` guard fails `test_activation_with_metrics_disabled_does_not_crash` with the exact `AttributeError` from the issue.

`ruff check` clean on both files (repo config).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Bedrock guardrail detection so responses without a guardrail action are not incorrectly reported as activated.
  - Prevented errors when guardrail activation metrics are unavailable or disabled.
  - Ensured activation metrics are recorded only for genuine guardrail interventions.

- **Tests**
  - Added regression coverage for absent, inactive, and active guardrail responses, including scenarios with metrics disabled.


linhongyu510:
Closing as a duplicate — I opened this without checking for existing PRs first. #4475 (2026-09-16) already fixes the false-positive activation with the same one-line default.

It doesn't yet cover the second half of #4471 (the `guardrail_activation.add` crash when metrics are disabled), so I've left the details and a minimal guard on #4475 so it can be folded in there and #4471 fully closed. Closing this in favour of #4475.

PR #4489, opened 2026-09-21
fix(mcp): never turn a tracing error into a broken tool call

### LINKED ISSUE #4463: 🐛 Bug Report: MCP client send_request is @dont_throw'd, so a tracing error returns None and breaks the tool call

Traceloop SDK

### 📜 Description

In `opentelemetry-instrumentation-mcp`, the client `send_request` wrapper is decorated with `@dont_throw`:

```python
def patch_mcp_client(self, tracer: Tracer):
    @dont_throw
    async def traced_method(wrapped, instance, args, kwargs):
        ...
```

`dont_throw` catches every `Exception`, logs it at `DEBUG`, and then falls through — so it **returns `None`**:

```python
async def async_wrapper(*args, **kwargs):
    try:
        return await func(*args, **kwargs)
    except Exception as e:
        _handle_exception(e, func, logger)   # logger.debug(...), no return
```

`dont_throw` is safe on a wrapper whose return value is discarded, but `BaseSession.send_request` **returns the RPC result**. So any exception raised inside the instrumentation replaces the real `CallToolResult` with `None`, and the MCP SDK then dereferences it:

```python
# mcp/client/session.py, ClientSession.call_tool
result = await self.send_request(..., types.CallToolResult, ...)
if not result.isError:          # AttributeError when instrumentation returned None
```

**Impact**

An error in *tracing* becomes an error in the *traced call*. The tool call fails with a confusing `AttributeError` that names neither MCP nor the instrumentation, and because the real cause is logged at `DEBUG`, any deployment running at `INFO` or above has no record of what actually went wrong.

**Observed error**

```
AttributeError: 'NoneType' object has no attribute 'isError'
```

Seen in production on a `tools/call` to a remote MCP server; the tool became unusable for that run while every other MCP tool kept working.

**Two candidate throwers inside the wrapper**

1. `carrier["traceparent"]` is read unconditionally after injection:

```python
carrier = {}
TraceContextTextMapPropagator().inject(carrier)
meta.traceparent = carrier["traceparent"]   # KeyError when nothing was injected
```

`inject()` writes no `traceparent` when the current span context is invalid or non-recording, so this raises `KeyError`.

2. Post-call span decoration in `_execute_and_handle_result` reads `result.content[0].text` unguarded:

```python
if hasattr(result, "isError") and result.isError:
    if len(result.content) > 0:
        span.set_status(Status(StatusCode.ERROR, f"{result.content[0].text}"))
```

This raises `AttributeError` for any non-text content block (image, resource, audio), all of which are valid MCP content types. This one is the more serious shape, because it happens **after** the RPC has already been made — so the result exists and is then thrown away.

**File**
- `packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/instrumentation.py`
- `packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/utils.py`

**Method**
- `McpInstrumentor.patch_mcp_client.traced_method`
- `McpInstrumentor._execute_and_handle_result`
- `dont_throw`

**Proposed fix**

Any of these, in rough order of preference:

1. Do not apply `dont_throw` to a wrapper whose return value is load-bearing. Instead, contain failures around the *instrumentation* work only, and always return the wrapped call's result.
2. Give `dont_throw` (or a `dont_throw_preserving_result` variant) an explicit fallback: `return await wrapped(*args, **kwargs)` is not safe once the call may already have run, so the fix belongs inside the wrapper where the result is in scope.
3. Independently, guard the two reads above: `carrier.get("traceparent")` and `getattr(result.content[0], "text", None)`.

Reproduced against 0.53.3 and confirmed present in 0.62.3.

### 👟 Reproduction steps

1. Instrument an MCP client with `opentelemetry-instrumentation-mcp`.
2. Make a `tools/call` whose result contains a single non-text content block (for example an image or an embedded resource) and `isError` set, so post-call span decoration reads `result.content[0].text`.
3. Observe that `_execute_and_handle_result` raises, `dont_throw` swallows it and returns `None`, and the call fails in `ClientSession.call_tool` with `AttributeError: 'NoneType' object has no attribute 'isError'`.

An equivalent path: invoke a tool while the current span context is non-recording, so `carrier["traceparent"]` raises `KeyError` before the RPC is even issued.

### 👍 Expected behavior

A failure inside tracing should never change the outcome of the traced call. `send_request` should return the MCP result it received (untraced or partially traced if necessary), and the swallowed instrumentation exception should be logged at a level visible in production rather than `DEBUG`.

linhongyu510:
## What

`patch_mcp_client` wraps `BaseSession.send_request`, whose return value **is** the RPC result the MCP SDK dereferences:

```python
# mcp/client/session.py, ClientSession.call_tool
result = await self.send_request(..., types.CallToolResult, ...)
if not result.isError:   # AttributeError if the wrapper returned None
```

The wrapper was decorated with `@dont_throw`, which catches every exception, logs it at `DEBUG`, and returns `None`. So any error inside the instrumentation replaced the real `CallToolResult` with `None`, and the SDK then raised:

```
AttributeError: 'NoneType' object has no attribute 'isError'
```

An error in *tracing* became an error in the *traced call*, and because the real cause was logged at `DEBUG`, deployments at `INFO`+ had no record of it. Fixes #4463.

Two concrete throwers inside the wrapper reached `@dont_throw`:

1. `meta.traceparent = carrier["traceparent"]` — `inject()` writes no `traceparent` when the current span context is invalid or non-recording, so this raised `KeyError`.
2. `_execute_and_handle_result` decorated an `isError` result via `result.content[0].text`, which raises `AttributeError` for any non-text content block (image, resource, audio) — all valid MCP content types.

## Fix

The core issue is that `@dont_throw` must never sit on a wrapper whose return value is consumed. So:

- **Drop `@dont_throw` from `patch_mcp_client`** and guard only the trace-context *preparation* in a `try/except` that logs at debug. The wrapped call is always awaited and its result/exception pass through unchanged — a tracing failure can no longer produce `None` or swallow a real error.
- **Read the injected header with `carrier.get("traceparent")`** and skip propagation when it is absent, instead of indexing.
- **Tolerate non-text content** when setting the error status: fall back to `str(block)` when the first content block has no `.text`.

The other `@dont_throw` uses in this file are left as-is: they wrap session enter/exit and event helpers whose return value is not the RPC result.

## Reproduction (offline)

```python
@dont_throw
async def traced_method(...):
    raise KeyError("traceparent")
    return await wrapped()

result = await traced_method(real_call, ...)   # -> None
result.isError                                  # AttributeError: 'NoneType' ...
```

## Tests

Added `tests/test_send_request_passthrough.py` (pure unit, no live MCP server):

```
test_result_passthrough_on_success                     PASSED
test_tracing_prep_failure_does_not_break_the_call      PASSED
test_real_exception_propagates_not_swallowed           PASSED
test_iserror_with_non_text_content_does_not_raise      PASSED
```

Load-bearing — against the current `main` (with `@dont_throw` and the two throwers) three of the four fail exactly as the bug predicts:

```
test_tracing_prep_failure_does_not_break_the_call  -> assert None is <result>   (returned None)
test_real_exception_propagates_not_swallowed       -> DID NOT RAISE             (swallowed)
test_iserror_with_non_text_content_does_not_raise  -> assert None is <result>   (raised then swallowed to None)
```

`ruff check` is clean on both files.

**Note on the existing suite:** `tests/test_error_type.py` (which also exercises the `isError` path) is an end-to-end test that spins up a real `fastmcp` server. I could not run it in this environment because of a `fastmcp`/`mcp` version incompatibility unrelated to this change (`ImportError: cannot import name 'IdentityAssertionParams'` during `fastmcp` import). The new tests are deliberately server-free so they run without that dependency; happy to adjust if CI exercises the e2e path differently.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MCP tracing reliability when trace context cannot be prepared.
  - Valid RPC results are now preserved even if tracing setup encounters an error.
  - MCP responses containing images, resources, audio, or other non-text content are handled without raising additional errors.
  - Genuine RPC errors continue to propagate correctly.

- **Tests**
  - Added coverage for request passthrough, tracing failures, RPC exceptions, and non-text response content.


linhongyu510:
Good catch, addressed in the follow-up commit. You're right that isolating only the trace-context prep left two gaps.

- **Before `wrapped`:** the input-span decoration in `_handle_tool_call` / `_handle_mcp_method` (span kind, entity name, input serialization) is now a `@dont_throw` helper (`_decorate_tool_input_span` / `_decorate_mcp_input_span`), so a failure there can no longer prevent the RPC.
- **After `wrapped`:** `_execute_and_handle_result` now awaits the call outside the tracing try, then does result serialization + span decoration in `@dont_throw` helpers (`_decorate_result_span` / `_decorate_error_span`). The RPC result/exception is the sole source of truth for the return value — a tracing failure can neither replace a valid result nor mask/fabricate the RPC exception.

Added `test_post_call_span_decoration_failure_does_not_break_result`, which points the wrapper at a tracer whose span raises on every `set_attribute`/`set_status`/`record_exception` and asserts the real result still passes through. All 5 tests pass; `ruff` is clean.

linhongyu510:
Closing this as a duplicate. I opened it for #4463 without first checking for existing PRs — my mistake. #4464 (from 2026-09-07) already addresses the same issue with essentially the same approach: drop `@dont_throw` from `patch_mcp_client`, guard only the trace-context prep, split the wrapped call from span decoration in `_execute_and_handle_result`, and handle non-text content blocks. #4468 is a second take on the same fix.

#4464 predates this by two weeks and already carries regression tests, so it should be the one reviewed. Apologies for the noise — closing in its favour.

